PR #26, opened 2022-05-09
Fixups

mudler:
A bunch of fixups:
- Fixes CI ARM builds on master branch
- Override alpine openrc service cmd writing over /etc/rancher/k3s/* instead of overwriting the default initrd service
- Enable openntpd on alpine
- Re-enable tests on alpine (hopefully they should pass now)


mudler:
green tests: https://github.com/c3os-io/c3os/actions/runs/2309279631, adding just couple of unrelated fixups

PR #35, opened 2022-07-01
Support builds from behind a proxy

paddy-hack:
Trying to run `./build.sh` at the office, I quickly found out that building from behind a proxy didn't work.

This fixes that as long as the `https_proxy`, `http_proxy` and `no_proxy` environment variables are set appropriately for the environment where the build is performed.  Any of these variables can be left unset if not needed.


mudler:
Hi @paddy-hack !

Thanks for taking time into this, looking good here!

mudler (approved):

PR #37, opened 2022-07-02
Split off cli into separate binaries

mudler:
This changeset splits the cli into 3 different executables to improve test coverage and allows to have featureset composable, making de-facto c3os also a base building block.

WIP as needs docs , samples and manual QA still


mudler:
manual QA is ok, missing docs/examples, but can be done likely in a follow-up

PR #54, opened 2022-08-03
Abstract Cluster Provisioning with Cluster Plugins

paynejacob:
This PR is part of a broader effort to abstract providers and publish more simplistic base images that can be extended later. 

New Features:
- `c3os-agent notify` command allows for issuing arbitrary events to c3os  agent plugins
- new sdk module allows plugins to install a minimal set of dependencies
- added engine plugin sdk for supporting additional kubernetes engines

Changes:
- moved pkg/bus to sdk/bus
- updated missing images in earthly file


mudler:
At first glance it looks good, but I'd like to go over it properly tomorrow

pkg/config/config.go
```diff
@@ -26,6 +53,8 @@ type Install struct {
 
 type Config struct {
 	Install *Install `yaml:"install,omitempty"`
+	Cluster Cluster  `yaml:"cluster,omitempty"`
```
mudler, line 56: I do wonder if it does belong here, providers can extend the config without attaching to the main config structure

cmd/agent/main.go
```diff
@@ -91,6 +91,26 @@ See https://docs.c3os.io/after_install/upgrades/#manual for documentation.
 		},
 	},
 
+	{
+		Name:      "boot",
+		Usage:     "boot",
```
mudler, line 96: A spare though: we could have a generic 'notify' that accepts an event as argument. In that case we could then stream down all the events coming from elemental too

pkg/config/config.go
paynejacob: We could move Cluster to the sdk and leave it to the plugin to parse.  Given C3OS and the agent don't interact with the cluster configuration I think this makes sense.

cmd/agent/main.go
paynejacob: I like that option, the only reason this is in the agent is to leverage the manager.

paynejacob:
replaced by https://github.com/c3os-io/c3os/pull/55

Issue #47, opened 2022-08-01
Alpine: test fails "Call to flock failed"


mudler:
**C3OS version:**
Current master, not released

**CPU architecture, OS, and Version:**

**Describe the bug**
Test failure: https://github.com/c3os-io/c3os/runs/7554332112?check_suite_focus=true

**To Reproduce**
N/A

**Expected behavior**

**Logs**
```
      <string>: {"level":"info","ts":1658989504.254293,"caller":"provider/bootstrap.go:96","msg":"Configuring VPN"}
      {"level":"info","ts":1658989504.3057752,"caller":"provider/bootstrap.go:96","msg":"Configuring VPN"}
      Provider c3os at /usr/bin/agent-provider-c3os had an error: Failed setup VPN: could not start svc: failed starting service: edgevpn.  * Call to flock failed: Resource temporarily unavailable
       * WARNING: edgevpn is already starting
       (exit status 1)
```

**Additional context**
Maybe a `/run` / `/var/run` issue?

PR #55, opened 2022-08-05
Abstract Cluster Provisioning with Cluster Plugins

paynejacob:
This PR is part of a broader effort to abstract providers and publish more simplistic base images that can be extended later. 

New Features:
- `c3os-agent notify` command allows for issuing arbitrary events to c3os  agent plugins
- new sdk module allows plugins to install a minimal set of dependencies
- added engine plugin sdk for supporting additional kubernetes engines

Changes:
- moved pkg/bus to sdk/bus
- updated missing images in earthly file


Earthfile
```diff
@@ -31,9 +31,10 @@ all-arm:
   BUILD +arm-image
 
 go-deps:
-    FROM golang
+    FROM golang:1.17
```
mudler, line 34: small nit: this sounds a good candidate for a top level `ARG`, so one place to bump the golang version used

Earthfile
```diff
@@ -82,7 +83,7 @@ build:
     BUILD +build-c3os-agent-provider
 
 dist:
-    FROM golang
+    FROM golang:1.17
```
mudler, line 86: ditto, a common ARG can be re-used here

cmd/agent/main.go
```diff
@@ -91,6 +91,26 @@ See https://docs.c3os.io/after_install/upgrades/#manual for documentation.
...
+		Aliases: []string{},
+		Flags:   []cli.Flag{},
+		Action: func(c *cli.Context) error {
+			dirs := []string{"/oem", "/usr/local/cloud-config"}
+			args := c.Args()
+			if len(args) > 1 {
+				dirs = args[1:]
+			}
+
+			return agent.Notify(args[0], dirs)
```
mudler, line 110: small nit: we should consider failing here if we pass an event which is not exposed in the bus. in that way we can signal the user earlier if no event is registered in the sdk's bus. otherwise the event would be silently ignored

overlay/files/system/oem/02_agent.yaml
```diff
@@ -6,6 +6,9 @@ stages:
        - systemctl start c3os-agent
        - systemctl enable c3os-agent
     initramfs:
+    - if: '[ ! -f "/run/cos/recovery_mode" ] && [ ! -f "/run/cos/live_mode" ]'
```
mudler, line 9: small nit: this sounds a good candidate for a separate oem file. maybe a `_notify.yaml` where we hook 1:1 all the elemental events

sdk/clusterplugin/plugin.go
```diff
@@ -0,0 +1,83 @@
...
+// stages after initramfs.
+type ClusterProvider func(cluster Cluster) yip.YipConfig
+
+// ClusterPlugin creates a cluster plugin from a `ClusterProvider`.  It calls the cluster provider at the appropriate events
+// and ensures it configuration is written where it will be executed.
+type ClusterPlugin struct {
+	Provider ClusterProvider
+}
+
+func (p ClusterPlugin) onBoot(event *pluggable.Event) pluggable.EventResponse {
```
mudler, line 26: a question here, it looks this code is not really used - i'm not sure if it's because there are plans for follow-ups. or is that a leftover? maybe this piece doesn't need to sit on c3os

sdk/go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/c3os-io/c3os/sdk
+
```
mudler, line 2: small nit: two go.mod would likely collide, is there a specific reason for it?

mudler (commented):
Just couple of nits and some questions, but overall it looks good and going in the correct direction

go.mod
```diff
@@ -195,18 +193,21 @@ require (
...
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
+	gopkg.in/yaml.v1 v1.0.0-20140924161607-9f9df34309c0 // indirect
 	howett.net/plist v1.0.0 // indirect
 	lukechampine.com/blake3 v1.1.7 // indirect
 )
 
 replace github.com/elastic/gosigar => github.com/mudler/gosigar v0.14.3-0.20220502202347-34be910bdaaf
+
+replace github.com/c3os-io/c3os/sdk => ./sdk
```
mudler, line 213: This looks like a leftover too, maybe caused by two go.mods?

sdk/clusterplugin/plugin.go
mudler: on a side note: feel to add unit tests?

Earthfile
mudler: update: addressing to those issues in a changes subset, I see this makes currently c3os fail building as go 1.19 was released and incompatible with the codebase. I'll try to open up a separate PR with bunch of fixes I had for #47 , including this

mudler: See https://github.com/c3os-io/c3os/pull/56

paynejacob: I agree with the common arg but ill leave that for your pr and remove it here.

sdk/clusterplugin/plugin.go
paynejacob: I was thinking we would build out a few plugin bases like this to encourage best practices with plugins, e.g. this is the preferred way to write cluster plugins but the framework is still flexible enough that users can write their own logic on the events if they like.  This sdk would be to c3os what kube builder is to kubernetes.

sdk/go.mod
paynejacob: Go modules support nesting like this.  It differentiates the requirements of the sdk from c3os as a whole.  It reduces the number of collisions when someone is writing an extension for c3os.  We could also solve it by moving the sdk to another repo but I prefer the sdk is released with the core framework as they are tightly coupled.

go.mod
paynejacob: This is intention, it prevents dev builds from trying to download the latest version of the sdk instead of the dev version.

sdk/clusterplugin/plugin.go
mudler: Sounds good to me :+1:

go.mod
mudler: I see now :+1: looks good, maybe just leave a comment there for reference?

mudler (approved):
This is looking good, I'd just wait for https://github.com/c3os-io/c3os/pull/56 as contains a fix for the flaky test that fails here too, otherwise it's good to merge

overlay/files/system/oem/02_agent.yaml
paynejacob: Wouldn't hooking on install mean we fire install multiple times? Ill break out this one but I don't think there are any other events we would hook on?

sdk/clusterplugin/plugin.go
paynejacob: see https://github.com/c3os-io/provider-k3s for an example

cmd/agent/main.go
paynejacob: I agree.  I think we should address error handling for plugins more broadly and make that change as part of that effort.

overlay/files/system/oem/02_agent.yaml
mudler: my intention here is to have a common file to stream all the stages events https://rancher.github.io/elemental-toolkit/docs/customizing/stages/#stages, so it becomes an alternative to move business logic out of the cloud config (if not for system level setup of things).

PR #79, opened 2022-08-09
swiched image naming scheme

paynejacob:
images now follow `quay.io/c3os/<variant>-<flavor>:<version>`.  Given framework images are only for internal use I did not change their naming schema.

### DIFF
--- a/.github/workflows/image-arm.yaml
+++ b/.github/workflows/image-arm.yaml
@@ -38,11 +38,11 @@ jobs:
       - name: Build  🔧
         env:
           FLAVOR: ${{ matrix.flavor }}
-          IMAGE: quay.io/c3os/c3os:${{ matrix.flavor }}-latest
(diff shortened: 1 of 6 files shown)


mudler:
This is looking good overall, altough I'm wondering if https://github.com/c3os-io/c3os/blob/master/internal/agent/upgrade.go#L14 would need adaptation - it ideally would check images back at the os-release file. 

We can just migrate away from parsing github releases, and we could even use git - the only issue I see there is just more machinery involved during the release process, but we can always automate it.

mudler (approved):
Let's go ahead with this and revisit the manual upgrade command later. I'd guess it's even a good candidate to have it's own event

paynejacob:
It could also be something we embed in the image: go to this github repo to find the next release.   With all the different flavors you may be upgrading from a private repo.

PR #81, opened 2022-08-10
Update arm image tag

dramich:
Update the arm image tag
Removed the push from the Build step so push can push
Removed the if check since the whole job only runs on tag

### DIFF
--- a/.github/workflows/release-arm.yaml
+++ b/.github/workflows/release-arm.yaml
@@ -53,10 +53,9 @@ jobs:
           ./earthly.sh +all-arm --IMAGE_NAME=c3os-$FLAVOR-$TAG.img --IMAGE=quay.io/c3os/core-$FLAVOR:$TAG --MODEL=$MODEL --FLAVOR=$FLAVOR
           docker push quay.io/c3os/core-$FLAVOR:$TAG
       - name: Push  🔧
-        if: startsWith(github.ref, 'refs/tags/')
         env:
           FLAVOR: ${{ matrix.flavor }}
-          IMAGE: quay.io/c3os/c3os:${{ matrix.flavor }}-latest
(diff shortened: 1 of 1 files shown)


paynejacob (approved):

paynejacob (approved):

dramich:
This PR should hold until the k3s stuff is moved out and we can verify the build succeeds

mudler:
Can you try current images from master? Images tagged with `latest` should be fine by now

.github/workflows/release-arm.yaml
```diff
@@ -53,10 +53,9 @@ jobs:
           ./earthly.sh +all-arm --IMAGE_NAME=c3os-$FLAVOR-$TAG.img --IMAGE=quay.io/c3os/core-$FLAVOR:$TAG --MODEL=$MODEL --FLAVOR=$FLAVOR
           docker push quay.io/c3os/core-$FLAVOR:$TAG
       - name: Push  🔧
-        if: startsWith(github.ref, 'refs/tags/')
         env:
           FLAVOR: ${{ matrix.flavor }}
-          IMAGE: quay.io/c3os/c3os:${{ matrix.flavor }}-latest
+          IMAGE: quay.io/c3os/c3os-${{ matrix.flavor }}:latest
```
mudler, line 58: I think this is not needed anymore there is a separate workflow pushing arm images now https://github.com/c3os-io/c3os/blob/b81e17ee026628a6cae70a6b0b1ff808145b99bf/.github/workflows/image-arm.yaml#L47

mudler (changes requested):

dramich:
Closed in favor of https://github.com/c3os-io/c3os/pull/86

PR #86, opened 2022-08-16
Update build command and docker.socket

dramich:
Added docker.socket to mounts
Removed the push from the Build step so push can push
Removed the if check since the whole job only runs on tag


mudler (approved):
Looking good, thanks!

PR #89, opened 2022-08-26
add support for json serialziation of cluster objects

paynejacob:
Cloud config is often written in yaml but go-pluggable and c3os do most of their serialization in json.  This was causing the config section of the cluster config to be lost.  I added support for serializing cluster objects as json and yaml.


mudler (approved):
Looks good!

PR #91, opened 2022-09-01
:sparkles: Boot options

mudler:
This changeset add boot options to the installation configuration. Also slightly extends the current tests to check back if the parameters was set.

Fixes https://github.com/c3os-io/c3os/issues/100


mudler:
Passing tests, merging

PR #104, opened 2022-09-09
Adding ubuntu c3os image changes, also overlay files for different pr…

vipsharm:
Ubuntu OS support and different provider overlay files.


images/Dockerfile.ubuntu
```diff
@@ -10,41 +10,45 @@ RUN apt install -y software-properties-common
...
+    debianutils \
+    curl \
+    openssh-server \
+    nano \
+    gawk \
+    haveged \
+    rsync \
+    neovim \
+    open-vm-tools \
+    netplan.io \
```
mudler, line 37: any specific reason for netplan/wireguard?

overlay/k8s/cncf/etc/k8sconfig/calico.yaml
```diff
@@ -0,0 +1,4648 @@
+---
+# Source: calico/templates/calico-config.yaml
+# This ConfigMap is used to configure a self-hosted Calico installation.
+kind: ConfigMap
```
mudler, line 4: this doesn't belong here - in the core there shouldn't be any provider specific configuration

Earthfile
```diff
@@ -226,6 +227,9 @@ docker:
         COPY overlay/files-opensuse-arm-rpi/ /
     ELSE IF [ "$FLAVOR" = "opensuse-arm-rpi" ]
         COPY overlay/files-opensuse-arm-rpi/ /
+    ELSE IF [ "$FLAVOR" = "ubuntu" ]
+        COPY overlay/files-ubuntu/ /
+        COPY overlay/k8s/$K8S_FLAVOR/ /
```
mudler, line 232: same, the overlay files here are very specific to the kubernetes engines - those changes are not meant to be here

mudler:
Also please enable ubuntu into the pipelines so we can see if they pass tests

overlay/k8s/cncf/etc/k8sconfig/calico.yaml
vipsharm: Ok, so better to move provider overlay to provider repo ? I will fix this

mudler:
the whole `overlay/k8s` folder is really specific and seems should go elsewhere, otherwise seems to go into the right direction

overlay/k8s/cncf/etc/k8sconfig/calico.yaml
mudler: yes exactly, this is provider specific

Earthfile
vipsharm: I will remove these.

images/Dockerfile.ubuntu
vipsharm: I used the OS file from PaletteOS that Venkat added for some earlier POC I believe. I will check with him if we need these.

mudler: those can be added later on when creating the image for the provider, no need to add them here

mudler, line 37:
```suggestion

```

images/Dockerfile.ubuntu
```diff
@@ -10,41 +10,45 @@ RUN apt install -y software-properties-common
...
+    wireguard \
```
mudler, line 38:
```suggestion
```

mudler:
please rebase/squash commits

images/Dockerfile.ubuntu
```diff
@@ -33,9 +33,6 @@ RUN apt install -y \
     haveged \
     rsync \
     neovim \
-    open-vm-tools \
```
mudler: can keep `open-vm-tools`, this one is for vmware guest integration

mudler: but either looks fine to me

mudler (approved):

mudler:
Looking good, thanks!

PR #135, opened 2022-09-21
Adding tmp to writable path

vipsharm:
Signed-off-by: vipsharm <sharma.vipin@gmail.com>

### DIFF
--- a/overlay/files-ubuntu/system/oem/11_persistency.yaml
+++ b/overlay/files-ubuntu/system/oem/11_persistency.yaml
@@ -1,0 +1,71 @@
+name: "Configure persistent dirs bind-mounts"
+stages:
+  rootfs.after:
+    - if: '[ ! -f "/run/cos/recovery_mode" ]'
+      name: "Layout configuration"
+      environment_file: /run/cos/cos-layout.env
+      environment:
+        VOLUMES: "LABEL=COS_OEM:/oem LABEL=COS_PERSISTENT:/usr/local"
+        OVERLAY: "tmpfs:25%"
+        RW_PATHS: "/var /etc /srv /tmp"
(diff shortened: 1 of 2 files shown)


overlay/files/etc/elemental/config.yaml
```diff
@@ -3,16 +3,16 @@ verify: false
 install:
   grub-entry-name: "kairos"
   system:
-    size: 2000
```
mudler, line 6: Why increasing size for all flavors by default? This bumps partition size for all. 

Why not applying this only on Ubuntu?

mudler (changes requested):
No need to increase default size for all flavors by default - it's only Ubuntu bigger here

overlay/files-ubuntu/system/oem/11_persistency.yaml
```diff
@@ -0,0 +1,71 @@
+name: "Configure persistent dirs bind-mounts"
+stages:
+  rootfs.after:
+    - if: '[ ! -f "/run/cos/recovery_mode" ]'
+      name: "Layout configuration"
+      environment_file: /run/cos/cos-layout.env
+      environment:
+        VOLUMES: "LABEL=COS_OEM:/oem LABEL=COS_PERSISTENT:/usr/local"
+        OVERLAY: "tmpfs:25%"
+        RW_PATHS: "/var /etc /srv /tmp"
```
mudler, line 10: this doesn't sound right. temp should be mounted by systemd

mudler:
Opened https://github.com/kairos-io/kairos/pull/136 which supersedes this PR

mudler:
https://github.com/kairos-io/kairos/pull/136 was merged, closing

PR #140, opened 2022-09-21
:lock: Secure boot

mudler:
This changeset adds Secure boot support.

QA Checklist:
- [x] bios with qemu
- [x] UEFI with qemu
- [ ] UEFI from USB stick
- [ ] BIOS from USB stick


mudler:
There is no reason why USB booting shouldn't work - I've tested both BIOS/EFI scenario and all looks good. ubuntu failures seems unrelated, tackling them separately.

PR #141, opened 2022-09-21
:sparkles: systemd-sysext and autologin

mudler:
This PR is multifold:

- Adds by default autologin with systemd
- Adds support for systemd-sysext for Live layering capabilities:
  An example can found at `ghcr.io/kairos-io/provisioning-examples:bundle-ipfs`.

How to use [systemd-sysext](https://www.freedesktop.org/software/systemd/man/systemd-sysext.html):
- Follow the standard docs, and create a bundle similar to https://github.com/kairos-io/provisioning-examples
 For testing, you can just `luet util unpack ghcr.io/kairos-io/provisioning-examples:bundle-ipfs /usr/local/lib/extensions/kubo`. Extensions can be installed with the bundle mechanism. Reboot, the ipfs binary should be now available and the `systemd-sysext` service active.


mudler:
cc @paynejacob

PR #151, opened 2022-09-27
New docs interface 

IonitaCatalina:


docs/scripts/publish.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+# dunno if this is used in netlify
```
mudler, line 4: should be good to drop it, this was used with gh-pages

docs/src/pages/quickstart/quickstart.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: "../../layouts/docs/Layout.astro"
+title: "Quickstart"
+index: 1
+---
+
+# Quickstart
```
mudler, line 7: Let's drop this - the quickstart should be top-level.

docs/src/pages/upgrade/upgrade.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: "../../layouts/docs/Layout.astro"
+title: "Upgrade"
+index: 1
+---
+
+# Installing kairos on baremetal, VMs, ...
```
mudler, line 7: Same here, this shouldn't have a section with just few sentences - the old docs have a different format that doesn't fit here

docs/src/pages/reference/reference.md
```diff
@@ -0,0 +1,7 @@
+---
+layout: "../../layouts/docs/Layout.astro"
+title: "Reference"
+index: 1
+---
+
```
mudler, line 6: ditto, this one shouldn't exist

mudler (changes requested):
Just small nits, otherwise looking good!

mudler (approved):

PR #168, opened 2022-09-30
added support for fast mode

paynejacob:
I added a setting to the agent configuration to skip the sleeps.  In the context of a user interaction the sleeps make sense however if a user has additional plugins that automate all aspects of configuration the sleeps are not needed.


mudler (approved):
Looks good!

Issue #163, opened 2022-09-29
Runtime Live layering


mudler:
**Is your feature request related to a problem? Please describe.**
Currently to add packages there are two options: or consume specially crafted bundles that installs packages into the persistent partition, or either customize the source container image and re-build the ISO with osbuilder.

**Describe the solution you'd like**
A runtime live layer mechanism. A draft for systemd based system is in #141 , but would be prefered a dracut-modules that acts similarly and extends supports to all the other distros (otherwise alpine would be left in the cold)

**Describe alternatives you've considered**
Creating custom images with osbuilder

**Additional context**
This is an additional features to complete the featureset around customization and #116

mudler:
https://github.com/kairos-io/kairos/pull/141 Covers systemd scenario - can we enable it in the initramfs and make it available for alpine? or can we either use an equivalent to it for openRC based systems?

mudler:
pretty sure we can just use systemd-sysext

mudler:
This belongs to immucore now, removing epic as it is now possible to do within immucore

mudler:
cc @Itxaka

cheese:
We need a CRD and operator to manage systemd-sysext on nodes, so we can easily apply live layering to massive nodes.

PR #174, opened 2022-10-03
:robot: Enable bundle tests on alpine

mudler:


mudler:
We will look at this again with #163

Issue #114, opened 2022-09-12
SELinux


mudler:

jmpolom:
What is the status of SELinux support with kairos built OS images? I have noticed (via the PRs mentioned here) it's explicitly disabled from the Fedora builds and `getenforce` reports disabled in the OpenSUSE Tumbleweed builds. I attempted to override the settings via kernel boot arguments in both, and both failed to boot (basically about what I expected). In the case of the Tumbleweed build, systemd halted the boot process when it couldn't load an SELinux policy.

This looks like a very interesting and useful project overall. However SELinux support is an important feature I look for in any distribution I'm considering using. It would be good to know what the barriers are to enabling SELinux when kairos is used to deploy OS images. For example: can I build a custom image with the necessary policy packages and enable SELinux post install? What breaks if I enable SELinux?

jimmykarily:
Related story: https://github.com/kairos-io/kairos/issues/2107

jmpolom:
@mudler Is selinux supported with kairos or not?

jimmykarily:
@jmpolom selinux support is planned for [3.5.0](https://github.com/kairos-io/kairos/issues/2128): https://github.com/kairos-io/kairos/issues/111

That said, what you describe it what needs to happen. The right policies must be in place. Then selinux can probably be enabled and it should work. There is nothing inherent in Kairos that prevents selinux from working afaik.

If you make any progress on this and make it work, please share your results with us. Even if only for a specific distro, it will make the work for the rest of them a lot easier.

imusmanmalik:
@jimmykarily I see that now `3.5.0` is released but i don't see SELinux support as part of it.

Is it still planned to support SELinux as it is disabled by default in the current state for RPM based OS.

jimmykarily:
@imusmanmalik is has been re-prioritized to 3.8.0 since a while now: https://github.com/kairos-io/kairos/issues/2133

Sorry for the confusion.

jmpolom:
Not to nag, but is there a barrier that is resulting in the continual delay with SELinux support across kairos flavors? Is it just a matter of insufficient demand or are there known technical blockers? (and a link to issues for those would be beneficial, if there are any)

jimmykarily:
> Not to nag, but is there a barrier that is resulting in the continual delay with SELinux support across (relevant) kairos flavors? Is it just a matter of insufficient demand or are there known technical blockers? (and a link to issues for those would be beneficial, if there are any)

[...]

PR #189, opened 2022-10-06
:penguin: Disable temporarly SELinux on fedora

### LINKED ISSUE #63: fedora flavor
Currently the image definition is there, but lacks few things in order to boot properly. This card is to make the fedora flavor functional and into releases.

mudler:
SELinux support have to be handled separately in https://github.com/kairos-io/kairos/issues/114. fedora based version has selinux, which is enabled by default but we don't have support for it yet.

Fixes #63 . Now fedora installs and boots correctly afterwards


mudler:
Fix #63. rockylinux seems to have stacktrace that was too fast to catch, will look into separately

PR #223, opened 2022-10-15
Drop wicked, long live to systemd-networkd

mudler:
This PR:
- Drops wicked references from openSUSE images
- Adds systemd-networkd settings to initramfs


overlay/files/etc/dracut.conf.d/99-kairos.conf
```diff
@@ -0,0 +1 @@
+install_items+=" /etc/systemd/network/20-dhcp.network "
```
mudler, line 1: This is best to go as a package so our initrds are aligned as well, but best for a follow-up

PR #242, opened 2022-10-18
✨ Added /etc/kubernetes to persistent path list

kpiyush17:
This change is in accordance with the support for the kubeadm provider.


mudler (approved):
Looking good, thanks!

PR #245, opened 2022-10-19
change the automated installation http link in container.md page to correctly point to the automated installation page

vfiftyfive:
In the architecture/container section of the documentation, the http link pointing to the automated installation page is incorrect, there's a leading slash missing.

Nic


mudler (approved):
Thanks! Good catch!

PR #260, opened 2022-10-22
✨ add env to config to use across all stages

### LINKED ISSUE #259: elemental process does not have context of env variables set in Config.Install.Env in reset
**Kairos version:**
1.1.3

x86_64, core alpine 1.1.3

**Describe the bug**
elemental process does not have access to env variables set in Config.Install.Env in reset
[here](https://github.com/kairos-io/kairos/blob/ca53fe6c13544610022e7259debfc7a5572c5e4d/internal/agent/reset.go#L73)
Elemental will fail to pull images during reset as it would not have access to http proxy variables set in Config.Install.env. The values in Config.Install.env needs to be set in the environment before kairos reset is triggered.

**To Reproduce**
set the http proxy environment variable in Config.Install.Env and run kairos reset

**Expected behavior**
kairos reset should be able to work successfully in proxy environments by pulling in the env vars set in Config.Install.env

santhoshdaivajna:
Adds support for elemental process to have access to env variables set in Config.Install.Env in reset

Fixes #259

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -10,9 +10,12 @@ import (
 	machine "github.com/kairos-io/kairos/pkg/machine"
 	"github.com/kairos-io/kairos/pkg/utils"
 	bundles "github.com/kairos-io/kairos/sdk/bundles"
+	"github.com/kairos-io/kairos/sdk/state"
 
 	"github.com/urfave/cli"
 )
+
+var configScanDir = []string{"/oem", "/usr/local/cloud-config", "/run/initramfs/live"}
 
 var cmds = []cli.Command{
 	{
@@ -73,8 +76,7 @@ See https://kairos.io/after_install/upgrades/#manual for documentation.
 			if len(args) == 1 {
 				v = args[0]
 			}
-
-			return agent.Upgrade(v, c.String("image"), c.Bool("force"), c.Bool("debug"))
+			return agent.Upgrade(v, c.String("image"), c.Bool("force"), c.Bool("debug"), configScanDir)
 		},
 	},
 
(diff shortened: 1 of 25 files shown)


mudler (changes requested):
let's have a ` reset.env` in the config file only for reset so it is more meaningful and we avoid having unexpected behaviors - I'd be suprised to see an `install` setting applied to a `reset` operation. we could think maybe to have a generic `env` block which applies to all.

mudler:
This is not a bug - please use another icon (:sparkles: sounds better, as you are extending a current functionality)

internal/agent/reset.go
```diff
@@ -70,6 +73,17 @@ func Reset() error {
 		args = append(args, "--reset-persistent")
 	}
 
+	cloudInit, ok := options["cc"]
```
mudler, line 76: this is not going to work - config during reset is in the system, so to retrieve it you have to scan it in the system like https://github.com/kairos-io/kairos/blob/ca53fe6c13544610022e7259debfc7a5572c5e4d/internal/agent/notify.go#L15

internal/agent/upgrade.go
```diff
@@ -64,7 +67,14 @@ func Upgrade(version, image string, force, debug bool) error {
 		discoveredImage = r.Data
 	})
 
-	_, err := bus.Manager.Publish(events.EventVersionImage, &events.VersionImagePayload{
+	options := map[string]string{}
```
mudler, line 70: ditto, this is not correct - you have to scan the system to get the config, options are just returned by providers for additional args injection

mudler (changes requested):
Overall looks good - however how config is retrieved is invasive and not correct - we have to rather scan the system and not expect plugins to return configs at this stage (also, this would break current sdk contract)

cmd/agent/main.go
```diff
@@ -73,8 +73,8 @@ See https://kairos.io/after_install/upgrades/#manual for documentation.
 			if len(args) == 1 {
 				v = args[0]
 			}
-
-			return agent.Upgrade(v, c.String("image"), c.Bool("force"), c.Bool("debug"))
+			dirs := []string{"/oem", "/usr/local/cloud-config", "/run/initramfs/live"}
```
mudler, line 76: what about making this a var so we avoid the code dup in the file?

PR #264, opened 2022-10-24
✨ add env to config to use across all stages 

santhoshdaivajna:
What this PR does / why we need it:
Adds support for elemental process to have access to env variables set in Config.Install.Env in reset

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes https://github.com/kairos-io/kairos/issues/259

linking the earlier PR here [260](https://github.com/kairos-io/kairos/pull/260)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ docs/public
 docs/.vscode
 
 .DS_Store
+
+# editor and IDE paraphernalia
+.idea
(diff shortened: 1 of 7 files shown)


mudler (approved):
Looking good!

PR #266, opened 2022-10-24
:penguin: Use ubuntu kernel with alpine flavor

mudler:

Fixes #


mudler:
This is kinda of working - however it looks like it doesn't take cloud config. might be that the initrd is missing the elemental cli

mudler:
seems rc services doesn't start after boot, but everything else seems to work

mudler:
What was missing: rsync and gpt disk utils were not installed when generating the initramfs - that yield dracut to not bundle those in the initramfs, thus `/etc/` content was not copied over.

PR #320, opened 2022-10-29
✨ add iptables and conntrack to ubuntu images

kpiyush17:
**What this PR does / why we need it**: This PR adds iptables and conntrack tools to ubuntu images. This is needed as a requirement of kubeadm to deploy k8s clusters using provider.


mudler (approved):

mudler:
I'd rather prefer packages to be added to the provider repositories - however those sounds core packages that should be in the base image. Thanks!

PR #338, opened 2022-10-30
:seedling: Add Better Windows Desktop Support to Kairos

luthermonson:
## What this PR does / why we need it
Adds additional Windows desktop support to the project. 

If anyone finds this later by searching, please note this was on win11 with docker for desktop. If you install earthly.exe to your PATH everything works fine if you do `earthly +test` in powershell if you want to try different versions of earthly. The ps1 will pin the specific version of earthly kairos is currently using and testing against in CI and runs all commands in a linux container.

## .gitattrbutes file with *.sh eol=lf
fixes issues with wsl when you try running .sh. default checkouts with git for windows will let windows manage the eol and set everything to CRLF. When you try and run .sh you get the following.
```
ubuntu@DESKTOP-F54SPNC:/mnt/c/Users/luthe/kairos/kairos$ ./earthly.sh
-bash: ./earthly.sh: /bin/bash^M: bad interpreter: No such file or directory
```
adding the gitattributes file will force git to make all .sh files eol be a linefeed and this will make life easier for any one new to the project on a windows machine

## adding earthly.ps1
added an earthly.ps1 file which is identical to the earthyl.sh file but in powershell so you can natively run the docker command and not have to switch to wsl. please note, the docker socket mount was verified to work as well as either --privileged or the -v were removed the following appears.

Without `--privileged`
```
~\kairos\kairos [windows] > .\earthly.ps1 +test
DOCKER_HOST is not set
Container appears to be running unprivileged. Currently, privileged mode is required when buildkit runs inside the container.
```
Without `-v /var/run/docker.sock:/var/run/docker.sock`
```
~\kairos\kairos [windows] > ./earthly.ps1 +test                       
DOCKER_HOST is not set
Docker appears not to be connected. Please check your DOCKER_HOST variable, and try again.
```

## go mod
my version of go1.19.2 found some mod changes it apparently wanted to fix so I let it, if you want this backed out just let me know.

Fixes #

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ docs/node_modules
 docs/dist
 docs/public
 docs/.vscode
-
+coverage.out
 .DS_Store
 
 # editor and IDE paraphernalia
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,9 +98,16 @@ Make sure that all tests and lint checks are passing before submitting a pull re
 
 You can run the lint and test checks locally with:
 
+### Linux
 ```bash
 ./earthly.sh +lint
 ./earthly.sh +test
+```
+
+### Windows
+```bash
+./earthly.ps1 +lint
+./earthly.ps1 +test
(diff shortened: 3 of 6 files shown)


mudler (approved):
Looking good here! thanks!

PR #353, opened 2022-11-01
🐛  Enable Open VM Tools on boot

3pings:
Signed-off-by: Justin Barksdale <3pings@users.noreply.github.com>

Open VM Tools is installed but it is not enabled.  This PR will enable open-vm-tools on boot.

Fixes # N/A


mudler (approved):
Looking good! Thanks!

PR #368, opened 2022-11-03
:bug: Do not run kcrypt hook without a oem mount

### LINKED ISSUE #358: kcrypt hook fails on ARM
We should skip this hook when there are no OEM partitions, otherwise it just prints on screen errors (which are not fatal)
https://github.com/kairos-io/kairos/blob/2efc9aeca8e4e6c123e70bb1ea9e6dca523b9001/overlay/files/system/oem/21_kcrypt.yaml#L8

mudler:
**What this PR does / why we need it**: It runs the hook only if there is a OEM partition accessible. It uses the state api to get the state and the device name programmatically

Fixes #358

### DIFF
--- a/overlay/files/system/oem/21_kcrypt.yaml
+++ b/overlay/files/system/oem/21_kcrypt.yaml
@@ -2,10 +2,11 @@ name: "Update discovery plugins"
 stages:
     after-upgrade:
     - name: "Update plugins"
+      if: "[ $(kairos-agent state get oem.found) == 'true' ]"
       commands:
       - |
           STATEDIR=/tmp/mnt/OEM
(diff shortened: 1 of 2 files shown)


overlay/files/system/oem/21_kcrypt.yaml
```diff
@@ -2,10 +2,11 @@ name: "Update discovery plugins"
 stages:
     after-upgrade:
     - name: "Update plugins"
+      if: "[ $(kairos-agent state get oem.found) == 'true' ]"
```
mudler, line 5: this actually gates the block if there is no oem partition found (ARM, for instance)

PR #381, opened 2022-11-04
✨ directories added to support - update-ca-certificates

santhoshdaivajna:
Typically _update-ca-certificates_ is executed to update the list of root CA certificates.  update-ca-certificates reads the added source certificates from _/usr/share/pki/trust, /usr/share/pki/trust/anchors_ and updates the files in _/var/lib/ca-certificates and /etc/ssl/certs_


mudler (approved):
Good catch, also just in time for the release window!

mudler:
@santhoshdaivajna commits needs to be signed off. cc @jimmykarily we need to update the contribution guidelines to reflect that we adopt DCO as well

Issue #377, opened 2022-11-03
Expose qcow/cloud disk images creation from the controller


mudler:
**Is your feature request related to a problem? Please describe.**
a way to build qcow and cloud images

**Describe the solution you'd like**
Have our osbuilder controller to be able to create cloud images

**Describe alternatives you've considered**

**Additional context**
Things should be already there - as scripts and tooling to do that, needs just to be tested out and properly exposed via the controller

mudler:
this is working now, @jimmykarily I'll close this one and prepare docs for it

PR #382, opened 2022-11-04
:book: Add advanced/build section

mudler:
Follow up of #377


mudler:
@jimmykarily merging so we have a base for docs, let's iterate on it later (it needs to add also a bit more on the API spec itself :) )

Issue #210, opened 2022-10-12
:seedling: Custom mounts


mudler:
**Is your feature request related to a problem? Please describe.**
Actually the persistent configuration is static and it cannot be extended. In order to configure additional mounts or bind mounts it is required to provide a cloud-config including directives as : https://github.com/kairos-io/kairos/blob/master/overlay/files/system/oem/11_persistency.yaml which replaces the system ones

**Describe the solution you'd like**
A way to extend what was declared by the base image - via the Kairos config file

**Describe alternatives you've considered**
N/A

**Additional context**

oz123:
The following cloud init should work with the linked commit:
```
#cloud-config
users:
 - name: kairos 
   passwd: "kairos"
   ssh_authorized_keys:
     - github:oz123
   groups:
     - admin

additional_persistent_mounts:
  - /data/bla
  - /foo/bar
```

However, it seems that my build-kairos agent isn't surving a reboot. Hence, I have no clue if my code works, yet.

mudler:
Closing, https://github.com/kairos-io/kairos/pull/692 is merged now

PR #407, opened 2022-11-10
:penguin: Add support for snapd

### LINKED ISSUE #401: run snapd
Would be nice if we could actually run snapd on the system and be able to install snaps. This would allow to install packages on an immutable system easily in runtime/deployment, and manage it afterward on day2

This is a spike card, there is no strings attached to a specific solution

mudler:

Note: This is creating snap folders for all flavors (`/var/lib/snap` and `/snapd`) - I'd prefer we inject persistent paths if we detect snapd but we don't have #210 yet, so it would get quite hairy. Once we get #210 we can get back at that and programmatically add mount points/paths if we detect bins

Fixes #401

### DIFF
--- a/images/Dockerfile.ubuntu
+++ b/images/Dockerfile.ubuntu
@@ -34,6 +34,7 @@ RUN apt install -y \
     jq \
     neovim \
     open-vm-tools \
+    snapd \
     conntrack \
     iptables \
     linux-image-generic-hwe-22.04 && apt-get clean
--- a/overlay/files/system/oem/11_persistency.yaml
+++ b/overlay/files/system/oem/11_persistency.yaml
@@ -13,7 +13,6 @@ stages:
           /etc/modprobe.d
           /etc/rancher
           /etc/sysconfig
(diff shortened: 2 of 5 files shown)


images/Dockerfile.ubuntu
```diff
@@ -34,6 +34,7 @@ RUN apt install -y \
     jq \
     neovim \
     open-vm-tools \
+    snapd \
```
mudler, line 37: I've checked, and this adds around ~30MB, I think that's a good compromise and add it in the image - no small libs, just snapd gets in

overlay/files/system/oem/11_persistency.yaml
```diff
@@ -13,6 +13,7 @@ stages:
           /etc/modprobe.d
           /etc/rancher
           /etc/sysconfig
+          /etc/profile.d
```
mudler, line 16: I think we don't even need that but I've added while testing - I'll double check

jimmykarily (approved):

Issue #380, opened 2022-10-17
Traceback label with disk partition


mudler:
Currently https://github.com/kairos-io/kcrypt-challenger is the component that sits on Kubernetes and gives out the passwords. It also contains the TPM client side.

the https://github.com/kairos-io/kcrypt is the low level part of the stack, which sits into the initramfs and decrypts the fs.

```yaml
apiVersion: keyserver.kairos.io/v1alpha1
kind: SealedVolume
metadata:
    name: test2
    namespace: default
spec:
  TPMHash: "fff"
  partitionSecrets:
    COS_PERSISTENT:
      name: mysecret
      path: pass
    unknown:
      name: mysecret
      path: pass
  quarantined: **false
```

This card is about the issue that "unknown" would decrypt "COS_PERSISTENT" because the client doesn't know which labels the partitions are since they are encrypted during boot. this is a workaround, so the card is about to make the state saved somewhere so we can pick it up later and avoid supplying the same information twice.

PR #408, opened 2022-11-10
Persist partition label/uuid mapping whenever we encrypt a partition

jimmykarily:
This depends on a change in kcrypt to return a proper string after encrypting a partition: https://github.com/kairos-io/kcrypt/pull/3

Part of: https://github.com/kairos-io/kairos/issues/380


internal/agent/hooks/kcrypt_test.go
```diff
@@ -0,0 +1,102 @@
+package hook
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestParsePartitionInfoFile(t *testing.T) {
```
mudler, line 9: I'm totally fine with this, but just wondering why not using ginkgo also here?

internal/agent/hooks/kcrypt.go
```diff
@@ -54,3 +74,46 @@ func (k Kcrypt) Run(c config.Config) error {
...
+		return err
+	}
+
+	parts := strings.Split(partitionData, ":")
+	partitionInfo[parts[0]] = parts[2]
+
+	return updatePartitionInfoFile(partitionInfo, file)
+}
+
+func parsePartitionInfoFile(file string) (PartitionInfo, error) {
```
mudler, line 94: This is all good, nice! I'm just wondering here - as the challenger will need to re-read the same file, if makes sense to make those public in a separate package dedicated for it instead

internal/agent/hooks/kcrypt_test.go
jimmykarily: Since ginkgo wasn't there already, I thought it would be simpler without it. Turns out ginkgo tests are a lot more readable :D . I will switch to ginkgo for consistency across the projects (and for readability)

internal/agent/hooks/kcrypt.go
jimmykarily: Moved to kcrypt: https://github.com/kairos-io/kcrypt/pull/3/commits/82c6e8fcd092bd05aeda8fd5ef292965957ee95f

I'll import it here as a module.

internal/agent/hooks/kcrypt_test.go
jimmykarily: Migrated to ginkgo here: https://github.com/kairos-io/kcrypt/pull/3/commits/82c6e8fcd092bd05aeda8fd5ef292965957ee95f

internal/agent/hooks/kcrypt.go
```diff
@@ -37,6 +39,21 @@ func (k Kcrypt) Run(c config.Config) error {
 			time.Sleep(10 * time.Second)
 			return nil // do not error out
 		}
+
+		partitionInfo, err := pi.NewPartitionInfoFromFile(pi.DefaultPartitionInfoFile)
```
jimmykarily, line 43: The first time this code runs, there will be no partition info file. I'm thinking this `NewPartitionFromFile` function should create the partition file if it doesn't exist and return and empty `PartitionInfo` object (with no mappings). I'll add a boolean return value to the function to indicate whether the file existed or not. This way the caller can ignore this value if it doesn't care. I'll open a PR on kcrypt to do that.

jimmykarily: https://github.com/kairos-io/kcrypt/pull/4

jimmykarily: done

mudler (approved):
Looking good!

jimmykarily:
It seems to work fine now after various fixes on the kcrypt side.

PR #434, opened 2022-11-15
:arrow_up: Bump repository and drop some workarounds

mudler:
**What this PR does / why we need it**: Alpine currently fails booting, this set of changes fixes boot and also drops workarounds

Fixes #

### DIFF
--- a/sdk/system/grub.go
+++ b/sdk/system/grub.go
@@ -37,7 +37,7 @@ func setGRUBOptions(opts map[string]string) error {
 	}()
 
 	for k, v := range opts {
-		out, err := utils.SH(fmt.Sprintf(`grub2-editenv /tmp/oem/grubenv set "%s=%s"`, k, v))
+		out, err := utils.SH(fmt.Sprintf(`%s /tmp/oem/grubenv set "%s=%s"`, machine.FindCommand("grub2-editenv", []string{"grub2-editenv", "grub-editenv"}), k, v))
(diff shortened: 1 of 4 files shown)


mudler:
Supersedes #433

mudler:
Can't remove symlink due to other breakage that would likely happen (e.g. boot assessment would stop working). Opened a follow-up #435

sdk/system/grub.go
```diff
@@ -37,7 +37,7 @@ func setGRUBOptions(opts map[string]string) error {
 	}()
 
 	for k, v := range opts {
-		out, err := utils.SH(fmt.Sprintf(`grub2-editenv /tmp/oem/grubenv set "%s=%s"`, k, v))
+		out, err := utils.SH(fmt.Sprintf(`%s /tmp/oem/grubenv set "%s=%s"`, machine.FindCommand("grub2-editenv", []string{"grub2-editenv", "grub-editenv"}), k, v))
```
mudler, line 40: some distros have grub-install, or grub-editenv instead of grub2-editenv and grub2-install. Hence some logic to detect that. note we can't drop symlinks still due to #435

mudler:
@jimmykarily I've tried to drop the grub symlinks in the images here but still can't due to https://github.com/kairos-io/kairos/issues/435

jimmykarily (approved):

PR #439, opened 2022-11-16
:book: Add slack link

mudler:
Signed-off-by: Ettore Di Giacinto <mudler@mocaccino.org>

**What this PR does / why we need it**: as there might be users interested in joining only with Slack as well. A bridge will be created too


mudler:
@saamalik is taking care of linking with our Matrix room

mudler:
Bridges are acting weirdly - we have tried it out but messages from slack to matrix doesn't work. Keeping only matrix for now, and work out ok a better doc and ease out joining from the browser too.

PR #445, opened 2022-11-17
:bug: Setup / as shared mount on alpine

### LINKED ISSUE #440: `/var/lib/longhorn` is not shared mounted
**Kairos version:**
1.1.7-alpine

**Describe the bug**
Alpine can't be installed as `/var/lib/longhorn` seems to not be shared mounted

**To Reproduce**

**Expected behavior**

**Logs**
```
Warning  Failed          2m4s                kubelet            Error: failed to generate container "8424cf1a165ad1928a20fff7f65e863896018aedb1d8cbf47437b7dd8e7b6cee" spec: failed to generate spec: path "/var/lib/longhorn/" is mounted on "/var/lib/longhorn" but it is not a shared mount
```

**Additional context**

mudler:
Since this is not handled in the boot process, we do this in our devoted file to setup the rootfs configuration. It also adds an e2e test to verify that the mounts are correctly set

Signed-off-by: Ettore Di Giacinto <mudler@kairos.io>

Fixes #440 

Thanks @christianprim for the finding!

### DIFF
--- a/tests/autoinstall_test.go
+++ b/tests/autoinstall_test.go
@@ -157,6 +157,12 @@ var _ = Describe("kairos autoinstall test", Label("autoinstall-test"), func() {
 			Expect(out).To(ContainSubstring("/var/lib/longhorn"))
 		})
 
+		It("has rootfs shared mount", func() {
+			out, err := Sudo(`cat /proc/1/mountinfo | grep ' / / '`)
+			Expect(err).ToNot(HaveOccurred(), out)
+			Expect(out).To(ContainSubstring("shared"))
+		})
+
 		It("has corresponding state", func() {
 			out, err := Sudo("kairos-agent state")
 			Expect(err).ToNot(HaveOccurred())
--- a/overlay/files-alpine/system/oem/23_rootfs.yaml
+++ b/overlay/files-alpine/system/oem/23_rootfs.yaml
@@ -3,3 +3,4 @@ stages:
   boot.before:
     - commands:
       - mount -o mode=1777,nosuid,nodev -t tmpfs tmpfs /tmp
+      - mount --make-rshared /
--- a/pkg/utils/system.go
+++ b/pkg/utils/system.go
@@ -52,7 +52,7 @@ func Flavor() string {
 
 func IsOpenRCBased() bool {
 	f := Flavor()
-	return f == "alpine" || f == "alpine-arm-rpi"
+	return strings.Contains(f, "alpine")
 }
 
 func GetInterfaceIP(in string) string {


tests/autoinstall_test.go
```diff
@@ -156,6 +156,12 @@ var _ = Describe("kairos autoinstall test", Label("autoinstall-test"), func() {
 			Expect(out).To(ContainSubstring("/var/lib/longhorn"))
 		})
 
+		It("has rootfs shared mount", func() {
+			out, err := Sudo(`cat /proc/1/mountinfo | grep " / / "`)
+			Expect(err).ToNot(HaveOccurred())
```
jimmykarily, line 161:
```suggestion
			Expect(err).ToNot(HaveOccurred(), out)
```
in case things break, this will provide more info

jimmykarily (dismissed):

tests/autoinstall_test.go
mudler: good catch! thanks!

jimmykarily (dismissed):

christianprim:
Considering the output of "qemu-acceptance-tests (alpine) ", the command "mount --make-rshared /sysroot" failed in the fs stage. I wonder why not use "mount --make-rshared /" when root-partition is under /. Since the command recursively changes the type of all mounts, this should do the trick. Could be the last command before starting k3s...

mudler:
> Considering the output of "qemu-acceptance-tests (alpine) ", the command "mount --make-rshared /sysroot" failed in the fs stage. I wonder why not use "mount --make-rshared /" when root-partition is under /. Since the command recursively changes the type of all mounts, this should do the trick. Could be the last command before starting k3s...

right, what I'm concerned here is try to break user-contract, and allow to have that set as soon as possible to avoid any possible race - and user assumption when it can be expected those to be set. I'd prefer to have those as early as possible (as soon as a fs is set up) but that seems to be tricky. I'll first try to get a baseline and we can revisit this later on.

christianprim:
> right, what I'm trying to be careful here is to not break user-contract, and allow to have that set as soon as possible to avoid any possible race - and user assumption when it can be expected those to be effectively ready in the system. I'd prefer to have those as early as possible (as soon as a fs is set up) but that seems to be tricky. I'll first try to get a baseline and we can revisit this later on.

Good point!

christianprim:
Just for your information: mounts are shared when the command is executed in the network stage (did this for testing purposes in  /oem/90_custom.yaml). But this may be very late in the boot process... [partition 2, 4 and 5 are remounted rather late in the boot-process, especially partition 4]

PR #450, opened 2022-11-18
:seedling: Don't build image files for arm, just container images

### LINKED ISSUE #383: openSUSE ARM image exceed release size
ARM images are exceeding release size: https://github.com/kairos-io/kairos/actions/runs/3394246214

## Action items

Stop building ARM images as part of our releases, as ARM images involve anyhow usage beside virtualization - so it doesn't help the user anyhow, but instead document on how to pull those manually:

- [x] Provide step by step instruction to pull an image from the released artifacts, extract the raw image, and dd that (raw images get pushed also as container images)
- [ ] Drop Github ARM release jobs from pipelines - mind to keep releases to push container images containing the raw image file used to document the process

jimmykarily:
Fixes #383

### DIFF
--- a/docs/content/en/docs/Reference/image_matrix.md
+++ b/docs/content/en/docs/Reference/image_matrix.md
@@ -23,6 +23,20 @@ The installable mediums part of the releases are generated with the methods desc
 | **openSUSE based (RaspberryPi 3 and 4, arm64)**     	| https://quay.io/repository/kairos/core-opensuse-arm-rpi 	| https://quay.io/repository/kairos/kairos-opensuse-arm-rpi 	|
 | **Alpine Linux based (RaspberryPi 3 and 4, arm64)** 	| https://quay.io/repository/kairos/core-alpine-arm-rpi   	| https://quay.io/repository/kairos/kairos-alpine-arm-rpi   	|
 
-<Info>
+{{% alert title="Note" color="info" %}}
 ** the `ubuntu` flavor tracks the latest available Ubuntu release (at the time of writing 22.10). the LTS flavors instead are tracking the latest LTS available in dockerhub. i.e. ubuntu-22-lts uses 22.04 as base image
-</Info>
+{{% /alert %}}
+
+{{% alert title="Note" color="info" %}}
+  The pipelines don't publish `img` artifacts for arm architecture because the files are too big for Github Actions (they are above the artifact size limit).
+  They can be extracted from the published docker images with the following command:
+
+  ```bash
+  export IMAGE=quay.io/kairos/core-alpine-arm-rpi:v1.1.7.img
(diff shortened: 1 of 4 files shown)


mudler:
Just a thought maybe we can keep ARM images for core as they are smaller and drop them only from the provider? or that might be complex to document and dive around? My only fear is that we might get into this again, and treating all of them in the same way it is far easier. J

mudler:
> Just a thought maybe we can keep ARM images for core as they are smaller and drop them only from the provider? or that might be complex to document and dive around? My only fear is that we might get into this again, and treating all of them in the same way it is far easier.

forget that - it affects both core and provider, we need just to propagate down the line this to kairos-io/provider-kairos

mudler (approved):

mudler:
I think we should just wait for the ARM CI to pass, no need to wait for x86 tests

docs/content/en/docs/Reference/image_matrix.md
```diff
@@ -25,4 +25,18 @@ The installable mediums part of the releases are generated with the methods desc
 
 <Info>
 ** the `ubuntu` flavor tracks the latest available Ubuntu release (at the time of writing 22.10). the LTS flavors instead are tracking the latest LTS available in dockerhub. i.e. ubuntu-22-lts uses 22.04 as base image
-</Info>
\ No newline at end of file
+</Info>
+
+<Info>
```
mudler, line 30: argh, my bad - we need to remove the <info> here, that was an oversight when switching from astro to docsy. The correct syntax is https://github.com/kairos-io/kairos/blob/master/docs/content/en/docs/Advanced/build.md?plain=1#L8

mudler: https://www.docsy.dev/docs/adding-content/shortcodes/#alert

docs/content/en/docs/Reference/image_matrix.md
```diff
@@ -23,6 +23,20 @@ The installable mediums part of the releases are generated with the methods desc
...
 
-<Info>
+{{% alert title="Note" color="info" %}}
 ** the `ubuntu` flavor tracks the latest available Ubuntu release (at the time of writing 22.10). the LTS flavors instead are tracking the latest LTS available in dockerhub. i.e. ubuntu-22-lts uses 22.04 as base image
-</Info>
\ No newline at end of file
+{{% /alert %}}
+
+{{% alert title="Note" color="info" %}}
+  The pipelines don't publish `img` artifacts for arm architecture because the files are too big for Github Actions (they are above the artifact size limit).
```
mudler, line 31: should we reference to that also on the raspberry page? The current docs guide to downloading assets https://kairos.io/docs/installation/raspberry/

mudler (approved):
Looking good!

Issue #116, opened 2022-09-12
:building_construction: Kairos factory - Streamline derivatives creation


mudler:
This card is the epic to track all the issues related to streamlining derivatives creation. This includes all the tools, design document, flows in place in order to get full OS image building easied out and also carried in Kubernetes native contexts, including onboarding.

The ultimate goal is to be able to convert upstream images directly with our CRD and artifacts built from our factory - if someone wants can build them manually or just use ours.

## Design goals
- Be able to create a derivative from a Dockerfile with standard container build tools (buildah, docker, etc.)
- Be able to create a derivative from a Kubernetes manifest CRD, using also artifacts from the Kairos factory in a "static", immutable way. Instead of running dynamic commands, we focus on statically creating high reproducible builds of images.
- Be able to create custom OS from Kairos core releases from the CRD and a dedicated CLI with additional customizations: such as kernel, packages, etc.
- Be able to create assets to boot the OS in different scenarios: raw disk images for cloud, ISOs for baremetal, disk images for embedded platforms, netboot artifacts for booting from network
- Be able to netboot seamlessly from network with zero-touch infrastructure configuration - plugging into the existing network with our CRDs.
- Be able to create your own Kairos factory artifacts
- Be able to SBOM, sign, and trace all the artifacts in #92 and #51 

--- 

MVP:
- [x] #270
- [x] #272
- [x] #297
- [x] #300
- [x] CRD to generate custom OS images. input: cloud-init, set of bundles, a base image https://github.com/c3os-io/osbuilder
- [x] #299
- [x] #377 
- [x] #296
- [x] #273
- [x] #515
- [x] https://github.com/kairos-io/kairos/issues/777
- [x] #298
- [x] #514
- [x] #755
- [x] #1044 
- [x] https://github.com/kairos-io/kairos/issues/776
- [x] https://github.com/kairos-io/osbuilder/pull/22
- [x] https://github.com/kairos-io/osbuilder/pull/23
- [x] https://github.com/kairos-io/kairos/issues/711
- [x] #513
- [x] https://github.com/kairos-io/kairos/issues/932
- [x] https://github.com/kairos-io/kairos/issues/1042

Deliverables:
- [x] https://github.com/kairos-io/kairos/issues/430
- [x] https://github.com/kairos-io/kairos/issues/712
- [x] #269
- [x] #301
- [ ] redesign osbuilder CRDs for v1
- [x] Create provider packages in our luet repositories, that can be consumed as a bundles during installation
[...]

PR #457, opened 2022-11-18
:penguin: Split alpine images into dedicated channels

mudler:
Depending on the kernel being used, we build now 2 distinct images (alpine-opensuse-leap and alpine-ubuntu).

**What this PR does / why we need it**: This is needed so there is a clear distinction on the kernels that are part of the alpine images. Now the alpine flavor got split into two, one featuring ubuntu and one the opensuse one until #116 is done and would allow to programmatically load providers and external binaries.


jimmykarily (approved):

saamalik (approved):

mudler:
The broken test now is just trying to fetch the latest release - but since this is a new flavor it fails. Going to merge and once tagged this should be green

PR #464, opened 2022-11-19
:book: Add remote reset examples

mudler:

### DIFF
--- a/docs/content/en/docs/Reference/reset.md
+++ b/docs/content/en/docs/Reference/reset.md
@@ -6,21 +6,112 @@ date: 2022-11-13
 description: >
 ---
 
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/Reference/reset.md
```diff
@@ -20,7 +111,7 @@ It is possible to reset the state of a node by either booting into the "Reset" m
 
 It is possible to trigger the reset manually by logging into the recovery from the boot menu and running `kairos reset` from the console.
```
jimmykarily, line 112: The sections in the description at the top don't match the sections below:

```
via the Boot menu, executed manually, or triggered via Kubernetes
```

and the sections:

- Remotely, via command line
- From Kubernetes
- From the boot menu
- Manual reset

I would change the order to match the description and add the "manual reset" in the description too.

mudler: good point, fixed!

jimmykarily (approved):

PR #483, opened 2022-11-24
Use kvm when running qemu to make things faster

jimmykarily:

### DIFF
--- a/scripts/start_vm_qemu.sh
+++ b/scripts/start_vm_qemu.sh
@@ -4,10 +4,13 @@ if [ ! -e disk.img ]; then
     qemu-img create -f qcow2 disk.img 40g
 fi
 
+[[ -n "${ENABLE_KVM}" ]] && KVM=(-enable-kvm)
+
 qemu-system-x86_64 \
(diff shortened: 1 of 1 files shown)


scripts/start_vm_qemu.sh
```diff
@@ -8,10 +8,11 @@ qemu-system-x86_64 \
     -m ${MEMORY:=2096} \
     -smp cores=2 \
     -nographic \
+    -enable-kvm \
```
mudler, line 11: care to add it as optional? I use it in codespaces and there is no kvm in there :see_no_evil:

mudler: enabled by default sounds perfectly fine too. we can also set an env var directly when creating the devcontainer here to turn off KVM https://github.com/kairos-io/kairos/blob/master/.devcontainer/post-create.sh

but yeah, don't want it to make it complex... otherwise we could also detect if we have /dev/kvm?

jimmykarily: done

mudler (approved):
Nais!

Issue #491, opened 2022-11-25
Header preserved only when reading files


mudler:
https://github.com/kairos-io/kairos/blob/e2b42a0f9e524102c06041672dbf3ecfcd74d97c/pkg/config/config.go#L153

We should preserve the header also when fetching from `config_url`, or via cmdline input we should assume "#cloud-config"

PR #492, opened 2022-11-25
:bug: Preserve header when fetching config via URL

mudler:
Fixes: https://github.com/kairos-io/kairos/issues/491

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -53,6 +53,8 @@ type Bundle struct {
 	Targets []string `yaml:"targets,omitempty"`
 }
 
+const DefaultHeader = "#cloud-config"
+
 func HasHeader(userdata, head string) (bool, string) {
 	header := strings.SplitN(userdata, "\n", 2)[0]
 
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -211,6 +213,16 @@ func Scan(opts ...Option) (c *Config, err error) {
 
 		yaml.Unmarshal(body, c)               //nolint:errcheck
 		yaml.Unmarshal(body, &c.originalData) //nolint:errcheck
+
+		if exists, header := HasHeader(string(body), ""); c.IsValid() || exists {
```
jimmykarily, line 217: `c.IsValid()` has no effect here. If `exists` is true, then we enter the block, if it's not then we may enter the block but the block will do nothing. Not sure why `IsValid` is called here.

mudler: right, good catch, totally missed it!

pkg/config/config_test.go
```diff
@@ -108,5 +130,24 @@ config_url: "https://gist.githubusercontent.com/mudler/ab26e8dd65c69c32ab2926857
 			Expect(c.Bundles[0].Targets[0]).To(Equal("package:utils/edgevpn"))
 			Expect(c.String()).ToNot(Equal(cc))
 		})
+
+		It("keeps header", func() {
+
+			var cc string = `
+config_url: "https://gist.githubusercontent.com/mudler/7e3d0426fce8bfaaeb2644f83a9bfe0c/raw/77ded58aab3ee2a8d4117db95e078f81fd08dfde/testgist.yaml"
```
mudler, line 137: we need to setup an http server here to get rid of those, but yep.. not related to this PR here

jimmykarily (approved):

PR #502, opened 2022-11-27
:seedling: Glob all configs found

mudler:
This allows to read all the configs found during scanning - this is especially useful in first boot as it allows to have separate config file logic split into several files

**What this PR does / why we need it**: Fixes up failures in https://github.com/kairos-io/provider-kairos/pull/103

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -276,14 +276,15 @@ func FindYAMLWithKey(s string, opts ...Option) ([]string, error) {
 			fmt.Printf("warning: skipping file '%s' - %s\n", f, err.Error())
 		}
 
-		data, err := unstructured.YQ(s, dat)
-		fmt.Println(data)
+		found, err := unstructured.YAMLHasKey(s, dat)
 		if err != nil {
 			fmt.Printf("warning: skipping file '%s' - %s\n", f, err.Error())
 		}
-		if len(data) > 0 {
+
+		if found {
 			result = append(result, f)
 		}
+
 	}
 
 	return result, nil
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -146,7 +146,8 @@ func Scan(opts ...Option) (c *Config, err error) {
 		b, err := os.ReadFile(f)
 		if err == nil {
 			// best effort. skip lint checks
-			yaml.Unmarshal(b, c) //nolint:errcheck
+			yaml.Unmarshal(b, c)               //nolint:errcheck
+			yaml.Unmarshal(b, &c.originalData) //nolint:errcheck
 			if exists, header := HasHeader(string(b), ""); c.IsValid() || exists {
 				c.location = f
 				yaml.Unmarshal(b, &c.originalData) //nolint:errcheck
```
jimmykarily, line 153: Is this needed now?

mudler: it is if we want to keep supporting cloud-config inputs without header (`#cloud-config`). After install now we push a `partition_info.yaml` file inside the oem folder - if the user doesn't use headers it would completely hide all the other fields gathered up to that point

jimmykarily (approved):

pkg/config/config.go
jimmykarily: We refactored in a pairing session. Comment no longer applies.

sdk/unstructured/yaml.go
```diff
@@ -9,6 +9,35 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+func YQFind(query string, content []byte) (bool, error) {
```
jimmykarily, line 12: `YQHasKey` maybe? I would expect a function called `YQFind` to fetch me the value for a key, not just a boolean of whether it exists or not.

mudler: Done!

jimmykarily (approved):

Issue #501, opened 2022-11-26
IPXE script in releases doesn't contain the release version


mudler:
**Kairos version:**
1.2.1

**Describe the bug**
This affects CI.

See https://github.com/kairos-io/kairos/releases/download/v1.2.1/kairos-alpine-opensuse-leap-v1.2.1.ipxe

```
set version
```

Should have been

```
set version v1.2.1
```

PR #516, opened 2022-11-30
:bug: Correctly detect default version in netboot

mudler:
The netboot target didn't had a default bound to the detected version, if nothing was specified it would result in an empty artifact.

Fixes: #501


jimmykarily (approved):

mudler:
mm there is still something not consistent, having a closer look

mudler:
ok locally looks good now!

PR #521, opened 2022-12-01
Docs: update customizing of kairos image

oz123:
The original document is a bit terse. Also, the kairos command no longer contains the `upgrade` subcommand, which was moved to the `karios-agent` binary.


mudler (approved):
Nais, thanks :100: !

PR #525, opened 2022-12-02
:book: Adjust documented cloud_config.yaml path for ISO remastering

alexander-bauer:
Signed-off-by: Alexander Bauer <sasha@linux.com>

**What this PR does / why we need it**: Fixes incorrect `cloud_config.yaml` path in `docs -> Installation -> Automated -> ISO Remastering`


mudler (approved):
Good catch, thank you!

PR #548, opened 2022-12-08
🐛 Set higher open file limit in Alpine

### LINKED ISSUE #537: Alpine: set a higher limit for open files
**Kairos version:** Alpine flavor

**CPU architecture, OS, and Version:** amd64

**Describe the bug**
The hard limit for open files is too low for many applications (ElasticSearch, TiDB, etc.). In the systemd-based flavor the limit is set via systemd to a much higher value.

**To Reproduce**
`ulimit -Hn` shows a value as low as 4096.

**Expected behavior**
`ulimit -Hn` should be much higher. k3s suggests 1048576 (see `/etc/systemd/system/k3s-agent.service`)

**Possible fix**
```
stages:
   initramfs.before:
     - name: "Setup higher limit for open files"
       if: '[ -e /sbin/rc-service ]'
       commands:
       - echo 'rc_ulimit="-n 1048576"' >> /etc/rc.conf
```

`ulimit` still shows the low value, since the login shell is not started as a service, but containerd has the correct high value and so does every command in a pod, which is good. (`sudo prlimit -p <PID of containerd for instance>`)

christianprim:
Signed-off-by: Christian Prim <christian.prim@gmx.ch>

Alpine only: Set a higher limit for open files, so that DB-Applications can be deployed on k3s, which is not possible with the default value of 4096.

Fixes #537


christianprim:
Sorry for the errors with spell-checking. It's the first PR I ever made in my life...

mudler:
> Sorry for the errors with spell-checking. It's the first PR I ever made in my life...

Don't worry at all about that! we are still implementing automation for spell-checks so it is giving errors that it shouldn't, thanks for the PR!

mudler (approved):
Looking good here! thank you @christianprim !

christianprim:
The patch of rc.conf was done in the wrong stage. I changed the state to boot and now there are no problems with volume mounts. I must confess, I don't know why...

I wanted to do the change as early as possible, but that was too early.

overlay/files/system/oem/30_ulimit.yaml
```diff
@@ -0,0 +1,6 @@
+stages:
+   boot:
```
mudler, line 2:
```suggestion
   boot.before:
```

mudler (approved):
Looking good for now, although I'm surprised it breaks tests when adding this into initramfs - worth investigation on our side as this look smelly cc @jimmykarily

PR #552, opened 2022-12-10
:book: Fix link pointing to missing documentation

mauromorales:
The link before was pointing to a "quickstart" page, which I couldn't find in the new docs. I thought the "getting started" page was a good substitute, but it could also be linked to the home of the documentation, or even something else. Let me know what you think.


mudler (approved):
ouch, good catch, thanks!

Issue #471, opened 2022-11-21
:penguin: Debian flavor


mudler:
**Is your feature request related to a problem? Please describe.**
So we have Ubuntu, but we should also have Debian! why not?

PR #561, opened 2022-12-15
:penguin: Add Debian flavor

mudler:
Live CD boot (interactive installer works)
Screenshot from 2022-12-15 11-58-48

After install:
Screenshot from 2022-12-15 12-13-48

Fixes: #471


framework-profile.yaml
```diff
@@ -2,14 +2,16 @@ common:
   - system/base-cloud-config
   - dracut/immutable-rootfs
   - dracut/kcrypt
-  - dracut/network
```
mudler, line 5: this was an oversight, we can't use it as is by default, as systemd-networkd support is not working enough yet with netboot

paynejacob (approved):

PR #562, opened 2022-12-15
:robot: Add rockylinux/fedora to the test matrix

mudler:

Change the `--cpu` flag when running tests. This is a test to see if CI would be impacted by the change


mudler:
eeeek nope

PR #569, opened 2022-12-19
:seedling: Add /usr/local/system/providers to bus

mudler:
This allows bundles to install providers in the persistent path

### DIFF
--- a/internal/bus/bus.go
+++ b/internal/bus/bus.go
@@ -32,7 +32,7 @@ type Bus struct {
 
 func (b *Bus) LoadProviders() {
 	wd, _ := os.Getwd()
-	b.Manager.Autoload("agent-provider", "/system/providers", wd).Register()
+	b.Manager.Autoload("agent-provider", "/system/providers", "/usr/local/system/providers", wd).Register()
 }
 
 func (b *Bus) Initialize() {


oz123:
@mudler is there an issue explaining what this solves? I am not really able to review this, so I'll refrain until another team member reviews.

jimmykarily (approved):

internal/bus/bus.go
```diff
@@ -32,7 +32,7 @@ type Bus struct {
 
 func (b *Bus) LoadProviders() {
 	wd, _ := os.Getwd()
-	b.Manager.Autoload("agent-provider", "/system/providers", wd).Register()
+	b.Manager.Autoload("agent-provider", "/system/providers", "/usr/local/system/providers", wd).Register()
```
mauromorales, line 35: Having the same structure in system makes sense to me, can I just ask, where the whole `system` comes from?

mauromorales (approved):
lgtm

internal/bus/bus.go
mudler: `system` it is just a convention, it contains additional config and binaries reserved for kairos. E.g. you will find the default cloud configs of kairos inside `/system/oem`

mauromorales: understood, thanks :)

PR #575, opened 2022-12-20
🐛 Remove openntpd

### LINKED ISSUE #519: Alpine: use busybox ntpd
**Kairos version:** alpine-ubuntu and probably alpine-opensuse-leap

**CPU architecture, OS, and Version:** amd64

**Describe the bug**
It could be just a problem for my server, but when openntpd is used, prometheus gives an alert:

```
 Clock on <hostname>:9796 is out of sync by more than 300s. Ensure NTP is configured correctly on this host.
```
I stopped the openntpd-service, patched `/etc/init.d/ntpd` to use the busybox binary (necessary since ntpd doesn't point to busybox), started the ntpd-service and the alert disappears.

(There would be no need for a patch if openntpd wasn't installed in the alpine flavor, ntpd would point to busybox)

**To Reproduce**
Join an alpine k3s-agent (no special config) to a cluster with prometheus monitoring and alerting installed. After a short period the alert is firing.

**Expected behavior**
No such errors.

**Question**
Is it mandatory to install a dedicated ntpd-package on alpine? Why not stick with the (working) busybox-version?

It could also help reduce confusion. See #476

christianprim:
Signed-off-by: Christian Prim <christian.prim@gmx.ch>

Removes openntpd from Alpine images as it makes some trouble

Fixes #519

### DIFF
--- a/images/Dockerfile.alpine-opensuse-leap
+++ b/images/Dockerfile.alpine-opensuse-leap
@@ -102,7 +102,7 @@ RUN rc-update add sshd boot && \
     rc-update add syslog boot && \
     rc-update add udev sysinit && \
     rc-update add udev-trigger sysinit && \
-    rc-update add openntpd boot && \
+    rc-update add ntpd boot && \
     rc-update add crond && \
     rc-update add fail2ban
 
--- a/images/Dockerfile.alpine-opensuse-leap
+++ b/images/Dockerfile.alpine-opensuse-leap
@@ -93,8 +93,7 @@ RUN apk --no-cache add  \
       open-vm-tools-deploypkg \
       open-vm-tools-guestinfo \
       open-vm-tools-static \
-      open-vm-tools-vmbackup \
-      openntpd
+      open-vm-tools-vmbackup
 
 RUN rc-update add sshd boot && \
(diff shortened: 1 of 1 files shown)


christianprim:
@mudler : where can I find the iso-image from this run so I can test if busybox ntpd really gets started?

mudler:
hey @christianprim thx for taking a stab at it! The ISOs are available in the action summary page after the build ends: https://github.com/kairos-io/kairos/actions/runs/3741359725

mudler (approved):
Looks good here, @christianprim if you are giving it a try I'll wait for your feedback - we can also add a e2e test in the acceptance suite I think  cc @jimmykarily

christianprim:
My laptop is an older one, so the install lasted longer under qemu... But I can report, that ntpd has started!

jimmykarily:
> My laptop is an older one, so the install lasted longer under qemu... But I can report, that ntpd has started!

I've found out that `--enable-kvm` makes a big difference in performance (in case you didn't know that).

jimmykarily (approved):

jimmykarily:
It's green, merged. Thanks @christianprim !

PR #583, opened 2022-12-23
:bug: Fix rockylinux build, curl-minimal is already shipped

mudler:

**What this PR does / why we need it**: Currently CI is red on rockylinux. Seems now curl-minimal is shipped in the base image, previously we were installing `curl` and now are conflicting


mudler:
Ok rockylinux builds now, merging as we have #181 already flying, and this currently blocks CI tests.

PR #615, opened 2023-01-06
:robot: Disable spellchecks

mudler:
It is really invasive. We did experimented with it  bit but I don't think it's worth it. It clashes periodically with robots PRs, making review almost impossible if not practical.

I think it's best to disable it until we get it implemented properly


mauromorales (approved):
lgtm

PR #617, opened 2023-01-06
:arrow_up: update github.com/mudler deps

### LINKED ISSUE #614: :arrow_up: Update go mod dependencies
Deps from my namespace `mudler` had license changes, so we want to pin to the new versions

Update the deps in go.mod

mauromorales:

License updates on mudler deps

Fixes #614


go.mod
```diff
@@ -13,8 +13,8 @@ require (
 	github.com/jaypipes/ghw v0.9.0
 	github.com/joho/godotenv v1.4.0
 	github.com/kairos-io/kcrypt v0.4.4
-	github.com/labstack/echo/v4 v4.6.3
-	github.com/mudler/go-nodepair v0.0.0-20220507212557-7d47aa3cc1f1
+	github.com/labstack/echo/v4 v4.9.0
+	github.com/mudler/go-nodepair v0.0.0-20221223092639-ba399a66fdfb
```
mudler, line 17: I wonder where go-nodepair is being tied here. It should be needed only by the provider, but maybe I'm overlooking something now

mauromorales: is it possible, and does it make sense to have different mod files for each project? for example I see docsy comes with it's own mod file. I wonder if we could do the same for the agent for example

mudler (approved):

Issue #48, opened 2022-08-01
Documentation: bundles


mudler:
Bundles has been already introduced and available to the master branch, however docs are not updated to reflect the new featureset.

## Action items
- [ ] Update docs to include references to c3os bundles
- [ ] Add examples on how to use bundles

PR #618, opened 2023-01-06
:book: Add examples and bundles section

mudler:

Fixes: https://github.com/kairos-io/kairos/issues/48

Adds examples:

- [x] Single node  
- [x] Multi-node cluster  
- [x] Multi node cluster with HA (moved from Advanced section)
- [x] single-node cluster with P2P  
- [x] multi-node cluster with P2P  
- [x] multi-node cluster with P2P and KubeVIP  
- [x] use the Kairos core image to deploy k3s
- [x] use bundles
- [x] Update p2p config reference 
- [x] Example of how to use a bundle in the `example` section (link to `advanced/build`, copy-paste example)


mudler:
I'll leave the webui docs once we have a release for it

mudler:
Just some rephrasing left and we are good to go

mudler:
Opened #627 for the webui docs

README.md
```diff
@@ -42,6 +42,7 @@ For comprehensive docs, tutorials, and examples see our [documentation](https://
 
 ## Project status
 
+- (Dec 28, 2022) announcing the release of Kairos 1.4, which includes self-coordinated automated HA deployment via p2p and with [KubeVIP](https://kube-vip.io/). Check out the [docs](https://kairos.io/docs/installation/p2p/) for more information on this exciting new feature! Support for Ubuntu, Debian, and Fedora is also added in the Kairos 1.4 release. Expanding the number of supported operating systems currently available to pick from!
```
mauromorales, line 45: Shall we create blog posts for major release announcements in the future?

docs/content/en/docs/Advanced/bundles.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+## Bundle types
+
+Bundles can carry also binaries that can be overlayed in the rootfs, either while [building images](/docs/advanced/build) or with [Live layering](https://kairos.io/docs/advanced/livelayering/).
+
+
+Kairos supports three types of bundles:
+
+- *Container*: This type is a bare container that simply contains files that need to be copied to the system. It is useful for copying over configuration files, scripts, or any other static content that you want to include on your system.
```
mauromorales, line 77: How about bold instead of italics?

docs/content/en/docs/Advanced/bundles.md
```diff
@@ -0,0 +1,149 @@
...
+
+- *Run*: This type is also a bare container, but it comes with a script that can be run during the installation phase to add custom logic. This is useful for performing any necessary setup or configuration tasks that need to be done before the cluster is fully deployed.
+
+- *Package*: This type is a [luet](https://luet.io) package that will be installed in the system. It requires the luet repository definition in order to work. Luet packages are a powerful way to manage dependencies and install software on your system.
+
+Note: In the future, Kairos will also support a local type for use in airgap situations, where you can pre-add bundles to the image before deployment.
```
mauromorales, line 83: Maybe put this in an alert?

docs/content/en/docs/Advanced/customizing.md
```diff
@@ -45,7 +45,7 @@ c58930881bc4: Pushed
 ...
 ```
 
-You can use your custom image with the `kairos-agent upgrade` command, or with the [system-upgrade-controller in Kubernetes](/docs/upgrade/kubernetes). Here's how to use the `kairos-agent` command:
+You can use your custom image when [upgrade nodes manually](/docs/upgrade/manual), [with Kubernetes](/docs/upgrade/kubernetes) or [specifying it in the cloud-config during installation](/docs/examples/core). Here's how to do it manually with the `kairos-agent` command:
```
mauromorales, line 48: maybe `when upgrading nodes manually` ?

docs/content/en/docs/Advanced/bundles.md
```diff
@@ -0,0 +1,149 @@
...
+This will push the "my-bundle" image to your specified registry.
+
+And use it with Kairos:
+
+```yaml
+#cloud-config
+
+bundles:
+- targets:
+  - run://<image>
```
mauromorales, line 143: the `<image>`  used here is the same as the tag in 123 but if I understand correctly here i need the full path e.g. `quay.io/kairos/community-bundles:metallb_latest`

docs/content/en/docs/Examples/multi-node-p2p-ha-kubevip.md
```diff
@@ -0,0 +1,75 @@
...
+     # A master node is always required for creating the cluster and is implied.
+     # The setting below adds 2 additional master nodes, for a total of 3.
+     master_nodes: 2
+```
+
+When configuring the `p2p` section, start by adding your desired `network_token` under the p2p configuration in the cloud-config file. To generate a network token, see [documentation](/docs/installation/p2p/#network_token).
+
+Next, set up an Elastic IP (`kubevip.eip`) with a free IP in your network. KubeVIP will advertise this IP, so make sure to select an IP that is available for use on your network.
+
+In the vpn configuration, the create and use options are disabled, so the VPN setup is skipped and not used to route any traffic into.
```
mauromorales, line 75:
```suggestion
In the VPN configuration, the create and use options are disabled, so the VPN setup is skipped and not used to route any traffic into.
```

docs/content/en/docs/Examples/single-node-p2p.md
```diff
@@ -0,0 +1,54 @@
+---
+title: "P2P single-node cluster"
+linkTitle: "P2P single-node cluster"
+weight: 6
+date: 2022-11-13
+description: >
+   This documentation page provides instructions on how to install Kairos with P2P support on a single-node cluster
+---
+
+{{% alert title="Note" %}}
```
mauromorales, line 10: Maybe `Warning` would be more appropriate for these alerts?

docs/content/en/docs/Examples/single-node.md
```diff
@@ -0,0 +1,78 @@
...
+          namespace: metallb-system
+        spec:
+          ipAddressPools:
+          - default
+```
+
+Notably:
+
+- we use the `k3s` block to disable `traefik` and `servicelb` (the default `k3s` load balancer)
+- we use `write_files` to write manifests to the default `k3s` manifest directory (`/var/lib/rancher/k3s/server/manifests/`) see [docs](/docs/reference/configuration#kubernetes-manifests) to deploy `MetalLB` and configure it with the `192.168.1.10-192.168.1.20` IP range. Make sure to pick up a range which doesn't interfere with your local DHCP network.
```
mauromorales, line 76: Maybe start these with capitals, and end both with dots (or both without)

docs/content/en/docs/Advanced/bundles.md
```diff
@@ -0,0 +1,149 @@
...
+
+```yaml
+#cloud-config
+install:
+ bundles:
+ - targets:
+   - run://<image>
+```
+
+One of the benefits of using bundles is that they can also extend the cloud-config keywords available during deployment. This means that by adding bundles to your configuration file, you can add new blocks of configuration options and customize your system even further.
```
jimmykarily, line 40: We called it "installation" right above, let's be consistent:

```suggestion
One of the benefits of using bundles is that they can also extend the cloud-config keywords available during installation. This means that by adding bundles to your configuration file, you can add new blocks of configuration options and customize your system even further.
```

jimmykarily: if a bundle is installed after installation (as described above), then and `#cloud-config` it provides, will be ignored right? Maybe we should clarify that here.

docs/content/en/docs/Advanced/bundles.md
```diff
@@ -0,0 +1,149 @@
...
+
+# Specify metallb settings
+metallb:
+  version: 0.13.7
+  address_pool: 192.168.1.10-192.168.1.20
+```
+
+## Bundle types
+
+Bundles can carry also binaries that can be overlayed in the rootfs, either while [building images](/docs/advanced/build) or with [Live layering](https://kairos.io/docs/advanced/livelayering/).
```
jimmykarily, line 72: Should we warn the users about implications? E.g. that the the bundle should be recompiled for each version of the kairos image (is this true?)

docs/content/en/docs/Advanced/bundles.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+Kairos supports three types of bundles:
+
+- *Container*: This type is a bare container that simply contains files that need to be copied to the system. It is useful for copying over configuration files, scripts, or any other static content that you want to include on your system.
+
+- *Run*: This type is also a bare container, but it comes with a script that can be run during the installation phase to add custom logic. This is useful for performing any necessary setup or configuration tasks that need to be done before the cluster is fully deployed.
+
+- *Package*: This type is a [luet](https://luet.io) package that will be installed in the system. It requires the luet repository definition in order to work. Luet packages are a powerful way to manage dependencies and install software on your system.
```
jimmykarily, line 81: "It requires the luet repository definition in order to work" -> what does this mean?

docs/content/en/docs/Advanced/bundles.md
```diff
@@ -0,0 +1,149 @@
...
+
+To create a bundle, you will need to define a base image and copy over any necessary files and scripts to the image. For example, you might use the following Dockerfile to create a bundle image that deploys everything inside `assets` in the Kubernetes cluster:
+
+```Dockerfile
+FROM alpine
+COPY ./run.sh /
+COPY ./assets /assets
+```
+
+And the associated `run.sh` that installs the assets depending on a cloud-config keyword can be:
```
jimmykarily, line 100: Given this bundle only copies files, couldn't it be a "container" type instead of a "run"?

docs/content/en/docs/Advanced/bundles.md
```diff
@@ -0,0 +1,149 @@
...
+
+- *Run*: This type is also a bare container, but it comes with a script that can be run during the installation phase to add custom logic. This is useful for performing any necessary setup or configuration tasks that need to be done before the cluster is fully deployed.
+
+- *Package*: This type is a [luet](https://luet.io) package that will be installed in the system. It requires the luet repository definition in order to work. Luet packages are a powerful way to manage dependencies and install software on your system.
+
+Note: In the future, Kairos will also support a local type for use in airgap situations, where you can pre-add bundles to the image before deployment.
+
+It's important to note that bundles do not have any special meaning in terms of immutability. They install files over paths that are mutable in the system, as they are simply overlaid during the boot process. This means that you can use bundles to make changes to your system at any time, even after it has been deployed.
```
jimmykarily, line 85: So what happens if a bundle tries to write a file to an "immutable" (ro) path? If I had a bundle that provides a binary (e.g. `k9s`), would I be able to put it in `/usr/bin/`? 

I'm not sure the first sentence adds any information: `It's important to note that bundles do not have any special meaning in terms of immutability.`.  In a way, there is one thing we are breaking with bundles and that's the atomic installation of just one container image. Now the final installation may comprise of multiple images.

Maybe we should explain when bundles should be used over creating the final image directly. Some reasons I can think of:

- Use third-party bundles (consume other people's work)
- Re-use the same bundles in multiple (otherwise different) installations

Is there a way, to use bundles in an osbuilder Job so that a final image is generated that would be the equivalent of using the same bundles in the config?

docs/content/en/docs/Examples/bundles.md
```diff
@@ -0,0 +1,59 @@
...
+#cloud-config
+
+hostname: metal-{{ trunc 4 .MachineID }}
+users:
+- name: kairos
+  # Change to your pass here
+  passwd: kairos
+  ssh_authorized_keys:
+  # Add your github user here!
+  - github:mudler
```
jimmykarily, line 36:
```suggestion
  # Replace with your github user and un-comment the line below:
  # - github:mudler
```

docs/content/en/docs/Examples/core.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+## Configuring the installation
+
+As you move through the installation process, there are a few key points to keep in mind when configuring your cloud config file:
+
+- We set `install.image` to the container image that we want to deploy. This can be an image from [our support matrix](/docs/reference/image_matrix), a [custom image](/docs/advanced/customizing) or an [image from scratch](/docs/reference/build).
+- After the installation is complete, the configuration in the `k3s` block will take effect. This is because after the installation, the system will boot into the image specified in the `install.image` field, which in the example above is an image with the Kairos K3s provider, as such the configuration in the k3s block will become active.
+
+With these steps, you should now be able to use Kairos core images as an installer to deploy other container images. The process is straightforward and gives you the flexibility to customize your deployments and build custom images as needed.
```
jimmykarily, line 58: Does this also mean that we can now make the provider images smaller, providing only the additional files, excluding the OS filesystem already provided by the core image?

docs/content/en/docs/Examples/metallb.md
```diff
@@ -1,20 +1,21 @@
...
 
-## Installation
+But first, let's talk a little bit about what [MetalLB](https://metallb.universe.tf/) and [K3s](https://k3s.io/) are. MetalLB is a load balancer implementation for bare metal Kubernetes clusters that uses standard routing protocols. It's particularly useful when used with K3s in Kairos, as it provides load balancing for bare metal clusters and helps manage IP addresses within the cluster. K3s is a lightweight Kubernetes distribution that is easy to install and maintain.
 
-Use the [provider-kairos](https://github.com/kairos-io/provider-kairos) artifacts which contains `k3s`.
+Now that you have an understanding of what we'll be working with, let's dive into the installation process.
 
-We will use the [k3s manifest method](/docs/reference/configuration#kubernetes-manifests) to deploy `MetaLB`.
+
+To get started, you'll need to use the [provider-kairos](https://github.com/kairos-io/provider-kairos) artifacts, which include k3s. We'll be using the [k3s manifest method](/docs/reference/configuration#kubernetes-manifests) to deploy MetalLB.
```
jimmykarily, line 18: backlink to the "bundle" method as an alternative?

docs/content/en/docs/Examples/multi-node-p2p-ha-kubevip.md
```diff
@@ -0,0 +1,75 @@
...
+
+```yaml
+#cloud-config
+
+hostname: kairoslab-{{ trunc 4 .MachineID }}
+users:
+- name: kairos
+  ssh_authorized_keys:
+  # Add your github user here!
+  - github:mudler
```
jimmykarily, line 35:
```suggestion
  # Change to your pass here
  passwd: kairos
  # Replace with your github user and un-comment the line below:
  # - github:mudler
```

docs/content/en/docs/Examples/multi-node.md
```diff
@@ -0,0 +1,64 @@
...
+#cloud-config
+
+hostname: metal-{{ trunc 4 .MachineID }}
+users:
+- name: kairos
+  # Change to your pass here
+  passwd: kairos
+  ssh_authorized_keys:
+  # Add your github user here!
+  - github:mudler
```
jimmykarily, line 29: same here

jimmykarily, line 29:
```suggestion
  # Replace with your github user and un-comment the line below:
  # - github:mudler
```

docs/content/en/docs/Examples/multi-node-p2p.md
```diff
@@ -0,0 +1,47 @@
...
+
+```yaml
+#cloud-config
+
+hostname: kairoslab-{{ trunc 4 .MachineID }}
+users:
+- name: kairos
+  ssh_authorized_keys:
+  # Add your github user here!
+  - github:mudler
```
jimmykarily, line 30:
```suggestion
  # Replace with your github user and un-comment the line below:
  # - github:mudler
```

docs/content/en/docs/Examples/single-node.md
```diff
@@ -0,0 +1,78 @@
...
+#cloud-config
+
+hostname: metal-{{ trunc 4 .MachineID }}
+users:
+- name: kairos
+  # Change to your pass here
+  passwd: kairos
+  ssh_authorized_keys:
+  # Add your github user here!
+  - github:mudler
```
jimmykarily, line 27:
```suggestion
  # Replace with your github user and un-comment the line below:
  # - github:mudler
```

docs/content/en/docs/Examples/single-node.md
```diff
@@ -0,0 +1,78 @@
...
+
+k3s:
+  enabled: true
+  args:
+  - --disable=traefik,servicelb
+
+# Additional manifests that are applied by k3s on boot
```
jimmykarily, line 34: This part has nothing to do with the single node cluster setup right? I think we already demonstrated this metallb setup here: https://github.com/kairos-io/kairos/pull/618/files#diff-e23d6395ab8483864c81358c557904814cd5a78f16193e2b086363638a188fde

docs/content/en/docs/Reference/configuration.md
```diff
@@ -13,18 +13,14 @@ The structure of the configuration file is as follows:
...
-  # Enable embedded DNS See also: https://mudler.github.io/edgevpn/docs/concepts/overview/dns/
-  dns: true
+# Additional system users
+users:
+- name: "kairos"
+  passwd: "kairos"
+  lock_passwd: true
+  groups: "admin"
+  ssh_authorized_keys:
+  - github:mudler
```
jimmykarily, line 23:
```suggestion
  # - github:mudler
```

jimmykarily (approved):

README.md
mudler: Yup agreed and makes sense. At least for impactful releases with changes that are worth mentioning, but let's make it opt-in or the process can become quickly cumbersome.

docs/content/en/docs/Advanced/bundles.md
mudler: right, but we do read the config file, based on that we copy the assets or we don't

docs/content/en/docs/Advanced/bundles.md
mudler: If a bundle writes to a **ro** path everything that would do would be discarded. It's responsibility of who writes the bundle to document its usage and implications.

Bundle are very generic ways to deliver specific tasks to agents,  that can span to dynamic configuration, adding statically files, but all of that needs to happen on paths that are mutable in the system.

osbuilder is privileged - as it can assembles new images can import directly what we call bundles in the root of the base image. but at that point bundles can just contain binaries - and I think should be treated separately or somehow push a layer with the dynamic configuration (our `run.sh`) as a hidden layer in the container image, but we do not have such a mechanism in place yet.

docs/content/en/docs/Examples/core.md
mudler: we could - however we should start to think about implications during upgrades if we go down that path

mudler: and for that I'd rather have a interface around CAPI to handle all of this transparently, but we aren't there yet

docs/content/en/docs/Advanced/bundles.md
mudler: that implication is valid only for Live Layering (indeed it's documented only there). bundles are not necessarly tied to a kairos version

docs/content/en/docs/Advanced/bundles.md
mudler: I didn't wanted to make it sound complex here as it's mostly an implication of how bundles deliver business logic. for instance, the bundles currently in community-bundle runs also before deployment of k3s (that is at the first boot), but let's iterate over this. I don't want to make this PR already too much big

PR #619, opened 2023-01-06
:seedling: Vendor webui dependencies

### LINKED ISSUE #608: :seedling: Vendor webui dependencies
As in https://github.com/kairos-io/kairos/pull/587 we introduced the webui, it currently requires internet access to render correctly. This card is about to vendor all the required assets such as css and js files.

Options:
- vendor all the code in git, with something like `npm`, or a golang equivalent to update them
- generate the required files while building in the CI, to avoid shipping them as part of the code base

See also:
_Originally posted by @mudler in https://github.com/kairos-io/kairos/pull/587#discussion_r1062365941_

mauromorales:

- [x] Add web UI dependencies using npm
- [x] generate node_modules during build time
- [ ] cache node_modules installation

Fixes #608


.github/workflows/image.yaml
```diff
@@ -43,6 +59,14 @@ jobs:
       - name: Login to Quay Registry
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
         run: echo ${{ secrets.QUAY_PASSWORD }} | docker login -u ${{ secrets.QUAY_USERNAME }} --password-stdin quay.io
+      - if: ${{ steps.cache-npm.outputs.cache-hit != 'true' }}
+        name: List the state of node modules
+        continue-on-error: true
+        working-directory: ./internal/webui/public
+        run: npm list
+      - name: Install WebUI dependencies
+        working-directory: ./internal/webui/public
```
mudler, line 68: I think it's best to fit those in the Earthfile target rather in the github action, so when someone calls "build" it builds the binary, with deps included

Earthfile
```diff
@@ -22,6 +22,7 @@ ARG GOLINT_VERSION=1.47.3
 ARG GO_VERSION=1.18
 
 all:
+  BUILD +webui-deps
```
mudler, line 25: better to connect it to the target that builds the agent instead of untied to +all. Something like this:

```
build-kairos-agent: 
        COPY +webui-deps/node_modules ./internal/webui/public
        DO ....
```

mauromorales:
@mudler after some fiddling around between GA and Earthly I decided to leave it the same way as go deps are currently installed. However, I'd like this to be a bit different in the future because I find it's a bit wasteful to re-build dependencies if the definition of the dependencies didn't change. I think that the CI should instead build the dependencies before calling the builds matrix, so that each of them don't need to do the building themselves. It's probably not going to be a huge time improvement but I like the idea of not having to do any unnecessary computations/downloads/etc. If we were doing it all on GA it would look something like this:

image

But I agree with you that the most we can stay independent from GA, the better. What do you think?

In the meantime, I think this can be reviewed

Earthfile
```diff
@@ -230,6 +230,9 @@ docker:
 
     # Copy kairos binaries
     COPY +build-kairos-agent/kairos-agent /usr/bin/kairos-agent
+
```
mudler, line 233: mmm I wonder how this can work. the webui-deps end up to be in the docker image, not in the source code that is going to be compiled. The target that builds the agent is `build-kairos-agent`.

When running `earthly +build-kairos-agent` I expect an agent binary with the webui assets included inside (also because otherwise the binary would not container the assets, and the webui would be broken)

mauromorales: good catch, it was only working because I didn't clean my local node_modules directory :man_facepalming: ... I'll look into it

mauromorales: ok, it should be working properly now

Earthfile
```diff
@@ -106,6 +106,8 @@ version:
 
 build-kairos-agent:
     FROM +go-deps
+    BUILD +webui-deps
```
mudler, line 109: you shouldn't need to specify BUILD. `COPY` should automatically make it build the deps

mauromorales: @mudler indeed it is not needed for the building of the agent, but if I don't keep it then, it is missing from my local

mudler: oh ok I see! otoh I'd be rather surprised by the build command to populate something in my source code, maybe it's just ok to let that to be called separately; I don't see us touching those and maybe just gets in the way

mauromorales: @mudler oki removed it

mudler (approved):

mudler:
tested locally and works!

PR #620, opened 2023-01-07
:seedling: Webui enhancements

mudler:

Currently the WebUI doesn't start on Alpine and this PR fixes it. This also avoids the tty to loop or stall if no provider is found, now it will drop to a shell by default and print the WebUI address and the available interfaces.


internal/agent/install.go
```diff
@@ -110,14 +133,24 @@ func Install(dir ...string) error {
...
 	}
 
-	_, err = bus.Manager.Publish(events.EventChallenge, events.EventPayload{Config: cc.String()})
+	agentConfig, err := LoadConfig()
 	if err != nil {
 		return err
 	}
 
+	// try to clear screen
+	fmt.Print("\033c")
```
mauromorales, line 142: would be nice to have these kind of comment + expression as functions part of a Kairos library, for readability and reuse

internal/agent/install.go
```diff
@@ -110,14 +133,24 @@ func Install(dir ...string) error {
...
+
 	cmd.PrintBranding(DefaultBanner)
 
-	agentConfig, err := LoadConfig()
+	// If there are no providers registered, we enter a shell for manual installation and print information about
+	// the webUI
+	if len(bus.Manager.Plugins) == 0 {
```
mauromorales, line 148: I'd recommend to start using predicate functions for readability and flexibility for future changes. For example something like `bus.Manager.HasRegisteredProviders()`

pkg/config/config.go
```diff
@@ -19,6 +19,8 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+const DefaultWebUIAddr = ":8080"
```
mauromorales, line 22:
```suggestion
const DefaultWebUIPort = ":8080"
```

internal/agent/install.go
```diff
@@ -38,6 +39,28 @@ func optsToArgs(options map[string]string) (res []string) {
 	return
 }
 
+func displayInfo(agentConfig *Config) {
+	fmt.Println("--------------------------")
+	fmt.Println("No providers found, dropping to a shell. \n -- For instructions on how to install manually, see: https://kairos.io/docs/installation/manual/")
+	if !agentConfig.WebUI.Disable {
+		if agentConfig.WebUI.ListenAddress == "" {
```
mauromorales, line 46: also could be a predicate funcion

internal/agent/install.go
mudler: good catch! addressed in https://github.com/kairos-io/kairos/pull/620/commits/8ae581024504ea5a84c730c547cbaa2c22e01d17 !

pkg/config/config.go
mudler: err, changed to WebUIListenAddr. it includes also the bind addr, e.g. "127.0.0.1:8080" is valid input

mauromorales (approved):
lgtm

Issue #643, opened 2023-01-12
Add annotated git version to `$CLI --version.`


oz123:
Currently, `karios-agent` and `kairos-cli` have hard coded version, e.g.

```
func Start() error {
	app := &cli.App{
		Name:    "kairos",
		Version: "0.1",
		Authors: []*cli.Author{
			{
				Name: "Ettore Di Giacinto",
			},
		},func main() {
	bus.Manager.Initialize()
...
```
And 
```
	app := &cli.App{
		Name:    "kairos-agent",
		Version: "0.1",
		Author:  "Ettore Di Giacinto",
		Usage:   "kairos agent start",
		Description: `
The kairos agent is a component to abstract away node ops, providing a common feature-set across kairos variants.
`,
`
		Usage: "kairos CLI to bootstrap, upgrade, connect and manage a kairos network",
		Description: `
...
```
(There might be other cli tools I am not aware of.)

This is not updated with each release, hence even current releases show version 0.1, e.g.
```
 ./build/kairos-agent --version
kairos-agent version 0.1
```

**Describe the solution you'd like**

I would like to have a version information which automatically update with each build, the version should include the release tag as well as the commit information and the VCS status. Here is an example of the theoretical output:
```
$ ./kairos-agent --version
Version: 1.25
FullVersion Info: {Major:"1", Minor:"25", GitVersion:"v1.25.4", GitCommit:"872a965c6c6526caa949f0c6ac028ef7aff3fb78", GitTreeState:"archive", BuildDate:"2022-12-18T13:01:35Z", GoVersion:"go1.19.4", Compiler:"gc", Platform:"linux/amd64"}
```

This is similar to `kubectl` (Hence, we can probably reuse code from there...).

**Describe alternatives you've considered**

There are a few ways to achieve this, some ways have even more information. 

The first one is adding LDFLAGS:

BUILD := $(git describe)
LDFLAGS=-ldflags "-X=$(GIT)cmd.VERSION=$(BUILD)"
go build -v $(LDFLAGS) -o ....

The second option (could be complicated) reusing the same mechanism as in kubectl.

The final and most verbose option is by exposing `debug/buildinfo` within our CLI. However, this is IMHO, redundant as one can do the following, to show it:
```
$ go version -m provider-kairos
provider-kairos: go1.18
	path	github.com/kairos-io/provider-kairos
	mod	github.com/kairos-io/provider-kairos	(devel)	
...
	build	-compiler=gc
	build	CGO_ENABLED=1
[...]

PR #651, opened 2023-01-13
:seedling: Add version information directly from git

oz123:
Partial fix for kairos-io/kairos#643.

Signed-off-by: Oz Tiram <oz@spectrocloud.com>

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -79,9 +79,13 @@ BUILD_GOLANG:
     ARG CGO_ENABLED
     ARG BIN
     ARG SRC
+    COPY +version/VERSION ./
+    ARG VERSION=$(cat VERSION)
     ENV CGO_ENABLED=${CGO_ENABLED}
-
-    RUN go build -ldflags "-s -w" -o ${BIN} ./cmd/${SRC} && upx ${BIN}
+    ARG LDFLAGS="-s -w -X 'github.com/kairos-io/kairos/internal/common.VERSION=$VERSION'"
+    RUN echo "Building ${BIN} from ${SRC} using ${VERSION}"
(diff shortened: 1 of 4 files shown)


Earthfile
```diff
@@ -79,9 +79,13 @@ BUILD_GOLANG:
     ARG CGO_ENABLED
     ARG BIN
     ARG SRC
+    COPY +version/VERSION ./
+    ARG VERSION=$(cat VERSION)
     ENV CGO_ENABLED=${CGO_ENABLED}
-
-    RUN go build -ldflags "-s -w" -o ${BIN} ./cmd/${SRC} && upx ${BIN}
+    ARG LDFLAGS="-s -w -X 'github.com/kairos-io/kairos/pkg/common.VERSION=$VERSION'"
```
mudler, line 85: There is no need to expose this in `pkg`, so better to consume a var inside `internal`. Also use camelcase here, I think the linter will complain

oz123: Ah, thanks for reminding me about `internal`!  I should use that, indeed!

oz123: Done.

mudler (dismissed):

cmd/profile-build/main.go
```diff
@@ -12,7 +13,7 @@ func main() {
 
 	app := &cli.App{
 		Name:    "profile-build",
-		Version: "0.1",
+		Version: common.VERSION,
 		Author:  "Ettore Di Giacinto",
```
mudler, line 17:
```suggestion
		Author:  "Kairos authors",
```

mudler (approved):

PR #667, opened 2023-01-18
[WIP] :seedling: Rework ISO_NAME on earthfile

Itxaka:
No need to override the ISO_NAME with OS_ID in several places, it already has the proper name.

Also use the ISO_NAME when needed instead of relaying in it being copied to kairos.iso, so we do less steps and we always refer to the ISO_NAME var

Fixes #


Itxaka:
Locally seems to pass tests  but Im not sure if Im missing something...

oz123:
Is it possible to include VERSION in the iso name? I think it's useful.

PR #675, opened 2023-01-18
Adapt CI to match version generation between build and qemu-bundles-tests steps

mauromorales:

The test with kubo was failing because there's a version check done by `systemd-sysext` and it didn't match. The issue comes from the new way of fetching the version from git, which in the build step we use the full history and tags so it will look something like `v1.2.3-sha-dirty` while in the `qemu-bundles-tests` we fetch only the last commit without tags or history so it looks just like `sha`. 

Let's make master green again!

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -1,4 +1,4 @@
-name: Build image
+name: build image
 
 on: 
  push:
@@ -14,7 +14,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     permissions:
-      id-token: write  # OIDC support
+      id-token: write  # oidc support
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/image.yaml
```diff
@@ -189,6 +189,8 @@ jobs:
          - flavor: "opensuse"
     steps:
     - uses: actions/checkout@v3
+      - run: |
```
mudler, line 192: what about the release pipeline?

mauromorales: @mudler for what I can tell both release.yaml and release-arm.yaml are ok, but I'm not entirely sure about release_bin.yaml does goreleaser somehow does any building too? if not then I think we're good

.github/workflows/image.yaml
```diff
@@ -34,21 +34,21 @@ jobs:
...
+      - name: login to quay registry
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
-        run: echo ${{ secrets.QUAY_PASSWORD }} | docker login -u ${{ secrets.QUAY_USERNAME }} --password-stdin quay.io
-      - name: Build  🔧
+        run: echo ${{ secrets.quay_password }} | docker login -u ${{ secrets.quay_username }} --password-stdin quay.io
+      - name: build  🔧
         env:
-          FLAVOR: ${{ matrix.flavor }}
-          IMAGE: quay.io/kairos/core-${{ matrix.flavor }}:latest
+          flavor: ${{ matrix.flavor }}
```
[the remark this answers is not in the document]
mauromorales: :facepalm: ah crap ... I blame it on Vim :P ... fixing

mudler:
awesome, good catch!

mudler (approved):

PR #681, opened 2023-01-19
:book: Rename references from kairos to p2p

### LINKED ISSUE #680: :book: Configuration page - kairos mentioned instead of p2
```
The kairos block is used to enable the p2p full-mesh functionalities of Kairos. If you do not want to use these functionalities, simply don’t specify a kairos block in your configuration file.

Inside the kairos block, you can specify the network_token field, which is used to establish the p2p full meshed network. If you do not want to use the full-mesh functionalities, don’t specify a network_token value.
```

Itxaka:
Leftover from the transition, kairos block is called p2p now

Less confusion on docs

Fixes #680


mudler (approved):
Looking good!

PR #697, opened 2023-01-20
:seedling: Drop references to node-config

mudler:

We have still traces of  `#node-config` that causes confusion among our users, this tries to be consistent by dropping the last reference (I cound find)


Itxaka (approved):
thanks you!

Issue #265, opened 2022-10-24
Intermittent download failures with openSUSE base images in CI


mudler:
**Kairos version:**
This doesn't affect Kairos, but rather Kairos CI

**Describe the bug**
This is a tracker for all the occurences of download failures with `zypper`.  Currently those are cumulating and hitting us regularly, for example:
```
#86 13.11  - File './repodata/6ea614a64061dd1840adfa7deee42986daaab8916caaa16408c540f17a8ac790-deltainfo.xml.gz' not found on medium 'http://download.opensuse.org/update/leap/15.4/backports/'
#86 13.11  - Can't provide ./repodata/6ea614a64061dd1840adfa7deee42986daaab8916caaa16408c540f17a8ac790-deltainfo.xml.gz
#86 13.11 
```

**To Reproduce**

**Expected behavior**
I'd rather prefer to switch to a different base or build an image with all the packages needed for tests inside

**Logs**

**Additional context**
- https://github.com/kairos-io/kairos/pull/264
- https://github.com/kairos-io/osbuilder/actions/runs/3311250394/jobs/5466461592

mudler:
https://github.com/kairos-io/kairos/actions/runs/3355901718/jobs/5560531307

mudler:
https://github.com/kairos-io/kairos/actions/runs/3357122458/jobs/5562721753

mudler:
https://github.com/kairos-io/kairos/actions/runs/3360669745/jobs/5570168439

mudler:
https://github.com/kairos-io/kairos/actions/runs/3362375827/jobs/5574015725

mudler:
https://github.com/kairos-io/kairos/actions/runs/3380983339/jobs/5614381109

mudler:
https://github.com/kairos-io/kairos/actions/runs/3384839737/jobs/5622287645

mudler:
```

                +iso | Problem retrieving files from 'Non-OSS Repository'.
                +iso | Location 'http://download.opensuse.org/distribution/leap/15.4/repo/non-oss/media.1/media' is temporarily unaccessible.
                +iso | Please see the above error message for a hint.
                +iso | Warning: Skipping repository 'Non-OSS Repository' because of the above error.
                +iso | Problem retrieving files from 'Main Repository'.
                +iso | Location 'http://download.opensuse.org/distribution/leap/15.4/repo/oss/media.1/media' is temporarily unaccessible.
[...]

PR #731, opened 2023-01-25
:penguin: Remove nohang from core images

mudler:
Providers, or downstream images should add it instead

This needs a PR in provider-kairos, to add it in there instead. Potential fix for #265

### DIFF
--- a/images/Dockerfile.debian
+++ b/images/Dockerfile.debian
@@ -13,7 +13,6 @@ RUN apt install -y \
     grub-efi-amd64-bin \
     grub2 \
     grub2-common \
-    nohang \
     sudo \
     iproute2 \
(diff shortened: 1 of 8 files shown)


Itxaka (approved):
This kind of shifts the issue into custom builds but at least it's faster. Is nohang required for something specific or is it just nice to have?

mudler:
> This kind of shifts the issue into custom builds but at least it's faster. Is nohang required for something specific or is it just nice to have?



> This kind of shifts the issue into custom builds but at least it's faster. Is nohang required for something specific or is it just nice to have?

not really, was just a nice to have, but we can add it to the provider image instead, I'd like to keep the core as small as possible

PR #736, opened 2023-01-26
Revert "Update dependency codemirror to v6"

### LINKED ISSUE #732: :bug: WebUI syntax highlight is broken
**Kairos version:**
master

Seems the CodeMirror syntax highlighter is broken:
Screenshot from 2023-01-25 23-47-46
Screenshot from 2023-01-25 23-48-32

possibly caused by #610 ? cc @mauromorales

mauromorales:
Reverts kairos-io/kairos#610

The upgrade is a bit trickier than expected and I don't want to block a release because of it so I thought it would be better to revert while I look into it

Fixes #732

### DIFF
--- a/internal/webui/public/package-lock.json
+++ b/internal/webui/public/package-lock.json
@@ -9,88 +9,12 @@
         "@popperjs/core": "^2.11.6",
         "alpinejs": "^3.10.5",
         "bootstrap": "^5.0.0",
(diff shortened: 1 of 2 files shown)


mudler:
manually tested, all back to normal :+1: 
Screenshot from 2023-01-26 11-45-47

mmm maybe we could add some webui automated test ? :)

PR #742, opened 2023-01-26
require userdata loaded before running install

### LINKED ISSUE #696: :bug: kairos.service may start before cos-setup-network
When running in a cloud provider specifying a cloud config file (e.g. in openstack) the kairos.server (the installer) may start before the cos-setup-network service is done downloading cloud config files from the provider. In that case, our cloud-config is ignored and the installer continues by showing the QR code as if no cloud-config was provided.

Restarting the `kairos.service` with systemdctl makes it work (because now the cloud-config files are in place).

paynejacob:
Signed-off-by: Jacob Payne <jacob@spectrocloud.com>

Wait for userdata to load if the loading sentinel exists.

when PR gets merged)*:
Fixes #696


internal/agent/install.go
```diff
@@ -102,11 +102,13 @@ func Install(dir ...string) error {
 		}
 	})
 
-	// Try to pull userdata once more. best-effort
-	if _, err := os.Stat("/oem/userdata"); err != nil {
-		if err := machine.ExecuteCloudConfig("/system/oem/00_datasource.yaml", "rootfs.before"); err != nil {
-			fmt.Println("Warning: Failed pulling from datasources")
+	for {
```
mudler, line 105: this is looking good! maybe we should timeout here and not let the loop go forever?

internal/agent/install.go
```diff
@@ -102,11 +102,13 @@ func Install(dir ...string) error {
...
+		if _, err := os.Stat("/oem/.userdata_load"); os.IsNotExist(err) {
+			break
 		}
+
+		fmt.Println("userdata configuration has not yet completed.")
```
mudler, line 110: maybe we should write down what we are waiting for (for debugging reasons)

```suggestion
		fmt.Println("userdata configuration has not yet completed (waiting /oem/.userdata_load) .")
```

mudler (dismissed):

mudler:
giving it a re-thought. maybe we should move away from having the sentinel in `/oem`, but instead use `/run`?

internal/agent/install.go
paynejacob: That is a good idea!

internal/agent/install.go
paynejacob: I like the idea of more detail, ill give some context to what that file means as well.

Itxaka (approved):
niiice

PR #768, opened 2023-01-30
:bug: Fix tumbleweed iso

### LINKED ISSUE #763: CI: opensuse-tumbleweed is not passing qemu-acceptance-test
**Kairos version:**
master

N/A

**Describe the bug**
on build images workflow, iso is properly build but acceptance test are failing

**To Reproduce**

**Expected behavior**

**Logs**

**Additional context**

Itxaka:
Looks like tumbleweed has the same issue with overlayfs folders missing as Debian.

This patch applies the same patch as Debian uses to create 2 missing dirs to be able to use overlayfs.

Also renames the patch so its not debian specific.

Fixes #763


mudler (approved):
I hoped this fancy patch won't go forever, but it is even expanding =D 

Anyway, looks good here, :+1:

Itxaka:
Im just astounded that this is not an issue on leap but instead on tumbleweed....

PR #794, opened 2023-01-31
🔧 Add earthlyignore file

### LINKED ISSUE #693: Add and maintain .earthlyignore file

We need to add `.earthlyfile` to enhance build speed locally. Without this we are sending a lot of unneeded
files to buildkit, especially true for large ISOs under `build/`.

For a start an earthlyingore with:
```
build/*iso
```

Manually clean the build directory every time I build an new ISO.

**Additional context**

https://docs.earthly.dev/docs/earthfile/earthlyignore

oz123:
We need to add .earthlyfile to enhance build speed locally. Without this we are sending a lot of unneeded
files to buildkit, especially true for large ISOs under build/.

Fixes #693

### DIFF
--- a/.earthlyignore
+++ b/.earthlyignore
@@ -1,0 +1,3 @@
+build/*iso
+# this is created by the strat_vm_qemu.sh script
+disk.img
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -162,11 +162,17 @@ jobs:
       matrix:
        include:
          - flavor: "alpine-opensuse-leap"
+           port: 50001
          - flavor: "opensuse-leap"
+           port: 50002
          - flavor: "opensuse-tumbleweed"
+           port: 50003
          - flavor: "ubuntu"
+           port: 50004
          - flavor: "debian"
+           port: 50005
          - flavor: "ubuntu-20-lts"
+           port: 50006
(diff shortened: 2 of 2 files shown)


paynejacob (dismissed):

Itxaka (commented):
I would just add the whole build dir instead of only the is, as you may have img files or xz files on it

mudler:
upgrade tests seems to use `build` https://github.com/kairos-io/kairos/actions/runs/4057373981/jobs/6983975639

Itxaka:
> upgrade tests seems to use `build` [kairos-io/kairos/actions/runs/4057373981/jobs/6983975639](https://github.com/kairos-io/kairos/actions/runs/4057373981/jobs/6983975639)

weird, that downloads the required iso into the root dir. I would instead change the job to move the ISO into the test dir or a special dir instead of removing the earthly file on those tests...

oz123:
TBH, I think you are correct. It's the right thing to do, but I don't want to mess further with the Earthlyfile.
Every commit push cycle is really slow, and we can do this in a separate PR to make all the tests use the same directory.
Right now, I just want all the team to use the earthlyignore in the most basic version.
Once this is the master, everyone uses it, so we can incrementally improve it.

mudler (approved):

Issue #781, opened 2023-01-30
:bug: CI is failing waiting for userdata to be loaded


paynejacob:
https://github.com/kairos-io/kairos/actions/runs/4027327620/jobs/6950379445

mudler:
just cross-posting from the chat:

Might be that `/oem` is not clean now that we wait, and elemental complains about mounting `/oem` after partitioning - as it is not clean anymore?

mudler:
switching prio to high as blocks all PRs from being automatically merged on approvals

mudler:
just for history reference, started since https://github.com/kairos-io/kairos/pull/742

PR #795, opened 2023-01-31
:🔧: Allow port forwarding for qemu script

oz123:
This is similar to what happens when we create a VM with Vbox or with the Test suite.
Useful for developers who test with qemu.

Use with:
```
$ PORT_FORWARD=8222 ./scripts/start_vm_qemu.sh build/kairos.iso
```
When the machine is running you can then do:
```
$ ssh -l kairos -p 8222 127.0.0.1:8882
\# login with password kairos
```


Itxaka (approved):

mudler:
hit by #781

Issue #280, opened 2022-10-24
🐛 systemd-networkd cannot set hostname from DHCP, always set to localhost


ApolloDS:
**Kairos version:**
v1.1.3

**CPU architecture, OS, and Version:**
Running on VM with i686

**Describe the bug**
systemd-networkd cannot set hostname from DHCP, always set to localhost

**To Reproduce**
Check the logs:
journalctl -u systemd-networkd
...
Oct 24 19:40:06 localhost systemd[1]: Starting Network Configuration...
Oct 24 19:40:07 localhost systemd-networkd[1490]: ens19: Link UP
Oct 24 19:40:07 localhost systemd-networkd[1490]: ens19: Gained carrier
Oct 24 19:40:07 localhost systemd-networkd[1490]: ens18: Link UP
Oct 24 19:40:07 localhost systemd-networkd[1490]: ens18: Gained carrier
Oct 24 19:40:07 localhost systemd-networkd[1490]: lo: Link UP
Oct 24 19:40:07 localhost systemd-networkd[1490]: lo: Gained carrier
Oct 24 19:40:07 localhost systemd-networkd[1490]: ens19: Gained IPv6LL
Oct 24 19:40:07 localhost systemd-networkd[1490]: ens18: Gained IPv6LL
Oct 24 19:40:07 localhost systemd-networkd[1490]: Enumeration completed
Oct 24 19:40:07 localhost systemd[1]: Started Network Configuration.
Oct 24 19:40:08 localhost systemd-networkd[1490]: ens19: DHCPv4 address 10.11.12.13/24 via 10.11.12.1
Oct 24 19:40:08 localhost systemd-networkd[1490]: ens18: DHCPv4 address 192.168.1.13/24 via 192.168.1.1
Oct 24 19:40:08 localhost systemd-networkd[1490]: Could not set hostname: Access denied
Oct 24 19:40:08 localhost systemd-networkd[1490]: Could not set hostname: Access denied
Oct 24 19:40:08 localhost systemd-networkd[1490]: Could not set hostname: Access denied
Oct 24 19:40:22 localhost systemd-networkd[1490]: flannel.1: Link UP
Oct 24 19:40:22 localhost systemd-networkd[1490]: flannel.1: Gained carrier
Oct 24 19:40:23 localhost systemd-networkd[1490]: cni0: Link UP
Oct 24 19:40:23 localhost systemd-networkd[1490]: cni0: Gained carrier
[...]

PR #805, opened 2023-02-02
:bug: Fix ownership of polkitd rules dir

Itxaka:

Due to a bug on earthly it seems that some of the directories for polkitd are not being copied under the proper user. This results into polkit not being able to read the dir properly and the rules that come with it.

This patch works around it by installing an oem file that wil restore the proper ownership of those dirs

Partial #280


mudler (dismissed):
good catch, awesome!

Itxaka:
tested on opensuse and works, and results in fixing the hostname from dhcp issue. 

currently testing on ubuntu

mudler (approved):

mudler (approved):

PR #806, opened 2023-02-02
:bug: Fix dhcp not able to set hostname

Itxaka:
Ubuntu flavors were missing the polkitd service so systemd-networkd was not able to send the set-hostname command to systemd-hostnamed due to permissions.

Also once that was fixed, the system defaulted to the localhost.localdomain static name.

This patch add the polkitd package to ubuntu flavors and removes the existing /etc/hostname file which prevents systemd-networkd from setting the hostname via dhcp

Partially #280


mudler (approved):
Looking good!

mudler (approved):

PR #809, opened 2023-02-02
added install lock

### LINKED ISSUE #781: :bug: CI is failing waiting for userdata to be loaded

paynejacob:
Signed-off-by: Jacob Payne <jacob@spectrocloud.com>

Added install lock to ensure we don't have concurrent installations, the introduction of a wait for cloud config caused tests to call install while the kairos service was attempting to install at the same time.

Fixes #781


mudler:
Good catch! I think we should also fixup the tests, no? if they are automatically attempting install, no need to trigger manually (and vice versa) anyhow, the lock seems a good idea to introduce in any case

mudler (dismissed):

paynejacob:
@mudler I was thinking about this, I am not sure when we would know to reboot, I suppose we could just add a reboot flag to all the test cloud configurations.

mudler:
> @mudler I was thinking about this, I am not sure when we would know to reboot, I suppose we could just add a reboot flag to all the test cloud configurations.

looking closely at the test I think we didn't reboot automatically in order to detach the CD or the following reboot would boot that one in vbox. I guess then we should drop `install.auto: true` from the fixtures should be enough

mudler (dismissed):

mudler (approved):
Looks good here !

PR #810, opened 2023-02-03
:bug: Remove workaround for polkit dirs

Itxaka:

Looks like there was missing docs for the --keep-own flag on earthly and we can indeed keep permissions for all dirs/files when copying artifacts, but the flag needs to be both on the SAVE ARTIFACT and in the COPY stanzas.

Partially #280


mudler (approved):

Itxaka:
Not waiting for it to spend more ci cycles as all passed before, just a earthfile conflict

PR #814, opened 2023-02-03
🔧 merge custom mounts tests

oz123:
Since the feature is now merged, we no longer need a single label for this. This saves a few CI cycles. To acheive this, we add the correct sections to the normal install sections and assert that these paths exist.

The previous assertions are kept, but in order to make code a bit more readable, I extracted the anonymous function that collects the data from the installed VM.

Instead of repeating the installation 4 times, we now only repeat the installation 2 times.


mudler (dismissed):
Looking good, nice !

oz123:
> Looking good, nice !

I am still hesitant to merge this. It's seems the test still fail

mudler (approved):

PR #821, opened 2023-02-06
:penguin: Fix Some Lint Issues for Debian-based Images

sdwilsh:
**What this PR does / why we need it**:
This fixes [DL3027](https://github.com/hadolint/hadolint/wiki/DL3027).  That surfaces [DL3088](https://github.com/hadolint/hadolint/wiki/DL3008), which this change adds to the ignore file for now.

Along the way, it also fixes [DL3009](https://github.com/hadolint/hadolint/wiki/DL3009) and partially fixes [DL3059](https://github.com/hadolint/hadolint/wiki/DL3059) for the Debian-based images.

PR #830, opened 2023-02-07
🐧 `xz` Compress initrd Image

sdwilsh:
In the logs for debian images, `dracut` provides this warning:
```
dracut: *** Creating image file '/boot/initrd-6.1.0-3-amd64' ***
dracut: Cannot execute compression command 'xz', falling back to default
dracut: Using auto-determined compression method 'pigz'
```

This change adds `xz-utils` to debian images so that it can use `xz` compression.  This was first discovered during the work in #821.

### DIFF
--- a/images/Dockerfile.debian
+++ b/images/Dockerfile.debian
@@ -38,6 +38,7 @@ RUN apt install -y \
     sudo \
     systemd \
     systemd-sysv \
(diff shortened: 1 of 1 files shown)


mauromorales (approved):
lgtm

PR #831, opened 2023-02-07
🐧 Remove `RUN` commands that change `/etc/network/interfaces`

sdwilsh:
This file is not used by `systemd-networkd`, but these images enable `systemd-networkd`.  This change removes the `RUN` commands since the file should be unused.

This was originally identified in #821.


mauromorales (approved):
lgtm

PR #832, opened 2023-02-07
🔧 Use random free port for Spice display and SSH

oz123:
When running the install test with qemu, the second test will fail because the 2nd machine will fail to start.
The failure is cause because the port of the previous machine is still not freed.
Hence, let the install-test use a differnet random port if not specified.
`GetFreePorts` copied from `kairos-io/kcrypt-challenger`. Credits @jimmykarily.


jimmykarily:
This is still a hack. The tests mutate the one-and-only VM and then they destroy it an re-create it in the AfterEach block so that the next test finds it in pristine state.

This doesn't allow the tests to run in parallel. We addressed this problem here: https://github.com/spectrocloud/peg/pull/3 by introducing the `VM` class which wraps the same methods of the Machine but allows you to have more than one instances. You can see how this was consumed here: https://github.com/kairos-io/kcrypt-challenger/pull/11/files

I think this is what we should do in this test suite too.

mudler:
I guess we can also stop specifying ssh ports with the github actions workflow after this?

tests/install_test.go
```diff
@@ -44,14 +44,23 @@ func eventuallyAssert(cmd string, m types.GomegaMatcher) {
...
+		vm.EventuallyConnects(1200)
 	})
 
 	AfterEach(func() {
 		Machine.Clean()
+		mc := Machine.Config()
+		port, _ := getFreePort()
+		mc.Display = fmt.Sprintf("-vga qxl -spice port=%d,addr=127.0.0.1,disable-ticketing=yes", port)
+		port, _ = getFreePort()
+		mc.SSH.Port = fmt.Sprintf("%d", port)
```
Itxaka, line 63: wont this collide with the port above? After all its still free, as it hasnt been locked by anything yet?

oz123: True, in seldom cases it will collide. Didn't think about this.

Itxaka: wont this still potentially happens?

Let me see if I understand it correctly.

 - on line 60 you get a free port (one that its not currently used by anything).
 - then on line 62 you ask again for a free port.
 - then on line 64, the machine is created AND those ports are used at that moment (As soon as qemu/vbox locks those ports to be used on display/ssh)

What prevents `getFreePort` returning the same port on both calls? After all, for the kernel the port returned in the first calls is still free. So if you call it 100 times you may get it back 100 times the same port?
No matter that its opened by the `getFreePort` as its still closed as soon as the function returns.

So this can cause a collision by having both ports being the same one. I know, its rare but if it happens randomly that CI fails due to this we are gonna be puzzled.

I can show you the port collision does indeed happen: https://goplay.tools/snippet/qktKbufupVc
I got it to fail at the 2 port obtained, so its indeed a possibility that both ports returned are the same.

I would do a check for those and rerun `getFreePort` until I get a different port just in case :D

oz123: You are still correct. It's still containing code from the previous strcuture. I have a nil pointer error which I still don't understand. So I haven't fixed that redundant code.

Itxaka (approved):

oz123:
@Itxaka Gracias! Thank you for keeping your attention on the PR.

oz123:
@jimmykarily I had to migrate all the tests to the new PEG API, can you take a look here?

oz123:
@jimmykarily as discussed on huddle this PR is currently broken in part because the latest PEG changes don't work with VBOX.
Unfortunately, I can't finish this PR this week. If it is urgent for you, feel free to pick it up from here.

jimmykarily:
CI requirements are not satisfied because this PR removes the ports form the jobs and the jobs changed name (so the required jobs no longer exist). I will update the requirements as soon as we merge. In the meantime this PR is green and ready to review and merge

mauromorales (approved):
nice!

Itxaka (approved):

jimmykarily:
I squashed all commits into one because rebasing to master was a pain.

mudler (approved):

mudler (approved):

PR #840, opened 2023-02-08
🐧 Add `systemd-timesyncd` to all images that use `systemd`

sdwilsh:
The project has [documentation about using `timesyncd`](https://kairos.io/docs/reference/configuration/#timesyncd), but the reality is that not all images included it.  This change adds `systemd-timesyncd` to all machines that also install `systemd`.

Originally discovered while working on #821, but fixing for all `systemd`-based images here.


mauromorales (approved):
Good catch, lgtm

mauromorales:
I guess the big benefit of using chrony is that we would not have to maintain two packages/configs, so IMO it would make sense. 

@mudler do you have more context of why systemd-timesyncd was introduced?

mudler:
Thanks! that's a good catch!

> I guess the big benefit of using chrony is that we would not have to maintain two packages/configs, so IMO it would make sense.
> 
> @mudler do you have more context of why systemd-timesyncd was introduced?

That's actually the benefit of `systemd-timesyncd`.  Some distros ship that separately, some doesn't and it is just inside the same systemd package.

Note, `systemd-timesyncd` was introduced almost ~9 years ago: https://lists.freedesktop.org/archives/systemd-devel/2014-May/019537.html so I expect it to be present - if not, at some point it definetly should =).

I don't have big preference - our `cloud config` supports configuring it, however, it's not expected it to be present e.g. in distros that don't use `systemd`.

PR #841, opened 2023-02-08
🔧 Have renovate manage image versions in `Earthfile`

sdwilsh:
This change adds support for renovate to be able to update the versions of `quay.io/luet/base`, `golang`, and `hadolint`.

Tested with:
* `docker run --mount type=bind,source=$(pwd)/renovate.json,target=/usr/src/app/renovate.json -it renovate/renovate renovate-config-validator`
* https://regex101.com/r/Ax3iG9/1

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -4,6 +4,7 @@ ARG VARIANT=core # core, lite, framework
 ARG FLAVOR=opensuse-leap
 ARG IMAGE=quay.io/kairos/${VARIANT}-${FLAVOR}:latest
 ARG ISO_NAME=kairos-${VARIANT}-${FLAVOR}
+# renovate: datasource=docker depName=quay.io/luet/base
 ARG LUET_VERSION=0.34.0
 ARG OS_ID=kairos
 ARG REPOSITORIES_FILE=framework-profile.yaml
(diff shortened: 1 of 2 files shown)


mauromorales (approved):
fancy! LGTM

For the Go version I think we will most likely have to close until all dependencies are working well with 1.19 but that's ok

mudler:
loving it! thanks @sdwilsh !

Itxaka:
pretty cool, stealing this for other places :)

sdwilsh:
Looking at #310, it doesn't look like renovate has noticed this change, or it isn't working.  Annoyingly, they don't have a way to test this really, as far as I could tell from the docs.

sdwilsh:
We might want to change the `fileMatch` entry to better match the [regular expression they use for docker files](https://github.com/renovatebot/renovate/blob/7cf131021a770c15c8b108297c29d5edea89f352/lib/config/presets/internal/regex-managers.ts#L10), so it'd look something like this instead of what I used here: `(^|/|\\.)Earthfile$`.  It's also possible that it just hasn't run to pick up the changes yet, but I don't see that configured in the repository (it might be configured at the org level).  I'll give it some time to see if it gets picked up, and if not, we'll have to play with the config more.

sdwilsh:
I was wrong; this is working, but it's under regex, not docker:
image

So this is working :D

PR #846, opened 2023-02-08
:seedling: Don't swallow yaml config errors

jimmykarily:
because installation continues and there is no indication that something went wrong

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -355,7 +355,10 @@ func parseConfig(dir []string, nologs bool) *Config {
 				continue
 			}
 
-			yaml.Unmarshal(b, c) //nolint:errcheck
+			err = yaml.Unmarshal(b, c)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


mudler (dismissed):

pkg/config/config.go
```diff
@@ -355,7 +355,10 @@ func parseConfig(dir []string, nologs bool) *Config {
 				continue
 			}
 
-			yaml.Unmarshal(b, c) //nolint:errcheck
+			err = yaml.Unmarshal(b, c)
+			if err != nil {
```
mudler, line 359: there is `nologs` to respect here ^^

mudler:
```suggestion
			if err != nil && !nologs {
```

jimmykarily: correct! done.

Itxaka (approved):

PR #848, opened 2023-02-08
Validate renovate config file during lint

sdwilsh:
Now that the config is more than the stock config, and folks are interested in using it more, we should have some automation to verify that the config is valid.


mudler:
I just love how this is getting so meta with renovate bumping renovate! Thank you @sdwilsh !

mudler (approved):

mudler:
Merging as lint tests passed, no need to book CI for full e2e tests

PR #849, opened 2023-02-09
🔧 Manage the version of Earthly with renovate

sdwilsh:
This ensures that `earthly.ps1` and `earthly.sh` are using a version of Earthly that is managed with renovate.

Filename regex tested here: https://regex101.com/r/eEtcqU/1
Version regex tested here: https://regex101.com/r/T8T9aL/1


mauromorales (approved):
lgtm

PR #850, opened 2023-02-09
🔧 Mange the version of quay.io/kairos/osbuilder-tools with renovate

sdwilsh:
This should trigger an update in short order as the current version of this image is v0.5.2.


mauromorales (approved):
LGTM :sunglasses:

Issue #779, opened 2023-01-30
:seedling: Kairos config validator library


jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/778

Implement a library that validates Kairos yaml

- [x] Validates (#798)
  - yaml syntax
  - schema (non-existent keys, etc)
  - business logic (contradicting keys shouldn't be set etc)

Configuration reference: https://kairos.io/docs/reference/configuration/

Release process

- [x] #828
- [x] #899

## Followups

- #826 Improvement (not required)
- #827 Improvement (not required)
- #829 will be required for a followup release but has not been decided yet

Issue #778, opened 2023-01-30
:sparkles:  GUI Kairos YAML configurator


jimmykarily:
This splits step 2 from here: https://github.com/kairos-io/kairos/issues/772

## Acceptance Criteria

- There is a binary (new or existing one) that shows a GUI (web based or not)
- The GUI allows the user to create a Kairos config file by clicking on GUI elements
- For the first iteration only the [basic options](https://kairos.io/docs/reference/configuration/) should be implemented
- The result should be a yaml snippet which the user can copy and use to install Kairos
- [The kairos yaml schema validator](https://github.com/orgs/kairos-io/projects/1/views/1?pane=issue&itemId=19231235) should be used in the project's test suite. This should be the single point of truth, whenever we validate yamls (webui, user provided, whatever).
- There is a test suite for the UI

jimmykarily:
Extracted the validator here: https://github.com/orgs/kairos-io/projects/1/views/1?pane=issue&itemId=19231235

jimmykarily:
I don't think a GUI builder for kairos configuration brings much value. It's too much work for something which will be too easy to get out of sync and thus very error prone. I suggest that close this until someone comes with good arguments on why this is really needed. @kairos-io/maintainers ?

mauromorales:
@jimmykarily makes sense.

Closing but feel free to re-open if needed

jimmykarily:
btw, this is a lot more useful in my opinion: https://github.com/kairos-io/provider-kairos/pull/969

PR #853, opened 2023-02-09
:sparkles: Integrate schema validation

### LINKED ISSUE #828: Integrate new validation when running a configuration
As a first step, this should not break, but only warn the user. There can be a flag to enforce the validation. A message should be displayed telling the user that breakage will be the default configuration in the future.

mauromorales:

The main purpose of this PR is to warn the user when their configuration has schema validation errors

```
kairos@localhost:~> sudo kairos-agent manual-install --device /dev/vda config.yaml
WARNING: jsonschema: '/users' does not validate with file:///home/kairos/schema.json#/properties/users/minItems: minimum 1 items required, but found 0 items
INFO[2023-02-10T09:30:58Z] Starting elemental version 0.20221121.1+kairos
INFO[2023-02-10T09:30:58Z] Install called
INFO[2023-02-10T09:30:58Z] Partitioning device...
```

For users who want to start enforcing those validations there's a flag `strict-validation` which they can use globally and it will fail the command:

```
kairos@localhost:~> kairos-agent -strict-validation manual-install config.yaml
ERROR: jsonschema: '/users' does not validate with file:///home/kairos/schema.json#/properties/users/minItems: minimum 1 items required, but found 0 items
kairos@localhost:~> echo $?
1
```

Finally there are two additional commands added:

1. `validate-config` will show all the validation errors a config has
2. `print-schema` will print the JSON Schema which I plan to use to keep versioned schemas online e.g. https://kairos.io/v1.5.0/schema.json or something like that

I think it would be good to request that cloud config files say the version they are written in

Fixes #828 
Relates to #779 & #778


mauromorales:
found an issue while testing, so I converted it back to a draft

internal/agent/validate.go
```diff
@@ -0,0 +1,66 @@
...
+			return err
+		}
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		//Convert the body to type string
+		yaml = string(body)
+	} else {
+		dat, err := os.ReadFile(file)
```
Itxaka, line 40: should we check if its a dir? Will this even work if we pass a dir with the cloud config on a file in it?

mauromorales: Good question, the use case I thought was: "hey let me check if this particular config is valid" just like what we do on the WebUI, but I can definitely extend it to pass dirs too

Itxaka (approved):

PR #862, opened 2023-02-10
🐧 Packages in docker files should be on a single line

sdwilsh:
This is not something that hadolint checks (yet, see https://github.com/hadolint/hadolint/issues/445), but a good practice anyway.

### DIFF
--- a/images/Dockerfile.debian
+++ b/images/Dockerfile.debian
@@ -13,7 +13,9 @@ RUN apt install -y \
     curl \
     debianutils \
     dosfstools \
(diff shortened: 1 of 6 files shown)


mauromorales (approved):
yeah much easier to spot this way, too bad the linter doesn't pick it up

LGTM

sdwilsh:
> yeah much easier to spot this way, too bad the linter doesn't pick it up

I don't know Haskell, but I'm tempted to give it a try to implement 😅

PR #867, opened 2023-02-10
:robot: Run some jobs on self-hosted runners

mudler:

A small experiment

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -196,7 +196,7 @@ jobs:
   qemu-bundles-tests:
     needs: 
     - build
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


sdwilsh:
Is this looking for better turn-around time, better network connectivity, or something else?

mudler:
> Is this looking for better turn-around time, better network connectivity, or something else?

it is for both actually, I've added 4 self-hosted machines which have bigger specs than GH runners (network, cpu, and ram). 

This should increase our capacity generally and have faster builds, but planning just to enable it selectively for certain tests; so giving it a try here.. and seems already faster than public runners =)

Itxaka (approved):

Issue #212, opened 2022-10-12
QR code display on ubuntu is broken


mudler:
**Kairos version:**
1.4.0

**CPU architecture, OS, and Version:**
Ubuntu flavors only

**Describe the bug**
The QR code on boot cannot be displayed correctly

**To Reproduce**
Boot the ubuntu flavor

**Expected behavior**
Have the QR code visible such as can be scanned with the CLI

**Logs**

**Additional context**

jtcressy:
Is this the same issue as having odd diamond shapes in the QR code output? Just making sure so I don't accidentally file a duplicate issue. 

Here's an example of what I see when booting the ubuntu flavor from provider-kairos:
<details><summary>Expand</summary>

![image](https://user-images.githubusercontent.com/7663028/211125061-fa272439-fe64-468b-b499-0692ac8e0ad3.png)

</details>

mudler:
yes, correct 👍 that's the issue we are tracking here

jtcressy:
indeed, my workaround is editing the command line in grub at boot-time and removing `nomodeset`. It's odd that this argument works fine on all other flavors. must be something different about ubuntu's kernel display drivers.

jtcressy:
@mudler I made a discovery on this issue while working with my own derivative

Our solution may simply be to:
1. install the package `console-data` which provides the font file at `/usr/share/consolefonts/default8x16.psf.gz`
    - this file is special because it seems to actually provide a unicode map while every other font shipped with the OS by default seems not to. In fact, this should be similar to the console font used by opensuse by default.
2. Override the font set by `setupcon` via `/etc/default/console-setup` and set `FONT="default8x16.psf.gz"` somewhere in the file - this should override other settings in the file and anything set via `dpkg-reconfigure console-setup`
---
How I got here:

[...]

PR #874, opened 2023-02-12
🐧 [ubuntu] Remove some packages that should not be needed

sdwilsh:
In https://github.com/kairos-io/kairos/pull/821 a number of packages were explicitly added that were previously installed automatically.  This change removes some of the more obvious ones that we should not need to install manually.

### DIFF
--- a/images/Dockerfile.ubuntu
+++ b/images/Dockerfile.ubuntu
@@ -11,10 +11,8 @@ RUN apt-get update \
   && add-apt-repository ppa:oibaf/test \
   && apt-get update \
   && apt-get install -y --no-install-recommends \
-    binutils \
     conntrack \
     console-data \
-    console-setup \
     coreutils \
     cryptsetup \
     curl \
@@ -27,7 +25,6 @@ RUN apt-get update \
     dracut-network \
     e2fsprogs \
     efibootmgr \
-    ethtool \
     firmware-sof-signed \
     fuse3 \
     gawk \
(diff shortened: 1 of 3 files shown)


sdwilsh:
Here's the delta's on size of packages installed

| branch | ubuntu | ubuntu-22-lts | ubuntu-20-lts |
| --- | --- | --- | --- |
| master | 2061 MB | 1855 MB | 1739 MB |
| reduce-deps | 1898 MB (-7.9%) | 1699 MB (-8.4%) | 1521 MB (-2.3%) |

images/Dockerfile.ubuntu
```diff
@@ -11,10 +11,8 @@ RUN apt-get update \
   && add-apt-repository ppa:oibaf/test \
   && apt-get update \
   && apt-get install -y --no-install-recommends \
-    binutils \
     conntrack \
     console-data \
-    console-setup \
```
mudler, line 17: I think we need `console-setup`, IIRC for displaying correctly the QR code: #212

sdwilsh: Based on the PR that fixed that issue, we need `console-data`, which this change keeps.

`console-setup` is not recommended by `console-data` either: https://packages.ubuntu.com/kinetic/console-data

sdwilsh: I can't test this to verify it because I can't seem to build isos locally.  I've put more details in #875.

mauromorales: Just built a new image without `console-setup` and it shows the QR properly.

mauromorales (approved):
lgtm!

PR #876, opened 2023-02-13
:wrench:  Add temp-image Earthly target

mauromorales:
This is helpful to quickly generate temp images and upload them to ttl.sh plus you can quickly distinguish the images by branch and sha

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -672,3 +672,31 @@ docs:
     RUN npm run prepare
     RUN HUGO_ENV="production" hugo --gc -b "/local/" -d "public/local"
     SAVE ARTIFACT public /public AS LOCAL docs/public
+
+## ./earthly.sh --push +temp-image --FLAVOR=ubuntu
+## all same flags than the `docker` target plus 
+## - the EXPIRATION time, defaults to 24h
+## - the NAME of the image in ttl.sh, defaults to the branch name + short sha
+## the push flag is optional
+## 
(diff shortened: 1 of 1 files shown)


oz123 (approved):
LGTM. Polishing our tools :-).

Earthfile
```diff
@@ -672,3 +672,31 @@ docs:
...
+temp-image:
+    FROM alpine 
+    RUN apk add git
+    COPY . ./
+
+    IF [ "$EXPIRATION" = "" ]
+        ARG EXPIRATION="24h"
+    END
+
+    ARG BRANCH=$(git symbolic-ref --short HEAD)
```
mudler, line 692: can we maybe reuse the version target here?

mudler (approved):
Looking good!

Earthfile
mauromorales: good question ... here I was aiming for ease in finding in my local docker images so I can group all the temp images for this branch, or that other branch this also makes it easier to clean up once I know I'm no longer working on that branch ... but I'm open to change it if version is more useful for everyone, or to have a way of choosing one or the other cuz I do agree that it might actually be helpful to have there the flavor and the version as well but then the name might be a bit too long?

Itxaka (approved):
nais

mudler:
No need to book CI cycles, merging

Issue #436, opened 2022-11-16
Add functional tests for dracut modules 


mudler:
**Describe the solution you'd like**
We miss tests from the basic lifecycle functionalities such as :
- boot assessment
- fallback mechanism
- hooks
- recovery

**Additional context**
#116

PR #877, opened 2023-02-13
:warning: :art: :sparkles: Use immucore :hedgehog:

Itxaka:
Build kairos with immmucore instead of cos-immutable-rootfs

Fixes #


Itxaka:
Really interested in seeing other base os running the tests. Currently I have only tested opensuse...

docs/content/en/docs/Reference/build.md
```diff
@@ -139,13 +139,14 @@ RUN mkdir -p /etc/luet/repos.conf.d && \
...
         system/base-cloud-config \
-        dracut/immutable-rootfs \
+        dracut/immucore \
         dracut/network \
         static/grub-config \
         system/suc-upgrade \
         system/shim \
         system/grub2-efi \
         system/elemental-cli \
+        system/immucore \
```
mauromorales, line 149: what's the difference between dracut and system immucore?

Itxaka: Stupid difference, I just made two different packages for no reason. As immucore is golang, the system/immucore generates the binary only while the dracut/immucore only brings the dracut files and systems services.

In the older cos-immutable-rootfs the cos-mount script was bundled in the dracut package. 

Maybe we should do the same here and bundle everything in the same package....

mauromorales: oki, doesn't need to block this PR I think, but it does sound a bit easier to dive in if it's all in one package

mauromorales (dismissed):
lgtm

docs/content/en/docs/Reference/build.md
Itxaka: I guess that with UKI + no pivot as we don't use dracut it may be handy to use only the binary, same with no systemd systems but I haven't tested those yet to be honest

mudler:
This should be requiring a major version bump of Kairos e.g. 2.0 to signal such a impactful change

mudler (changes requested):
I'm ok with this - just placeholder ro avoid to be merged now (in case tests are passing?)

Itxaka:
> I'm ok with this and really happy to see it in - just placeholder ro avoid to be merged now (in case tests are passing?)

Don't worry, I can still force push it 😈

mudler:
I just love how all emojis can be applied to this PR so nicely :dancers:

Itxaka:
cc @mudler ...green... :D :hedgehog:

Itxaka:
Recovery doesnt work...which makes me guess why we dont have any recovery tests...

mudler:
> Recovery doesnt work...which makes me guess why we dont have any recovery tests...

https://github.com/kairos-io/kairos/issues/436

mudler (approved):
Nais! looking good here!

Itxaka:
Now includes https://github.com/kairos-io/kairos/pull/979 which should fix the ubuntu 20-lts acceptance test

mudler (dismissed):

Itxaka:
temp config file is not created under /tmp but code doesnt fail. Also running elemental with --logfile into /tmp also did not create any file there....

Itxaka:
woops, immucore should not mount /tmp :D

mudler (approved):
Awesome work! let's go!

PR #893, opened 2023-02-15
🌱 Added support for Amazon SSM Agent

3pings:
Added support for Amazon SSM Agent config file persistence.

Users that are looking to add SSM support for their nodes need /etc/amazon/ssm to be persistent as this is where the connection information to AWS is stored.  Previously users could snap install amazon-ssm-agent but the configuration file is not stored in the snaps directory.

N/A


mudler:
Custom mounts should be available within the next release as of #210 (thanks to @oz123 ) so we can also drop some of them which aren't really required here - would that be enough and cover your case here as well? 

I think Amazon SSM support is quite specific and doesn't really fit the core images

3pings:
Custom mounts will solve the issue.  I am closing this PR.

PR #895, opened 2023-02-15
🐧 Install ubuntu-advantage-tools

3pings:
Ubuntu advantage tools allows for users to register their node with Canonical.  Since Snaps are supported this will allow them to enable Live Kernel Patches.

Ubuntu Advantage Tools enables registration with Canonical for support and things like Live Kernel Patches.

N/A

### DIFF
--- a/images/Dockerfile.ubuntu-22-lts
+++ b/images/Dockerfile.ubuntu-22-lts
@@ -64,6 +64,7 @@ RUN apt-get update \
     systemd-hwe-hwdb \
     systemd-timesyncd \
     thermald \
+    ubuntu-advantage-tools \
     unattended-upgrades \
     xdg-user-dirs \
     xxd \


Itxaka (approved):

sdwilsh:
Would this apply to other versions of Ubuntu that are also supported?  Also, does supporting snaps require the `snapd` package that was removed in #874?

mudler:
> Would this apply to other versions of Ubuntu that are also supported? Also, does supporting snaps require the `snapd` package that was removed in #874?

ouch, that was an oversight. We definitely want snapd support to allow snaps packages to be installed alongside.

sdwilsh:
I filed #930 so we can hopefully catch this kind of thing with automation in the future.  It's bound to happen if we leave it up to us fallible humans to catch this stuff!

PR #910, opened 2023-02-17
:wrench: Lint overlay files

mauromorales:

Better readability for our overlay files, and make sure CI catches any linting issues in them


Itxaka (approved):
that its cool!

PR #912, opened 2023-02-17
:bug: Add record to hosts file when missing

### LINKED ISSUE #887: :bug: `hostname` does not work on debian
**Kairos version:** 1.5.1

**CPU architecture, OS, and Version:** amd64, kairos-debian-v1.5.1-k3sv1.26.1+k3s1

**Describe the bug**

I am having issues using the hostname configuration option as per [documentation](https://kairos.io/docs/reference/configuration/#hostname).
If I use this option the file `/etc/hosts` is not created.
I have tried also with:
```yaml
#cloud-config

hostname: "your_hostname_here"
```
It sets the hostname, but does not create the `/etc/hosts` file.

**To Reproduce**

Install kairos with the ISO **kairos-debian-v1.5.1-k3sv1.26.1+k3s1** and apply the a configuration like the following one:
```yaml
#cloud-config

hostname: test
# Define the user accounts on the node.
users:
- name: "kairos"
  passwd: "kairos"
  ssh_authorized_keys:
  - "ssh-rsa AAAA..."

# Enable K3s on the node.
k3s:
  enabled: true
  args:
  - --disable=traefik,servicelb

stages:
  initramfs:
    - files:
        - path: /etc/systemd/network/10-config.network
          permissions: 0644
          content: |
            [Match]
            Name=enp1s0

            [Network]
            Address=192.168.11.55/24
            Gateway=192.168.11.250
            DNS=192.168.1.181 8.8.8.8
  boot:
    - name: "Setup hostname"
      hostname: "test"
```

mauromorales:

For debian based distributions, /etc/hosts is present but empty. This is because the file is populated when running the configuration. [1] This PR inserts a record before initramfs so it can be manipulated later on by yip's hostname plugin

[1] https://wiki.debian.org/ConfigPackages

Fixes #887

### DIFF
--- a/overlay/files/system/oem/31_hosts.yaml
+++ b/overlay/files/system/oem/31_hosts.yaml
@@ -1,0 +1,11 @@
+stages:
+  initramfs.before:
+    # For debian based distributions, /etc/hosts is present but empty. This is because the file
+    # is populated when running the configuration. For those cases we insert a record so it can be
+    # manipulated later on by yip's hostname plugin
+    # Read more: https://wiki.debian.org/ConfigPackages
+    - name: "Make sure hosts file is present and includes a record for 127.0.0.1"
+      if: |
+        ! [[ -f /etc/hosts ]] || ! [[ $(grep '127.0.0.1' /etc/hosts) ]]
+      commands:
+        - echo '127.0.0.1\tlocalhost' >> /etc/hosts


mudler:
I think the real issue is that there isn't an `/etc/hosts` file at all : https://img.kairos-labs.io/quay.io/kairos/core-debian/etc/

indeed https://img.kairos-labs.io/quay.io/kairos/core-debian/etc/hosts returns 404, maybe would be enough to touch it in the Dockerfile? I don't remember at the moment if it's mounted and get lost in the process, but worth a try

mauromorales:
> I think the real issue is that there isn't an `/etc/hosts` file at all : https://img.kairos-labs.io/quay.io/kairos/core-debian/etc/
> 
> indeed https://img.kairos-labs.io/quay.io/kairos/core-debian/etc/hosts returns 404, maybe would be enough to touch it in the Dockerfile? I don't remember at the moment if it's mounted and get lost in the process, but worth a try

@mudler What is this sorcery? I didn't know we could check the contents of the images online, very cool! 

If I add this to the Dockerfile I get the following error, am I missing something?

```
--> RUN touch /etc/hosts
touch: cannot touch '/etc/hosts': Read-only file system
```

mudler (approved):

PR #927, opened 2023-02-20
:bug: Fix job "Push latest ARM images" caused by docs target

mauromorales:
The problem was caused because the outer target runs on arm64 but when it gets to the docs target the `USERARCH` was amd64. This change checks also for `MODEL` and in case the `USERARCH` is `linux/arm64` and not just `arm64`

you can see it working properly in this run https://github.com/mauromorales/kairos/actions/runs/4223111756/jobs/7332444446#step:9:1170 (cancelled manually after it passed, so ignore that it is red :tongue: )

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -685,6 +685,7 @@ docs:
           rm -r hugo_extended_${HUGO_VERSION}_linux-${USERARCH}.tar.gz
     RUN npm install postcss-cli
     RUN npm run prepare
+    RUN ls -las .
     RUN HUGO_ENV="production" ./hugo --gc -b "/local/" -d "public/local"
     SAVE ARTIFACT public /public AS LOCAL docs/public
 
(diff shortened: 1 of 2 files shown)


Earthfile
```diff
@@ -674,17 +674,28 @@ webui-deps:
 docs:
     FROM node:19-bullseye
     ARG USERARCH
+    ARG MODEL
+    IF [ "$MODEL" = "rpi64" ] || [ "$USERARCH" = "arm64" ] || [ "$USERARCH" = "linux/arm64" ]
```
Itxaka, line 678: you should be able to use the auto-define `TARGETARCH` (arm, amd64, arm64) which is set from earthly directly to identify the arch:
https://docs.earthly.dev/docs/earthfile/builtin-args

Itxaka (dismissed):
good catch!

Earthfile
mauromorales: Oh I didn't see this one, thanks!

mudler (approved):
looking good!

Issue #771, opened 2023-01-30
🏗  infra dogfooding


mudler:
A  kairos cluster for the team to maintain, keep up to date and use for services (on need)

Use cases:

- Provide public accessible services ( e.g. provide ARM images to download via http )
- Pairing sessions
- Shared IDE (gitpod, e.g.)
- KubeVirt for Github actions workers
- Good resource and testing for blog posts!
 
From the moment is up, it should be part of our release process to test it in the infra (up to the interrupt)

## Acceptance criteria

- [x] There is a cluster (!)
- [x] Documentation how to access it, maintain it
- [x] Services:
  - [x] GH action on Kubevirt
  - [ ] Slack bot (@Interrupt + @DailyBot, ask @mudler)
  - [x] containerbay for OCI-based artifacts (e.g. ARM)
  - [ ] DNS Settings
  - [ ] ARM Hybrid Kubernetes workers for testing (?)
  - [ ] [Jitsi for our meetings](https://github.com/jitsi-contrib/jitsi-helm)

mudler:
@jimmykarily I think we just need to setup fluxcd as a missing piece?

jimmykarily:
Installed fluxcd on the "services" cluster and moved containerbay to a kustomization. It will now install automatically when a cluster is bootstrapped as described on the README.

TODO:
- Install fluxcd on the "CI" cluster too and migrate the GA workers to a kustomization. This might be tricky since it needs a fresh token (they expire after some time).

jimmykarily:
> Installed fluxcd on the "services" cluster and moved containerbay to a kustomization. It will now install automatically when a cluster is bootstrapped as described on the README.
> 
> TODO:
> 
>     * Install fluxcd on the "CI" cluster too and migrate the GA workers to a kustomization. This might be tricky since it needs a fresh token (they expire after some time).

Deployed org level runners following this: https://github.com/actions/actions-runner-controller/blob/master/docs/choosing-runner-destination.md#adding-runners-to-an-organization

We now have 5 runners for the whole kairos-io organization. We can probable scale it up a bit but let's see first how it behaves with 5.

jimmykarily:
On kcrypt-challenger we are using the self hosted runners for the e2e tests. The build time dropped to 50%:

https://github.com/kairos-io/kcrypt-challenger/actions/runs/4222474847

![image](https://user-images.githubusercontent.com/2794419/220111791-98189580-2985-4d1b-9b55-53755ca8d08b.png)

[...]

PR #929, opened 2023-02-20
Use KVM on tests that run on self-hosted runners

jimmykarily:
to make the run faster

Part of #771

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -210,13 +210,36 @@ jobs:
       uses: actions/download-artifact@v3
       with:
         name: kairos-${{ matrix.flavor }}.iso.zip
-    - run: |
+    - env:
+        KVM: true
+      run: |
             ls -liah
             export ISO=$PWD/$(ls *.iso)
             mkdir build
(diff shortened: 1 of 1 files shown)


.github/workflows/image.yaml
```diff
@@ -210,13 +210,35 @@ jobs:
...
+            global:
+              buildkit_additional_config: |
+                [registry."docker.io"]
+                  mirrors = ["registry.docker-mirror.svc.cluster.local:5000"]
+                [registry."registry.docker-mirror.svc.cluster.local:5000"]
+                  insecure = true
+            EOF
+
+            earthly.sh +prepare-bundles-tests
+            earthly.sh +run-qemu-bundles-tests --FLAVOR=${{ matrix.flavor }}
```
mauromorales, line 241: shouldn't these be just `earthly` without the script?

jimmykarily: oops, yes!

mauromorales (approved):
nice :sunglasses:

jimmykarily:
KVM is not used when using earthly. This needs some more work to run tests without earthly.

PR #937, opened 2023-02-21
Fix elemental grub with newer versions of elemental cli

### LINKED ISSUE #733: :bug: Reset fails when encryption is enabled
**Kairos version:**
[maser build](https://github.com/kairos-io/kairos/commit/2ada227c63ad1501173469ed52bb540c54fa64dc)

**Describe the bug**
When trying to reset the machine, elemental can't find anymore `COS_PERSISTENT` (as it is encrypted), so the reset process skips deleting the `COS_PERSISTENT` partition

**To Reproduce**
Enable encryption, try to reset the machine

**Expected behavior**
The `COS_PERSISTENT` partition should be removed and recreated as usual. the `kcrypt` hook kicks in afterward to encrypt the partition with the passphrase.

**Logs**
VirtualBox_test-encryption_26_01_2023_09_47_53

paynejacob:

This pr adds a grub cfg to /boot/grub. Some distros still look in `/boot/grub` even with grub2. This change introduces a file that redirects grub to always use the grub 2 folder. 

Fixes #733

### DIFF
--- a/internal/agent/hooks/hook.go
+++ b/internal/agent/hooks/hook.go
@@ -17,9 +17,7 @@ var AfterInstall = []Interface{
 	&Lifecycle{}, // Handles poweroff/reboot by config options
 }
 
-var AfterReset = []Interface{
-	&Kcrypt{},
-}
+var AfterReset []Interface
 
 var FirstBoot = []Interface{
 	&BundlePostInstall{},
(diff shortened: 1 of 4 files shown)


Itxaka:
This only applies if we update to a newer elemental-cli rigth? With our fork, grub should just work the same as it was forked especially because of that (also they dropped hybrid isos)

paynejacob:
The grub changes are no longer needed, we do still need to remove the kcrypt call after reset.  ill update open a separate pr for this change.

PR #943, opened 2023-02-22
:art: Refactoring of validator and config 1/n

mauromorales:
This PR is the first of N, in order to use a single schema structure to do the parsing and validation of the cloud config files. The change cannot be done in a single PR because there are external dependencies in Kcrypt, for this reason I think I'll need to open all these PRs (I think only this one will be this big and tricky :crossed_fingers: ):

1. this one to introduce KScan, KRun and other temporary functions, and move everything in this repo to consume them
2. change Kcrypt and any other external sources to consume the temporary functions
3. switch the original functions to be pretty much the same as the new temporary functions
4. re-adapt Kcrypt and any other external sources
5. remove everyhing temporary and any other cleaning

Before this can be merged, the multiple scenarios schemas need to be QAed:

- [x] PowerManagement
  - [x] manual-install config specified reboot, poweroff
  - [x] manual-install flag specified reboot, poweroff
  - [x] webui config specified reboot, poweroff
  - [x] webui checkbox specified reboot, poweroff
- [x] P2PNetworkExtended

### DIFF
--- a/internal/agent/hooks/gruboptions.go
+++ b/internal/agent/hooks/gruboptions.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	config "github.com/kairos-io/kairos/pkg/config"
+	schema "github.com/kairos-io/kairos/pkg/config/schemas"
 	"github.com/kairos-io/kairos/sdk/system"
 )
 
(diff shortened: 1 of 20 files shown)


Itxaka (commented):
I know its on draft still, but looking good to me

mauromorales:
did some tests and both scenarios which I was a bit concerned seem to be working properly. I managed to do the reboot and poweroff in multiple different ways and was able to setup a cluster with p2p networking. Haivng said that if anyone thinks is worth doing a bit more testing, please let me know or if you can try some weird setup you have and think might break it please do so and ping me so I can fix it

internal/agent/hooks/gruboptions.go
```diff
@@ -17,6 +18,19 @@ func (b GrubOptions) Run(c config.Config) error {
 	return nil
 }
 
+// KRun is a temporary function that does the same as Run. It will be removed as soon as the transition from config.Config to schema.KConfig is finished.
+func (b GrubOptions) KRun(kc schema.KConfig) error {
```
mudler, line 22: I'm trying to understand the context - it is hard to follow up with dups methods. Why the duplication? can't we try this directly modifying the configurations here?

mauromorales: yeah sorry, wasn't sure what the best approach for the refactoring was. I wanted to switch directly but there's one dependency in [kcrypt](https://github.com/kairos-io/kcrypt/blob/fa9eb79998b3642576ac26d49c133bbb9d372f1f/pkg/config/config.go#L50) so in the end I decided to go with the duplicated methods. But I don't mind approaching it differently if you think is best it just felt easier to have both for a while, switch everything including kcrypt and then end up with just one version of them again

mauromorales:
Decided to go with a different approach in #1037

PR #944, opened 2023-02-22
:bug: Fix custom binds set as RW_PATHS

Itxaka:
RW_PATHS are meant for overlay dirs which go away after a reboot. Custom binds/binds are mounted under COS_PERSISTENT, so they persist after reboot AND are RW by default.

This patch removes adding the custom binds into the RW_PATHS on the cos-layout file as that can lead to unintended consequences

Fixes #

### DIFF
--- a/overlay/files/system/oem/11_persistency.yaml
+++ b/overlay/files/system/oem/11_persistency.yaml
@@ -44,7 +44,6 @@ stages:
              source /run/cos/cos-layout.env
              source /run/cos/custom-layout.env
              PERSISTENT_STATE_PATHS="${CUSTOM_EPHEMERAL_MOUNTS} ${PERSISTENT_STATE_PATHS}"
-             RW_PATHS="${CUSTOM_BIND_MOUNTS} ${RW_PATHS}"
              echo CUSTOM_BIND_MOUNTS=\"${CUSTOM_BIND_MOUNTS}\" >> /run/cos/cos-layout.env
              echo CUSTOM_EPHEMERAL_MOUNTS=\"${CUSTOM_EPHEMERAL_MOUNTS}\" >> /run/cos/cos-layout.env
              echo "# rw paths with user bind mounts" >> /run/cos/cos-layout.env


mauromorales (approved):
LGTM

Itxaka:
This also sets the defaults for recovery to be sure that we have a proper cos-layout file AND prevents the custom layout for running on anything that its not active/passive

Itxaka:
Umm this still doesnt fix the bind mounts being set properly

overlay/files/system/oem/11_persistency.yaml
```diff
@@ -36,15 +36,21 @@ stages:
...
+        VOLUMES: "LABEL=COS_OEM:/oem"
+        OVERLAY: "tmpfs:25%"
+    - if: '[ -r /run/cos/custom-layout.env ] && [ ! -f "/run/cos/recovery_mode" ] && [ ! -f /run/cos/live_mode ]'
       name: "append custom bind and ephemeral mounts to /run/cos/cos-layout.env"
       commands:
         - |
              source /run/cos/cos-layout.env
              source /run/cos/custom-layout.env
              PERSISTENT_STATE_PATHS="${CUSTOM_EPHEMERAL_MOUNTS} ${PERSISTENT_STATE_PATHS}"
-             RW_PATHS="${CUSTOM_BIND_MOUNTS} ${RW_PATHS}"
```
mudler, line 47: is this going to be read by immucore? the custom mount is injecting RW_PATHS in this case to provide extra ephemeral paths

overlay/files/system/oem/11_persistency.yaml
```diff
@@ -36,15 +36,21 @@ stages:
           /usr/share/pki/trust/anchors
           /var/lib/ca-certificates
         PERSISTENT_STATE_BIND: "true"
-    - if: |
-        [ -r /run/cos/custom-layout.env ]
+    - if: '[ -f "/run/cos/recovery_mode" ]'
```
mudler, line 39: currently defined here: https://github.com/kairos-io/packages/blob/2ea0ddd0ed30d393b34c1d0356aedb634eba9031/packages/system/cloud-config/oem/00_rootfs.yaml#L16

overlay/files/system/oem/11_persistency.yaml
Itxaka: this is wrong, it needs to inject the ephemeral paths but its doing bind mounts! There is another patch incoming that reworks this, currently testing it locally

mudler: ouch!

overlay/files/system/oem/11_persistency.yaml
mudler: I think actually we have an issue to have all the configs in one place only, part of the #116 epic

mudler (approved):

Issue #755, opened 2023-01-27
:sparkles: Publish images for BYOI


mudler:
Currently we have cloud config files split in [packages](https://github.com/kairos-io/packages/tree/main/packages/system/cloud-config) and in the [various overlay folders](https://github.com/kairos-io/kairos/tree/498b8ec7fa85cddfcea675393553628b85c76264/overlay/files/system/oem)
Currently the files under `overlay` are per-flavor.

## Problem

When convering images to Kairos (Image onboarding), there is no way to include the kairos overlay files, and, especially without being distro specific.

We want those files to be flavor-independant so can be shipped with the `framework` images, and publish single images so those can be included. This is also true for the agent binary, that currently is not published anywhere

## Acceptance criteria
- [ ] https://github.com/kairos-io/kairos/pull/956 - Introduce the necessary changes in the agent to identify a system, so the cloud configs can leverage the agent to known which flavor is
- [ ] unify `overlay/files-*`, and package it as part of framework images
- [ ] Merge cloud-config logics between flavors, inside one single file overlay from the packages repo, and kairos
- [ ] Have a published, container image with cloud configs
- [ ] Have a published, container image with the agent binary
- [ ] Have a published, container image with the framework
- [ ] Check it doesn't break any flavor
- [ ] QA
- [ ] Add the overlay/files inside the framework images

## Notes

This is for instance, a Dockerfile that should consume a framework image (and note overlay files should be included)

```Dockerfile
FROM alpine as build
# This is to be fixed in the framework image still:
RUN apk add git
RUN git clone https://github.com/kairos-io/kairos.git /kairos

FROM fedora:36

COPY --from=quay.io/kairos/framework:v1.4.0 / /
# with kernel:
# COPY --from=quay.io/kairos/framework-generic:v1.4.0 / /

# <------> THIS shouldn't be required!
COPY --from=build /kairos/overlay/files /
COPY --from=build /kairos/overlay/files-fedora /

RUN echo "install_weak_deps=False" >> /etc/dnf/dnf.conf

RUN dnf install -y \
    NetworkManager \
    squashfs-tools \ 
    dracut-live \
    livecd-tools \
    dracut-squash \
    dracut-network \
    efibootmgr \
    dhclient \
    audit \
    sudo \
    systemd \
    parted \
    dracut \
    e2fsprogs \
    dosfstools \
    coreutils \
    device-mapper \
    grub2 \
    which \
    curl \
    nano \
    gawk \
    haveged \
    tar \
[...]

PR #956, opened 2023-02-24
:seedling: Detect more information about runtime

mudler:
This introduces a `system` and a `kairos` block available in kairos-agent get state.

This allows for instance to query the agent for the kairos version as such:

```bash
$ kairos-agent state get kairos.version
v1.5.1
$ kairos-agent state get kairos.flavor
opensuse-tumbleweed
$ kairos-agent state get system.os.vendor
kairos
```

Part of #755


go.mod
```diff
@@ -25,10 +25,12 @@ require (
 	github.com/pterm/pterm v0.12.54
 	github.com/qeesung/image2ascii v1.0.1
 	github.com/santhosh-tekuri/jsonschema/v5 v5.2.0
+	github.com/spectrocloud-labs/herd v0.3.0
```
Itxaka, line 28: Not sure why herd is pulled, I can't see it used anywhere?

mudler: no idea to be honest.. good point, I've just run go mod tidy..

mudler: running it again doesn't remove it, so must be throwed in by something

mudler: ah, maybe we bumped yip, that uses herd now

Itxaka: Ha, interesting, I would expect it to be marked as indirect. The more I know about go deps the less I understand them lol

Itxaka (approved):

Itxaka (approved):

Issue #949, opened 2023-02-23
:bug: Ubuntu 20-lts doesnt boot the iso sometimes.


Itxaka:
```
[   12.122339] Run /init as init process
[   12.212128] systemd[1]: Inserted module 'autofs4'
[   12.258607] systemd[1]: systemd 245.4-4ubuntu3.19 running in system mode. (+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid)
[   12.265975] systemd[1]: Detected virtualization qemu.
[   12.269242] systemd[1]: Detected architecture x86-64.
[   12.272447] systemd[1]: Running in initial RAM disk.

Welcome to kairos-core-ubuntu-20-lts v1.5.0-114-g9620f71 dracut- (Initramfs)!

[   12.332070] systemd[1]: No hostname configured.
[   12.335550] systemd[1]: Set hostname to <localhost>.
[   12.338858] random: systemd: uninitialized urandom read (16 bytes read)
[   12.342228] systemd[1]: Initializing machine ID from random generator.
[   12.615796] random: ln: uninitialized urandom read (6 bytes read)
[   12.642638] random: ln: uninitialized urandom read (6 bytes read)
[   13.084970] systemd[1]: sysroot.mount: Found ordering cycle on dracut-pre-mount.service/start
[   13.088456] systemd[1]: sysroot.mount: Found dependency on dracut-initqueue.service/start
[   13.091723] systemd[1]: sysroot.mount: Found dependency on iscsid.service/start
[   13.094852] systemd[1]: sysroot.mount: Found dependency on iscsid.socket/start
[   13.098027] systemd[1]: sysroot.mount: Found dependency on sysinit.target/start
[   13.101147] systemd[1]: sysroot.mount: Found dependency on systemd-tmpfiles-setup.service/start
[   13.104288] systemd[1]: sysroot.mount: Found dependency on local-fs.target/start
[   13.109324] systemd[1]: sysroot.mount: Found dependency on sysroot.mount/start
[   13.112387] systemd[1]: sysroot.mount: Job dracut-pre-mount.service/start deleted to break ordering cycle starting with sysroot.mount/start
[...]

PR #959, opened 2023-02-24
:bug: Add back gdisk to ubuntu/debian images

mudler:
**What this PR does / why we need it**: Otherwise EFI installs are broken

See also https://github.com/kairos-io/kairos/issues/949 and https://github.com/kairos-io/kairos/pull/874


Itxaka (approved):
Checked yesterday and this works :+1:

jimmykarily (approved):

PR #972, opened 2023-02-27
:seedling: Unify flavor-specific logic

mudler:
Unify overlay configuration by detecting in runtime the flavor and write down the necessary pieces. 

This is a WIP still, as I'd prefer to pin this down to a specific configuration supplied by the user too. I think detection is nice, but we should give a way to set this up explictly during building the image in the BYOI phase.

Part of #755


mudler:
mmm I don't really like it. This ends up that the user must specify the flavor in the os-release. On the other hand, the os-release file to be changed as well to properly write down the OS version.

mudler:
The alternative would be to ship framework files for each flavor, with and without the kernel

mudler:
Superseded by #973

PR #975, opened 2023-02-28
✨  add support to upgrade bundles with runbundles

### LINKED ISSUE #974: ✨  add support to upgrade bundles with runbundles
bundles.RunBundles luet installer as of now only supports luet installs, it would be good to have a upgrade support to it.

Add support for luet upgrade conditionally and run `luet repo update` before the upgrade.

force install and executing with with container runner.

**Additional context**

santhoshdaivajna:
bundles.RunBundles luet installer as of now only supports luet installs. This PR adds support to perform upgrades.

Fixes #974

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -62,6 +62,7 @@ type Bundle struct {
 	DB         string   `yaml:"db_path,omitempty"`
 	LocalFile  bool     `yaml:"local_file,omitempty"`
 	Targets    []string `yaml:"targets,omitempty"`
+	Upgrade    bool     `yaml:"upgrade,omitempty"`
 }
 
 const DefaultHeader = "#cloud-config"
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -62,6 +62,7 @@ type Bundle struct {
 	DB         string   `yaml:"db_path,omitempty"`
 	LocalFile  bool     `yaml:"local_file,omitempty"`
 	Targets    []string `yaml:"targets,omitempty"`
+	Upgrade    bool     `yaml:"upgrade,omitempty"`
```
mudler, line 65: I don't think we really need an "upgrade" option here, but rather make the install idempotent and upgrade the bundle. Also, having an upgrade setting for a first-deployment would be quite odd

mudler (changes requested):

pkg/config/config.go
santhoshdaivajna: >  make the install idempotent and upgrade
how do I go about this one 

`luet search` for the package and if its already present switch to `luet upgrade` ?

PR #977, opened 2023-02-28
:book: Add cloud-config.json for v1.6.0

mauromorales:
We still had the issue with the job while 1.6 was released, so adding this one manually too, hopefully it's the last one :crossed_fingers: :stuck_out_tongue:


mudler (approved):
Looks good here ! :+1:

PR #982, opened 2023-02-28
:bug: remove kcrypt hook after reset

### LINKED ISSUE #733: :bug: Reset fails when encryption is enabled

paynejacob:

With the changes to elemental the elemental cli can now clear content from encrypted partitions.

Fixes #733

### DIFF
--- a/internal/agent/hooks/hook.go
+++ b/internal/agent/hooks/hook.go
@@ -17,9 +17,7 @@ var AfterInstall = []Interface{
 	&Lifecycle{}, // Handles poweroff/reboot by config options
 }
 
-var AfterReset = []Interface{
-	&Kcrypt{},
-}
+var AfterReset = []Interface{}
 
 var FirstBoot = []Interface{
 	&BundlePostInstall{},


internal/agent/hooks/hook.go
```diff
@@ -17,9 +17,7 @@ var AfterInstall = []Interface{
 	&Lifecycle{}, // Handles poweroff/reboot by config options
 }
 
-var AfterReset = []Interface{
```
mudler, line 20: so I assume we don't re-encrypt anymore? Isn't elemental-cli deleting and recreating them?

Itxaka: I have no idea how this works but cli is finding it, and formatting it, but only the mapped device (so /dev/mapper/sda5 for example) which means that the encryption is still in place no? Its only formatting the decrypted device, while the encrypted one hasnt changed.

Again, no idea, but its like erasing the contents of it only.

mudler: Yup that sounds good then 👍

mudler (approved):

PR #1021, opened 2023-03-02
:robot: Update workflow permissions

mudler:
Otherwise we don't have permission to upload security reports

See: https://github.com/github/codeql/issues/8843


mudler:
ok this has effect only on master.. merging and let's see if works :)

PR #1022, opened 2023-03-02
:robot: Run framework build on self-hosted

mudler:
Those jobs are small (typically runs for 4/5mins) and might be useful to re-use the cache in such case which makes it faster. This is a test - It is also better probably to free some slots, as no other jobs depends on the fw images.

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -139,7 +139,7 @@ jobs:
   build-framework:
     needs:
       - get-matrix
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     permissions:
       id-token: write
     strategy:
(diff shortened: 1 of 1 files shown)


mauromorales (dismissed):
lgtm

oz123 (dismissed):

jimmykarily (approved):

jimmykarily:
'setup-docker' is doing something wrong:

```
    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor | sudo cat >/usr/share/keyrings/docker-archive-keyring.gpg
```
the last command `sudo cat` is not correct. This is not the right way to redirect output with sudo persmissions.

Similarly, this fails in our runners:

```
runner@kairos-runners-f6cw7-cmzng:/$ sudo echo "test" > /usr/share/keyrings/test
bash: /usr/share/keyrings/test: Permission denied
```

but this works:

```
 sudo /bin/sh -c 'echo "test" > /usr/share/keyrings/test'
```

jimmykarily:
This works:

```
runner@kairos-runners-f6cw7-cmzng:/$ echo "test" | sudo tee /usr/share/keyrings/test
test
```

but this doesn't:

```
runner@kairos-runners-f6cw7-cmzng:/$ echo "test" | sudo cat >/usr/share/keyrings/test
bash: /usr/share/keyrings/test: Permission denied
```

other than the fact that `tee` also prints the text, it seems to be the simplest fix. We could try to upstream it. Unless someone has a better idea, I don't speak good bash.

mudler:
all good now :)

PR #1027, opened 2023-03-03
:seedling: Make sure to remove generated keys during setup

mudler:
**What this PR does / why we need it**: package managers during setup may generate keys, in this way we cleanup and make sure they don't end up to be in the final image


mauromorales (dismissed):

Itxaka (dismissed):

mauromorales (approved):
lgtm

PR #1030, opened 2023-03-03
:bug: (WIP - Do not merge) Ensure systemd-networkd starts properly

kreeuwijk:
This change ensures any network config files updated during `initramfs` are read into systemd-networkd and the stack is restarted. This fixes issues with the DHCP IP address getting dropped under certain circumstances in Equinix Metal. Also fixes dynamically defined network bonding not coming up properly until the daemon-reload and restart commands are issued.

### DIFF
--- a/overlay/files/system/oem/05_network.yaml
+++ b/overlay/files/system/oem/05_network.yaml
@@ -13,6 +13,7 @@ stages:
       commands:
         - rm /etc/resolv.conf
         - ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
+# This change ensures any network config files updated during initramfs are read into systemd-networkd and the stack is restarted. This fixes issues with the DHCP IP address getting dropped under certain circumstances in Equinix Metal. Also fixes dynamically defined network bonding not coming up properly until the daemon-reload and restart commands are issued. See also: https://github.com/kairos-io/kairos/pull/1030
   boot.after:
     - name: "Restart network"
       commands:
--- a/overlay/files/system/oem/05_network.yaml
+++ b/overlay/files/system/oem/05_network.yaml
@@ -1,5 +1,10 @@
 name: "Default network configuration"
 stages:
+  boot:
+    - name: "Restart network"
+      commands:
+       	- systemctl daemon-reload
+       	- systemctl restart systemd-networkd
   initramfs:
     - name: "Setup network"
       if: '[ -e "/sbin/systemctl" ] || [ -e "/usr/bin/systemctl" ] || [ -e "/usr/sbin/systemctl" ] || [ -e "/usr/bin/systemctl" ]'
--- a/overlay/files/system/oem/05_network.yaml
+++ b/overlay/files/system/oem/05_network.yaml
@@ -14,7 +14,7 @@ stages:
         - rm /etc/resolv.conf
         - ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
 # This change ensures any network config files updated during initramfs are read into systemd-networkd and the stack is restarted. This fixes issues with the DHCP IP address getting dropped under certain circumstances in Equinix Metal. Also fixes dynamically defined network bonding not coming up properly until the daemon-reload and restart commands are issued. See also: https://github.com/kairos-io/kairos/pull/1030
-  boot.after:
+  initramfs.after:
     - name: "Restart network"
       commands:
         - systemctl daemon-reload
--- a/overlay/files/system/oem/05_network.yaml
+++ b/overlay/files/system/oem/05_network.yaml
@@ -15,12 +15,13 @@ stages:
         - ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
 # This change ensures any network config files updated during initramfs are read into systemd-networkd and the stack is restarted. This fixes issues with the DHCP IP address getting dropped under certain circumstances in Equinix Metal. Also fixes dynamically defined network bonding not coming up properly until the daemon-reload and restart commands are issued. See also: https://github.com/kairos-io/kairos/pull/1030
   initramfs.after:
-    - name: "Restart network"
+    - name: "Restart network stack when booting in live mode"
+      if: '[ -f /run/cos/live_mode ]'
       commands:
         - systemctl daemon-reload
         - systemctl restart systemd-networkd
         - |
-          if [ -f /run/cos/live_mode ]; then ip link | awk -F: '$0 !~ "NO-CARRIER|lo|vir|wl|usb|^[^0-9]"{print $2;getline}'| while read line; do /sbin/dhclient -4 $line; done; fi
+          ip link | awk -F: '$0 !~ "NO-CARRIER|lo|vir|wl|usb|^[^0-9]"{print $2;getline}'| while read line; do /sbin/dhclient -4 $line; done
 #     dns:
 #      path: /etc/resolv.conf
 #      nameservers:
--- a/overlay/files/system/oem/05_network.yaml
+++ b/overlay/files/system/oem/05_network.yaml
@@ -20,6 +20,9 @@ stages:
       commands:
         - systemctl daemon-reload
         - systemctl restart systemd-networkd
+  before-install:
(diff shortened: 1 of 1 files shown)


overlay/files/system/oem/05_network.yaml
```diff
@@ -1,5 +1,10 @@
 name: "Default network configuration"
 stages:
+  boot:
```
mudler, line 3:
```suggestion
  initramfs.after:
```

overlay/files/system/oem/05_network.yaml
```diff
@@ -1,5 +1,10 @@
...
+    - name: "Restart network"
+      commands:
+       	- systemctl daemon-reload
+       	- systemctl restart systemd-networkd
```
mudler, line 7:
```suggestion
       	- systemctl restart systemd-networkd
```

overlay/files/system/oem/05_network.yaml
```diff
@@ -13,6 +13,13 @@ stages:
       commands:
         - rm /etc/resolv.conf
         - ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
+  initramfs.after:
+    - name: "Restart network"
+      commands:
+        - systemctl daemon-reload
+        - systemctl restart systemd-networkd
+        - |
+          ip link | awk -F: '$0 !~ "NO-CARRIER|lo|vir|wl|usb|^[^0-9]"{print $2;getline}'| while read line; do /sbin/dhclient -4 $line; done
```
mudler, line 22: this approach is really invasive and might clash with static IPs setups. I wonder, isn't equivalent doing the same when dumping your network configuration in the first place?

kreeuwijk: The `dhclient -4` command has no effect on statically configured interfaces. It will only trigger a DHCP refresh on interfaces that are supposed to get DHCP addresses. The goal of the `awk` command is to find the interfaces that wired up and can actually get their DHCP address, skipping interfaces that not connected and could cause DHCP delays for other interfaces. We need to have the DHCP address in place before stylus-agent starts flashing the local HDD, hence it is extremely important that we do our best to ensure the addresses are in place.

mudler: indeed, I'm just asking if this can't be provided with the cloud config supplied during installation? what's the catch I'm missing here?

I say this also because I'm still not really convinced by this solution. relies on too many moving pieces that might not _always_ do what you expect here: 
- on `awk` - which version do we have ? do we have `awk`? we never explictly relied on it until now
- a certain implementation on `dhclient` present (did you checked that the same flags works for e.g. in the dhclient busybox? or in the opensuse/ubuntu version of dhclient?
- assumes that there is a configuration going on between the stages, which is the specific case, and not a general one

kreeuwijk: The problem is that the NICs in Equinix Metal are not predictable. Some servers have only 2 onboard NICs, others have both the onboard NICs and an additional NIC card. When the additional NIC card is present, that one is wired up and the onboard NICs are not. Hence there is no way to know upfront which `eth/eno/enp` interface is the one that needs to be used for DHCP in that server. Hence the iPXE bootfile uses `ip=dhcp` (without specifying a specific NIC) and the cloud-config uses a script to dynamically query the correct NIC config from the Equinux Metal metadata API and build the systemd-networkd networking files for the next boot (registration phase).

It's just the installation phase that we need to get through. What could be done is to gate the `ip link | awk | dhclient` command behind an `if: "[ -f /run/cos/live_mode ]"` statement, so that it only fires during the installation phase, which is the only time it's really needed.

I'm using OpenSUSE since that's in Stylus, but I imagine that since we use systemd-networkd everywhere for Kairos, that `dhclient` will thus be present everywhere too?

kreeuwijk: I tried using just `/sbin/dhclient -4` to avoid the `awk` statement but this suffered from the same DHCP delays and caused kairos to have already installed the image onto the HDD before the system had an IP address. Hence it couldn't retrieve the metadata from the `config_url` parameter before flashing and the necessary customization from the cloud-config does not happen. Since that customization is responsible for building the network bonding info, upon the next boot, the system has no working network stack.

mudler: @kairos-io/maintainers  thoughts?

kreeuwijk: >Is this happening only on the installation? That sounds less invasive, let's do this.

Yes the problem only occurs on the installation. I will test this variant and update the PR if it works:
```
  initramfs.after:
    - name: "Restart network"
      commands:
        - systemctl daemon-reload
        - systemctl restart systemd-networkd
    - name: "Force DHCP refresh before installation"
      if: "[ -f /run/cos/live_mode ]"
      commands:
        - |
          ip link | awk -F: '$0 !~ "NO-CARRIER|lo|vir|wl|usb|^[^0-9]"{print $2;getline}'| while read line; do /sbin/dhclient -4 $line; done
```

This assumes Kairos will execute the "Force DHCP refresh before installation" after the "Restart network" step, if this is not guaranteed, please let me know how I can ensure this.

kreeuwijk: >are you sure this is still the case with 1.6.0? this looks a timing issue

Yes this is on 1.6.0, no change vs 1.5.0 as this issue occurs when the Login Prompts targets is reached, so all the way at the end of the first startup cycle, just before kairos starts installing the OS onto the disk.

overlay/files/system/oem/05_network.yaml
```diff
@@ -13,6 +13,13 @@ stages:
       commands:
         - rm /etc/resolv.conf
         - ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
```
mudler, line 15:
```suggestion
        - ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
# This change ensures any network config files updated during initramfs are read into systemd-networkd and the stack is restarted. This fixes issues with the DHCP IP address getting dropped under certain circumstances in Equinix Metal. Also fixes dynamically defined network bonding not coming up properly until the daemon-reload and restart commands are issued. See also: https://github.com/kairos-io/kairos/pull/1030
```

overlay/files/system/oem/05_network.yaml
kreeuwijk: Execution order could not be guaranteed, so I moved the command back into the "Restart network" step but gated it with a bash `if` statement. This works correctly.
Also changed `initramfs.after` to `boot.after`, since I noticed that `initramfs.after` was too early for subsequent boots (i.e. registration mode).

overlay/files/system/oem/05_network.yaml
```diff
@@ -13,6 +13,14 @@ stages:
...
+# This change ensures any network config files updated during initramfs are read into systemd-networkd and the stack is restarted. This fixes issues with the DHCP IP address getting dropped under certain circumstances in Equinix Metal. Also fixes dynamically defined network bonding not coming up properly until the daemon-reload and restart commands are issued. See also: https://github.com/kairos-io/kairos/pull/1030
+  boot.after:
+    - name: "Restart network"
+      commands:
+        - systemctl daemon-reload
+        - systemctl restart systemd-networkd
+        - |
```
mudler, line 22: I seem to fail to suggest the change, but this should be : 

```yaml
 boot.after:
    - name: "Restart network"
      if: ...
```

kreeuwijk: It seems the `boot` stage doesn't get executed during a live boot, at least I didn't get the proper result when using `boot.after` for some servers. Switching back to `initramfs.after` turns out to be more reliable for the live mode boot. I'll gate the entire step then to only be executed during live mode boot as you suggested, and use a seperate network-stack restart in the boot stage as part of the cloud-init configuration.

kreeuwijk:
Turns out that the `before-install` stage is too late; Kairos already failed to retrieve the metadata from the config_url at this point. Retesting with the DHCP refresh in the `fs.after` stage now.
Also looks like the network restart commands are not necessary, will remove those from the PR when validated.

PR #1034, opened 2023-03-03
:robot: Fixup test targets

mudler:
We were listing on a dir that now doesn't exist necessarly anymore, and when picking up the iso from latest release, we are selectively picking one flavor for the upgrade test.


mudler:
All green!

PR #1057, opened 2023-03-07
:robot: Add grype scan reports

### LINKED ISSUE #1056: :seedling: Use `grype` for scanning SBOM and images
We use `trivy`, that so far collects reports only for `debian` and `ubuntu-*` flavors

We can also use [grype](https://github.com/anchore/grype), and upload a SARIF file to Github as we do already with trivy. Moreover can be also used with #1055 and also analyze SBOM output. We can also think to release this file separately too later on

mudler:

Fixes #1056

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -38,6 +38,7 @@ all:
   BUILD +docker
   BUILD +image-sbom
   BUILD +trivy-scan
+  BUILD +grype-scan
   BUILD +iso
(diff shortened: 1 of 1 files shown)


mudler:
The fact that reports more is already good :) I see now reports for almost all the flavors, coolio!

jimmykarily (approved):

PR #1062, opened 2023-03-08
:bug: Improve network stage

Itxaka:

 - First disable NetworkManager and wicked
 - Then enable systemd services for network and resolv
 - Then remove and link /etc/resolv.conf

This fixes a case in which systemd-networkd could be enabled before wicked is disabled and because network is linked to wicked it would fail to enable systemd-networkd properly during initramfs.

Also adds a check before removing the resolv.conf, to skip it if it doesnt exist.

Found by immucore running the initramfs as it spits out the errors into the log and we can see them more easily

Fixes #

### DIFF
--- a/overlay/files/system/oem/05_network.yaml
+++ b/overlay/files/system/oem/05_network.yaml
@@ -1,15 +1,20 @@
 name: "Default network configuration"
 stages:
   initramfs:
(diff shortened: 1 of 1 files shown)


mudler (approved):
Looks good here, nice cleanup!

PR #1063, opened 2023-03-08
🔧 try and fix path exclusion in CI

oz123:


oz123:
@Itxaka I reverted exclusion for `.github` paths too.

Itxaka (dismissed):
looks good to me!

Itxaka:
will this pick changes tot he files in the root dir? Like Earthfile and such?

Itxaka:
I dont think so, maybe we need to also add `/*` ¬_¬

.github/workflows/image.yaml
```diff
@@ -8,7 +8,13 @@ on:
...
     paths:
       - '!docs/**'
-      - '!.github/**'
+      - 'cmd/**'
+      - 'tests/**'
+      - 'pkg/**'
+      - 'internal/**'
+      - 'images/**'
+      - 'overlay/**'
+      - 'skd/**'
```
mudler, line 17:
```suggestion
      - 'sdk/**'
```

.github/workflows/lint.yml
```diff
@@ -8,7 +8,13 @@ on:
...
     paths:
       - '!docs/**'
-      - '!.github/**'
+      - 'cmd/**'
+      - 'tests/**'
+      - 'pkg/**'
+      - 'internal/**'
+      - 'images/**'
+      - 'overlay/**'
+      - 'skd/**'
```
mudler, line 17:
```suggestion
      - 'sdk/**'
```

.github/workflows/unit-tests.yml
```diff
@@ -8,7 +8,13 @@ on:
...
     paths:
       - '!docs/**'
-      - '!.github/**'
+      - 'cmd/**'
+      - 'tests/**'
+      - 'pkg/**'
+      - 'internal/**'
+      - 'images/**'
+      - 'overlay/**'
+      - 'skd/**'
```
mudler, line 17:
```suggestion
      - 'sdk/**'
```

.github/workflows/unit-tests.yml
```diff
@@ -8,7 +8,13 @@ on:
   pull_request:
     paths:
       - '!docs/**'
```
mudler, line 10: I guess we should include .github too?

.github/workflows/lint.yml
```diff
@@ -8,7 +8,13 @@ on:
   pull_request:
     paths:
       - '!docs/**'
-      - '!.github/**'
+      - 'cmd/**'
+      - 'tests/**'
+      - 'pkg/**'
+      - 'internal/**'
```
mudler, line 14: ditto here

.github/workflows/unit-tests.yml
```diff
@@ -8,7 +8,13 @@ on:
...
-      - '!.github/**'
+      - 'cmd/**'
+      - 'tests/**'
+      - 'pkg/**'
+      - 'internal/**'
```
mudler, line 14: would then skip if we touch the Earthfile?

.github/workflows/unit-tests.yml
oz123: @Itxaka had concerns, so I removed it.

mudler: but if we don't include it among the `paths`, any modification to it won't trigger a workflow build, no?

mudler (commented):
I think this approach isn't very flexible. Looking now at the PR here and how it grows makes it really error prone to maintain. Do we have also to include each file in the top root? I think if we can't make it _exclude_-only, that's going to bite us over and over again.

.github/workflows/unit-tests.yml
mudler: I think that's the minimum requirement - everything we touch _except_ docs should trigger the CI. there are just few cases where we don't want them to trigger

oz123: That's what I tried to achieve with `!docs/**` according to GH documentation, it should work.
But should is a very rare fish :face_with_spiral_eyes:

mudler (approved):
Looks good! let's give it a try

PR #1064, opened 2023-03-08
:robot: Add wildcard also on master workflow

mudler:
Seems a leftover


mudler:
ok this should work, just backporting the patch to the master branch

PR #1068, opened 2023-03-08
updated mock configuration to be compliant with the configuration schema

paynejacob:

tests were failing because the example configuration we were using was not valid

Fixes #


mudler:
Fixed in https://github.com/kairos-io/kairos/pull/1091

PR #1071, opened 2023-03-09
:robot: Consinstently install earthly from our packages repository

mudler:
**What this PR does / why we need it**: CI keep failing to download earthly due to pull rate limits. This makes everything consistent and use the earthly version in our packages repository.

I've created a github composite action here: https://github.com/Luet-lab/luet-install-action could be also a good starting point if we want also to improve our setup of self-hosted workers.


mudler:
if this works, I'll tag a v1 of the action and update the PR

Itxaka:
arent the earthly packages in our repo very old? Like we are using 0.7.1 on other jobs but our packages have 0.6.19

Itxaka (approved):

mudler:
works, unit-tests are failing but should be fixed by #1068

PR #1072, opened 2023-03-09
:robot: Be consistent and install earthly in the CI

mudler:
**What this PR does / why we need it**: I think this generally speeds up thing. We were using the convenience script, but that runs earthly inside its own docker daemon inside a container, slowing down things unnecessarily

Now that we have a composite action, just use it consistently.


mudler:
This saves at least ~7m in builds:

https://github.com/kairos-io/kairos/actions/runs/4372896125/jobs/7650391317
https://github.com/kairos-io/kairos/actions/runs/4373125518/jobs/7650954600

Itxaka:
ni-ni-ni-niiiiice

Also if you add the env var `FORCE_COLOR: 1` we will get the nicer colors that we are used to :=)

Itxaka (dismissed):

sdwilsh:
This runs the risk of using a different version of Earthly, however, right?

mudler:
> This runs the risk of using a different version of Earthly, however, right?

not really, we control the bumps from here: https://github.com/kairos-io/packages/blob/main/packages/utils/earthly/definition.yaml

See: https://github.com/kairos-io/packages/pull/138 for instance

Issue #1087, opened 2023-03-10
:robot: Sporadic failures with trivy


mudler:
**Kairos version:**
master

**CPU architecture, OS, and Version:**

**Describe the bug**
Seems from last release of trivy that happened yesterday, occasionally trivy fails and complains on files of the `/`

**To Reproduce**
It doesn't happen always, but it is easy to reproduce

**Expected behavior**
To generate a report

**Logs**
```
         +grype-scan | [0015]  WARN some package(s) are missing CPEs. This may result in missing vulnerabilities. You may autogenerate these using: --add-cpes-if-none
         +grype-scan | [0015]  WARN unable to determine linux distribution: unable to determine distro type
         +grype-scan | [0018]  WARN unable to determine linux distribution: unable to determine distro type
         +grype-scan | Report written to "report.json"
         +trivy-scan | 2023-03-10T07:40:58.483Z	FATAL	filesystem scan error: scan error: scan failed: failed analysis: walk filesystem: walk error: unknown error with /: unknown error with //tmp: unknown error with //tmp/rpm2110412680: no such file or directory
         +trivy-scan | ERROR Earthfile line 538:4
         +trivy-scan |       The command
         +trivy-scan |           RUN /trivy filesystem --format sarif -o report.sarif --no-progress /
         +trivy-scan |       did not complete successfully. Exit code 1
```

**Additional context**
This was never observed before.

https://github.com/kairos-io/kairos/actions/runs/4382296439/jobs/7671199183
https://github.com/kairos-io/kairos/actions/runs/4382296439/jobs/7671200474

mudler:
This started to happen since yesterday - there was a new release of trivy yesterday indeed. Coincidence? https://github.com/aquasecurity/trivy/releases/tag/v0.38.2

mudler:
other occurrences: 
https://github.com/kairos-io/kairos/actions/runs/4382296439/jobs/7671200474
https://github.com/kairos-io/kairos/actions/runs/4382296439/jobs/7671199556

mudler:
Opened an issue upstream, it seems downgrading made everything green again

Itxaka:
Should probably set `--skip-dirs /tmp` so it stops scanning the temp dir, which makes no sense and Im not sure why its not a built it skipped dirs... I mean they already have a few decent ones, why would they scan the /tmp dir??? https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/secret/builtin-allow-rules.go

[...]

PR #1090, opened 2023-03-10
:robot: Pin trivy version

mudler:
Fixes: #1087


mudler:
if this works possibly there is a regression, will create an issue on trivy if this confirms it

Itxaka (approved):

Itxaka:
@mudler opensuse passed correctly which was the one that kept failing. Need to wait for the job to finish?

PR #1100, opened 2023-03-10
:robot: Various enhancement to security scans

mudler:
**What this PR does / why we need it**: Extracted from #1093

Refactor the docker target. It is now split in base-image and image. the old docker target is equivalent to image. The base-image spits an image with an untouched os-release file so scanners can pick up the original base-image version. I've also removed some dups present in the Earthfile


Earthfile
```diff
@@ -315,39 +316,15 @@ docker:
...
 
-    DO +OSRELEASE --HOME_URL=https://github.com/kairos-io/kairos --BUG_REPORT_URL=https://github.com/kairos-io/kairos/issues --GITHUB_REPO=kairos-io/kairos --VARIANT=${VARIANT} --FLAVOR=${FLAVOR} --OS_ID=${OS_ID} --OS_LABEL=${OS_LABEL} --OS_NAME=${OS_NAME} --OS_REPO=${OS_REPO} --OS_VERSION=${OS_VERSION}
-
     RUN rm -rf /etc/machine-id && touch /etc/machine-id && chmod 444 /etc/machine-id
 
     # Avoid to accidentally push keys generated by package managers
     RUN rm -rf /etc/ssh/ssh_host_*
 
-    # Copy flavor-specific overlay files
-    IF [[ "$FLAVOR" =~ "alpine" ]]
```
mudler, line 335: this is a dup.. we already copy in the framework target the same files. It probably was an oversight when doing the framework image refactoring

mudler:
back to business @Itxaka ! :D

PR #1110, opened 2023-03-12
:robot: Use earhly script for ARM builds

mudler:
This seems to be much more stable as we use loopback mounts, and those could be used by the system already. Running earthly inside its own container fixes this


Itxaka (approved):
nais

PR #1111, opened 2023-03-12
:sparkles: generate arm64 isos properly

Itxaka:

This allows to build a generic arm64 iso with the same features as the x86 iso when running on arm64 platforms.
Useful for testing arm64 in generic arch64 qemu platforms.

Fixes #


Itxaka:
NEeds a valid default value of x86_64 for arch and then an IF to change that to arm64 instead

Earthfile
```diff
@@ -427,11 +427,12 @@ iso:
     ARG ISO_NAME=${OS_ID}
     ARG IMG=docker:$IMAGE
     ARG overlay=overlay/files-iso
+    ARG TARGETARCH
     FROM $OSBUILDER_IMAGE
     WORKDIR /build
     COPY . ./
     COPY --keep-own +image-rootfs/rootfs /build/image
-    RUN /entrypoint.sh --name $ISO_NAME --debug build-iso --squash-no-compression --date=false dir:/build/image --overlay-iso /build/${overlay} --output /build/
+    RUN /entrypoint.sh --name $ISO_NAME --debug build-iso --arch $TARGETARCH --bootloader-in-rootfs --squash-no-compression --date=false dir:/build/image --overlay-iso /build/${overlay} --output /build/
```
mudler, line 435:
```suggestion
    RUN /entrypoint.sh --name $ISO_NAME --debug build-iso --arch $TARGETARCH --bootloader-in-rootfs --squash-no-compression --date=false dir:/build/image --overlay-iso /build/${overlay} --output /build/
```
uh, we don't have `bootloader-in-rootfs`, not a fan of it tbh, it leaves too much room for disambiguity, better accept a folder with the bootloader - also.. osbuilder should have all the bits to create an iso and do not demand anything on the rootfs (why the rootfs would even care?) and we also cover airgap scenarios in one pass

Itxaka: ah shit, sorry this is a leftover of my local tests! Tried to generate the isos in a few different ways as the TARGETARCH wasnt working for me...dur not adding the ARG argument :D

Itxaka: ahhh this is a draft! I was wondering why I did sent this to review...I didnt :D

mauromorales (approved):
lgtm

Itxaka:
ahh crap, this doesnt work. For some reason is picking up the bootx64.efi instead og bootaa64.efi... probably the reson to use the bootloader-in-rootfs flag....now I wonder where its picking the efi stuff from...probably the osbuilder needs a fixup

Itxaka:
This may be fixed now with the packages fix to provide artifacts for arm 64

Itxaka:
nah, not working, way too many things to change, not enough time. Dropping this as there is not much interest in fixing it

Issue #209, opened 2022-10-12
Custom partitioning


mudler:
**Is your feature request related to a problem? Please describe.**
A way to configure partitions via config file

**Describe the solution you'd like**
A simple and detailed way in the config to declare the partitions after install. Ideally should just enlarge the elemental config so it just exposes the underlying configuration

**Describe alternatives you've considered**
Dropping the config file manually, but the solution is more a workaround. It can be achieved by adding an initramfs stage which overwrites the elemental-config

**Additional context**

jimmykarily:
When the config is mounted as a datasource iso (as [suggested here](https://github.com/kairos-io/kairos/issues/391#issuecomment-1456313786)), it partly works:

```
localhost:~ # cat /oem/userdata.yaml 
#cloud-config

users:
- name: "kairos"
  passwd: "kairos"

# Tell elemental to not create partitions
options:
  no-format: "true"

# User has to copy this file inside /oem
# because elemental will try to find stage steps there. Won't repect it here.
# Then the user has to install with
# `kairos-agent manual-install` with this file.
# Netbooting doesn't work for the same reason (elemental ignores the stages).
stages:
  before-install:
  - name: "Create a file"
    commands:
      - |
        touch /tmp/touched-before-install
  boot:
  # TODO: check which device exists
  - if:  '[ -e /dev/vda ] && (kairos-agent state get boot | grep -q unknown)'
    name: "Create partitions"
    commands:
      - |
        touch /tmp/touched-boot
    layout:
      device:
        path: /dev/vda
      add_partitions:
        - fsLabel: COS_STATE
          size: 16240 # At least 16gb
          pLabel: state
```

```
localhost:~ # ls /tmp/touched-*
/tmp/touched-boot
kairos-agent state get boot; echo
unknown
```

- before-install didn't run (maybe I need more config to enable auto-install?)
- the partition was not created:

```
localhost:~ # lsblk
NAME  MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
fd0     2:0    1     4K  0 disk 
loop0   7:0    0 812.9M  1 loop /run/rootfsbase
sr0    11:0    1   364K  0 rom  
sr1    11:1    1  1024M  0 rom  
vda   253:0    0    30G  0 disk 
```

jimmykarily:
I commented out the `options.no-format` section and at least it tries now (but fails):


```
$ journalctl
...
[...]

PR #1117, opened 2023-03-13
:bug: Fix configuration not being merge

### LINKED ISSUE #687: 🐛 Manual install doesnt pick options from config file
**Kairos version:**
master

all

**Describe the bug**
When using a config file and doing manual install `kairos-agent manual-install CONFIG` supposedly there is an `options` key in the config that should be passed down to the installer.
Unfortunately this does not work as it uses a different path, so it loads the file and dumps it into `option["cc"]` and there is nothing to merge those options into the config properly.

An example of how its merged during a normal install is in https://github.com/kairos-io/kairos/blob/master/internal/agent/install.go#L85

**To Reproduce**
create config file with options, call `kairos-agent manual-install CONFIG`

**Expected behavior**

Options are respected and passed down

**Logs**

**Additional context**

oz123:
Fix #687.


oz123:
@jimmykarily I'm having troubles understanding why upgrade with opensuse works, but fails with alpine. Is this random?

Itxaka (commented):
codewise looks good to me, but it seems like we are changing the outcome somehow...I also dont get it :(

oz123:
The tests run perfectly well locally with:
```
earthly +all --IMAGE=quay.io/kairos/core-alpine-opensuse-leap:latest --FLAVOR=alpine-opensuse-leap
MEMORY=9048 \
KVM=true \
USE_QEMU=false \
CPUS=4 \
MACHINE_DISPLAY=true \
KEEP_VM=true \
CREATE_VM=true \
CONTAINER_IMAGE=ttl.sh/kairos-alpine-opensuse-leap-3b5156b9945a9180077d5c7c157265079c72b6e4:8h \
ISO=/home/xxx/Software/spectro-cloud/kairos/build/kairos-core-alpine-opensuse-leap.iso \
FLAVOR=alpine \
ginkgo run -v --label-filter upgrade-with-cli tests/
```

jimmykarily:
This will probably change when we are done with the refactoring described here: https://github.com/kairos-io/kairos/issues/209#issuecomment-1464224461

Up to now, the kairos config tried to be both: a specific config for the kairos-agent and a generic config for every component (kcrypt, elemental-cli etc).

After the refactoring, any consumer that needs the full (generic) config, will use the new collector package and if needed will unmarshal that to a more "specific" struct.

Having said this, there is no need to block this PR since the refactoring is in progress and hasn't touched this file yet.

oz123:
@jimmykarily you'll have to approve and merge since I can't do it.

mudler (approved):

PR #1137, opened 2023-03-14
:seedling: remove duplicate installation of `which` package in opensuse-leap Dockerfile

flesser:

The `which` package was listed twice in the `zypper install` command. One time should be enough.


mudler (approved):
Looks good, thank you!

PR #1140, opened 2023-03-15
:sparkles: Use kairos-sdk module

Itxaka:
Exports the sdk into its own module so it can be released, fixed and worked without the kairos dependency and instead have kairos depend on it

Fixes #


Itxaka:
IMHO the machine module under pkg should also be moved into the sdk. Plus it seems that there are some duplicated functions like BootFrom which checks where it has booted from and the runtime also provides the same.

pkg/config/config.go
```diff
@@ -12,10 +12,10 @@ import (
 	retry "github.com/avast/retry-go"
 	"github.com/imdario/mergo"
 	"github.com/itchyny/gojq"
+	"github.com/kairos-io/kairos-sdk/bundles"
+	"github.com/kairos-io/kairos-sdk/unstructured"
 	schema "github.com/kairos-io/kairos/pkg/config/schemas"
```
mudler, line 17: wondering, shall we make the config part of the sdk? in any case doesn't matter with this PR

mudler (dismissed):
Looking good!

cmd/agent/main.go
```diff
@@ -9,12 +9,12 @@ import (
 	"github.com/kairos-io/kairos/internal/bus"
 	"github.com/kairos-io/kairos/internal/webui"
 
+	bundles "github.com/kairos-io/kairos-sdk/bundles"
```
mudler, line 12: maybe we should have named the repo just "sdk" and make the golang files top level? so imports are cleaner:
```
import (
 "github.com/kairos-io/sdk/..."
)
```

mudler:
> IMHO the machine module under pkg should also be moved into the sdk. Plus it seems that there are some duplicated functions like BootFrom which checks where it has booted from and the runtime also provides the same.

yep, definetly :+1: feel free to take a stab at it

mudler (approved):

pkg/config/config.go
Itxaka: yes, indeed, it would be the perfect part.

cmd/agent/main.go
Itxaka: Thougth about that but feel like it was very...generic. Like if we had down the line a different sdk for something else? kairos-sdk looked like it was very clear what it is for. I dunno...we have time to change this, I have no strong feelings one way or the other

PR #1163, opened 2023-03-17
:arrow_up: Bump repositories

### LINKED ISSUE #1116: :bug: :hedgehog: Revisit logging for immucore
**Kairos version:**
master

**Describe the bug**
Immucore logging is a bit broken currently. Im pretty sure its not picking the debug cmdline due to how Im setting the logger (broken) and some parts will also always log in debug.
Also there may be some debug logs around that are useless now and were part of the testing that should be removed.

**To Reproduce**
run 2.0.0-alpha1 and check the log output of immucore. 

**Expected behavior**
Clean info level logging if nothing is touched. When rd.immucore.debug is set int he cmdline, debug logs to appear.

**Logs**

**Additional context**

Itxaka:
Includes new packages for immucore and grub2 livecd Removes install_test script, just run ginkgo directly

Fixes #1116

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -251,7 +251,7 @@ jobs:
           export PATH=$PATH:$GOPATH/bin
           export CREATE_VM=true
           export FLAVOR=${{ matrix.flavor }}
-          ./.github/run_test.sh "install-test"
+          go run github.com/onsi/ginkgo/v2/ginkgo --label-filter "install-test" --fail-fast -r ./tests/
       - uses: actions/upload-artifact@v3
         if: failure()
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/image.yaml
```diff
@@ -251,7 +251,7 @@ jobs:
           export PATH=$PATH:$GOPATH/bin
           export CREATE_VM=true
           export FLAVOR=${{ matrix.flavor }}
-          ./.github/run_test.sh "install-test"
+          go run github.com/onsi/ginkgo/v2/ginkgo --label-filter "install-test" --fail-fast -r ./tests/
```
Itxaka, line 254: this uses the ginkgo version provided in the go.mod file, so no more downgrades or upgrades and adding deps in the targets.

mudler (approved):
Looking good!

PR #1168, opened 2023-03-17
:book: Make sample systemd networking filename work out of the box

robarnold:
I tried setting up a new Kairos installation with Ubuntu & Debian on bare metal using the documentation, and I realized that `systemd-networkd` was pulling in the default dhcp config (`20-dhcp.network`) instead of the one I had authored (`config.network`). Since systemd appears to read the config files in alphabetical order, `config.network` will match after the DHCP file (`Name=en*`) and not be applied. Leaving around the default dhcp configuration in the default image seems useful for first boot of the installer, so I am hoping to address this by fixing future copy & pastes.


Itxaka (approved):
good catch!

PR #816, opened 2023-02-03
:penguin: Add ubuntu arm rpi flavor

c0ffee:
**What this PR does / why we need it**:
Ubuntu flavors which support raspberrypi

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #645

PR #1170, opened 2023-03-20
:penguin: Add Ubuntu flavors for RPi

### LINKED ISSUE #645: :penguin: Add ARM/RPI support for Ubuntu 20.04 and 22.04 images
Currently kairos rpi images are available for alpine and opensuse. Would like ubuntu images to be added as well.

ubuntu  20.04 and 22.04 rpi images for rpi 3 and 4

mauromorales:
QA:

- [x] rolling release
- [x] 22 lts
- [x] 20 lts 

 Ubuntu flavors which support raspberrypi

Fixes #645
supersedes #816


mauromorales:
Sizes needed to be expanded in order to build the images, but now I can finally do. Pending QA

sdwilsh:
@mauromorales, I take it you can't actually get into the system to see any logs to help there, based on the video, yeah?

mauromorales:
@sdwilsh yeah indeed, I'll try to have a better setup for it next week and give it another shot

images/.hadolint.yaml
```diff
@@ -6,8 +6,12 @@ ignored:
   - DL3006
   # Pin versions in apt get install: https://github.com/hadolint/hadolint/wiki/DL3008
   - DL3008
+  # Avoid additional packages by specifying --no-install-recommends: https://github.com/hadolint/hadolint/wiki/DL3015
+  - DL3015
   # Pin versions in apk add: https://github.com/hadolint/hadolint/wiki/DL3018
   - DL3018
+  # Do not use apt as it is meant to be an end-user tool, use apt-get or apt-cache instead: https://github.com/hadolint/hadolint/wiki/DL3027
+  - DL3027
```
sdwilsh, line 14: I cleaned these up in the past (#821 handled both), and I'd prefer to not globally regress.  If we need to, we can allowlist these for a specific file, IIRC.

mauromorales: Good catch, yeah I'll remove it

mauromorales:
For faster iteration I wanted to try using the generic arm image. It builds successfully, startups and shows the grub menu, but then it gets stuck here quite quickly

image

images/Dockerfile.ubuntu-20-lts-arm-rpi
```diff
@@ -0,0 +1,63 @@
...
+    neovim \
+    open-vm-tools \
+    openssh-server \
+    parted \
+    policykit-1 \
+    rpi-eeprom \
+    rsync \
+    snapd \
+    snmpd \
+    squashfs-tools \
```
Itxaka, line 42: dracut complained that its missing sgdisk, so we may need to install it as agetn will use it for partition stuff.

overlay/files-ubuntu-arm-rpi/system/oem/03-setupcon.yaml
```diff
@@ -0,0 +1,7 @@
+name: "Start installer on tty1"
```
Itxaka, line 1: Maybe we can move this to the generic overlay files and do a check to see if we are on ubuntu? So we dont have to track it separatedly

overlay/files-ubuntu-arm-rpi/system/oem/11_persistency.yaml
```diff
@@ -0,0 +1,37 @@
+name: "Configure persistent dirs bind-mounts"
```
Itxaka, line 1: dont we have this already on the generic overlay files?

Earthfile
```diff
@@ -433,21 +441,32 @@ netboot:
...
   ARG IMAGE_NAME=${FLAVOR}.img
   WORKDIR /build
-  ENV STATE_SIZE="6200"
-  ENV RECOVERY_SIZE="4200"
   ENV SIZE="15200"
-  ENV DEFAULT_ACTIVE_SIZE="2000"
+  IF [[ "$FLAVOR" =~ ^ubuntu* ]]
+    ENV STATE_SIZE="6900"
+    ENV RECOVERY_SIZE="4600"
+    ENV DEFAULT_ACTIVE_SIZE="2300"
```
sdwilsh, line 453: Probably outside the scope of this PR, but it'd be nice if these `_SIZE` variables had the expected unit in their name.  It'd make it much more approachable by new folks.

images/.hadolint.yaml
```diff
@@ -6,6 +6,8 @@ ignored:
   - DL3006
   # Pin versions in apt get install: https://github.com/hadolint/hadolint/wiki/DL3008
   - DL3008
+  # Avoid additional packages by specifying --no-install-recommends: https://github.com/hadolint/hadolint/wiki/DL3015
+  - DL3015
```
sdwilsh, line 10: Can we not add this globally?  I had fixed this in a previous PR.  If you must, you can exempt it from specific Docker files: https://github.com/hadolint/hadolint#inline-ignores

mauromorales:
https://user-images.githubusercontent.com/433958/236797895-9ee08ea4-0945-48d4-a99c-66e83d87f779.MOV

images/.hadolint.yaml
mauromorales: yes definitely, just focusing first on making the system boot before looking at any other comments/issues in here

mauromorales:
Generic image booting properly you can try this out by building it locally from the branch via: `earthly -P +all-arm-generic --FLAVOR=ubuntu-arm-rpi`

mauromorales:
Ok, finally have a better setup, so I was able to SSH into the machine and the first issue I see is this. Will continue investigating:

```
kairos@localhost:~$ sudo systemctl list-units --failed
  UNIT                      LOAD   ACTIVE SUB    DESCRIPTION                          
● polkit.service            loaded failed failed Authorization Manager
● rpi-eeprom-update.service loaded failed failed Check for Raspberry Pi EEPROM updates
```

mauromorales:
The video problems are now fixed. The overlay files were not properly being copied so the vc4 module was enabled which is not supported in RPi. This is the same we do on other RPi flavors so maybe it should really be refactored to be copied whenever the module is rpi64

Earthfile
mauromorales: Created an issue since it also requires changes in osbuilder

https://github.com/kairos-io/kairos/issues/1397

overlay/files-ubuntu-arm-rpi/system/oem/11_persistency.yaml
mauromorales: tried with the generic one and it seems to be working well so removing the rpi one

overlay/files-ubuntu-arm-rpi/system/oem/03-setupcon.yaml
mauromorales: Very good point, I will keep them for this PR so we can merge and opened a discussion on how to address this properly in the future https://github.com/kairos-io/kairos/discussions/1400

Earthfile
```diff
@@ -293,7 +299,7 @@ base-image:
     END
     # END
 
-    # TEST KCRYPT FROM BRANCH
+# TEST KCRYPT FROM BRANCH
```
jimmykarily, line 302: accidental?

Earthfile
```diff
@@ -366,6 +373,7 @@ image:
         ARG OS_VERSION=${KAIROS_VERSION}
     END
     ARG OS_ID
+    # should we add the model to the resulting iso?
```
jimmykarily, line 376: you mean in `/etc/os-release`? Consider moving it here: https://github.com/kairos-io/kairos/issues/1365

images/Dockerfile.ubuntu-20-lts-arm-rpi
```diff
@@ -0,0 +1,63 @@
...
+    e2fsprogs \
+    gawk \
+    grub-efi-arm64-bin \
+    grub2-common \
+    haveged \
+    iproute2 \
+    iptables \
+    jq \
+    linux-firmware-raspi2 \
+    linux-raspi \
```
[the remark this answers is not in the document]
jimmykarily: so should we remove `rpi` from the dockerfile name? Also, shouldn't this be in the rpi64 section above then?

jimmykarily (dismissed):

Earthfile
```diff
@@ -436,21 +444,33 @@ netboot:
...
   FROM $OSBUILDER_IMAGE
   ARG MODEL=rpi64
   ARG IMAGE_NAME=${FLAVOR}.img
   WORKDIR /build
-  ENV STATE_SIZE="6200"
-  ENV RECOVERY_SIZE="4200"
+  # These sizes are in MB
   ENV SIZE="15200"
-  ENV DEFAULT_ACTIVE_SIZE="2000"
+  IF [[ "$FLAVOR" =~ ^ubuntu* ]]
```
Itxaka, line 454: I hate this. This should probably come from some env files where its clear what goes with what flavor and such. Not for this PR

framework-profile.yaml
```diff
@@ -15,6 +15,15 @@ flavors:
   ubuntu:
     - systemd-base
     - dracut-network-legacy
+  ubuntu-arm-rpi:
+    - systemd-base
```
Itxaka, line 19: we should probably need to start grouping this below and just have a single "metapackage" here to not repeat ourselves this much.

images/Dockerfile.ubuntu-20-lts-arm-rpi
```diff
@@ -0,0 +1,79 @@
...
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+FROM base AS rpi64
+# Normally we install also linux-modules-extra-raspi but this is not available for 20.04, not sure what's the impact to
+# remove it.
+RUN apt-get update && apt-get install -y linux-firmware-raspi2 \
+    linux-raspi \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+FROM ${MODEL} AS all
```
Itxaka, line 23: I...dont get this.

So if we call this with the generic model it will run:
 - FROM $BASE_IMAGE as base
 - FROM base AS generic
 - FROM ${MODEL} AS all

but all its still installing the `linux-raspi` kernel no? dont we want the rpi to install those packages only? Wont we end up with a generic image that has both kernels inside and the possibility of linking to it?

overlay/files-ubuntu-arm-rpi/etc/default/console-setup
```diff
@@ -0,0 +1,7 @@
+ACTIVE_CONSOLES="/dev/tty[1-6]"
```
Itxaka, line 1: same file as the ubuntu files, wont copying the ubuntu files if it matches ^ubuntu* would be enough so we dont duplicate?

Itxaka (dismissed):
Looking good, a cleanup can come after about the overlay stuff

Itxaka:
Just thinking out loud, but maybe we should drop to bootargs for rpi and during building add the vc4.conf to /etc/modprobe.d with a `blacklist vc4` line to not even have to worry about the vc4 module and stop adding it to the cmdline?

Earthfile
```diff
@@ -45,15 +45,18 @@ all:
   BUILD +ipxe-iso
 
 all-arm:
-  BUILD --platform=linux/arm64 +image
-  BUILD +image-sbom
-  BUILD +trivy-scan
-  BUILD +grype-scan
-  BUILD +arm-image
+  ARG SECURITY_SCANS=true
```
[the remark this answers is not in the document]
mauromorales: good point, will do

Earthfile
mauromorales: Good point, can it be done in the `/etc/elemental/config.yaml` that we include?

mauromorales: ah probably not cuz this here has precedence when creating the image ... ok will open an issue

images/Dockerfile.ubuntu-20-lts-arm-rpi
mauromorales: Yeah I didn't find it very intuitive, my understanding is that `all` is the requested target (because it's the last one?) and depending on it's base it will only build generic or rpi64. And it properly ignores all the targets it doesn't need to build

see: https://docs.docker.com/build/building/multi-stage/#differences-between-legacy-builder-and-buildkit

overlay/files-ubuntu-arm-rpi/etc/default/console-setup
mauromorales: we just have the one difference with the `/etc/elemental/config.yaml` for ubuntu arm rpi. Here's where I started thinking about also refactoring this but then I thought, I better start the discussion before changing everything LOL

mauromorales:
> Just thinking out loud, but maybe we should drop to bootargs for rpi and during building add the vc4.conf to /etc/modprobe.d with a `blacklist vc4` line to not even have to worry about the vc4 module and stop adding it to the cmdline?

good point, added a new issue https://github.com/kairos-io/kairos/issues/1407

very cool that you can create issues from comments, wasn't aware of this

framework-profile.yaml
mauromorales: do you mean using anchors? or what is a meta package?

Itxaka: no, see how you mark here `systemd-base` and then below you can see that it contains 2 packages: `init-svc/systemd` and `systemd-default/network` ? So Im thinking that we may need to just merge more of the stuff seeing that there is a clear repetition here with systemd-base and dracut-network-legacy*

so something similar to those, like ubuntu-base that includes all those packages, or something shared that includes all of that so its a one line for all. Dont know how, just thinking out loud seeing that this keeps frowing and lines repeating :D

Itxaka (approved):

framework-profile.yaml
mauromorales: Ok, I get your point now. Yeah I agree, I had the same feeling that this was lots of duplication but didn't know how this part integrates to be able to suggest a recommendation. I'll create an issue out of your comment.

PR #1179, opened 2023-03-22
:seedling: Add auth flags to upgrade

Itxaka:
This is stupid. we should be using elemental-cli which is tested. Passing all this parameters like this is terrible and I dont like it.


mudler:
yeap, not really nice for now, but it's a good workaround for the moment :+1:

cmd/agent/main.go
```diff
@@ -79,7 +85,13 @@ See https://kairos.io/docs/upgrade/manual/ for documentation.
 			if c.Args().Len() == 1 {
 				v = c.Args().First()
 			}
-			return agent.Upgrade(v, c.String("image"), c.Bool("force"), c.Bool("debug"), c.Bool("strict-validation"), configScanDir)
+
+			return agent.Upgrade(
```
mudler, line 89: I'm wondering, since we are in control of the pull on the elemental side, if we should just use ENV vars as those are secrets

internal/agent/upgrade.go
```diff
@@ -41,7 +41,10 @@ func ListReleases() []string {
 	return releases
 }
 
-func Upgrade(version, image string, force, debug, strictValidations bool, dirs []string) error {
+func Upgrade(
+	version, image string, force, debug, strictValidations bool, dirs []string,
```
mudler, line 45: another thing is installation too. we should think about how to pass the auth when specifying images with `install.image` :thinking:

mudler (approved):
Looking good!

internal/agent/upgrade.go
Itxaka: yeah, we currently dont have those either

cmd/agent/main.go
Itxaka: good idea, but Im not sure those are loaded via env vars...let me test that

Itxaka: ah no, this is not loaded via env vars becuase its not part of the Config struct, its just flags :/

Itxaka:
Works but requires a bumped elemental-cli wiht support for this flags

Itxaka:
This will fail until https://github.com/kairos-io/kairos/pull/1182 is merged

cmd/agent/main.go
```diff
@@ -37,6 +37,12 @@ var cmds = []*cli.Command{
 				Name:  "image",
 				Usage: "Specify an full image reference, e.g.: quay.io/some/image:tag",
 			},
+			&cli.StringFlag{Name: "auth-username", Usage: "Username to authenticate to registry"},
```
mudler, line 40: I guess we could expose some env var as well? but this looks good anyway :+1:

Itxaka: This should go away if we kind of move the installation stuff over elemental directly and remove it from the agent in the future-ish

PR #1180, opened 2023-03-23
:sparkles: Custom partitioning refactor config

mauromorales:

Fixes #

Needs https://github.com/kairos-io/kcrypt/pull/14


pkg/config/collector/collector.go
```diff
@@ -0,0 +1,372 @@
...
+		if err != nil {
+			break
+		}
+		res += string(dat)
+	}
+	return
+}
+
+// TODO check if doing the right thing ... also checking remote files and cmdline.
+func FindYAMLWithKey(s string, opts ...Option) ([]string, error) {
```
jimmykarily, line 344: I can't find any references to this function. I think it was used before we were able to define the same block of configuration in 2 different files (deep merge). Back then, we just put the configuration in an existing file and this was a way to find which file defined the block. Now we simply create a new file (e.g. for the label mappings) so we don't need to lookup the other files.

It would still be useful if we wanted to make sure the same block is not defined in any other file (e.g. to make sure our file is the "effective" one) but afaik there is only one use-case (kcrypt) and that one always uses the same file.

Besides the above, it's also complex to return a sane return value when the block is defined in a remote `config_url`. What file do we return then? Even worse, when there is a chain of `config_urls` and the block is defined somewhere deeply in the chain.

I think we should remove this function. Please check that I didn't miss any references first.

pkg/config/collector/options.go
```diff
@@ -15,6 +17,16 @@ var NoLogs Option = func(o *Options) error {
 	return nil
 }
 
+// SoftErr prints a warning if err is no nil and NoLogs is not true.
+// It's use to wrap the same handling happening in multiple places.
+//
+// TODO: Switch to a standard logging library (e.g. verbose, silent mode etc).
```
jimmykarily, line 23: Move this TODO on a new issue?

pkg/config/collector/options.go
```diff
@@ -35,17 +47,17 @@ func WithBootCMDLineFile(s string) Option {
 		return nil
 	}
 }
-func Directories(d ...string) Option {
+
+func StrictValidation(v bool) Option {
```
jimmykarily, line 51: (Probably on another PR) We should find a way for the collector options to only hold values that make sense in the context of the collector. Since the collector is not doing any validation (other than yamls being valid etc), this option doesn't belong here.

I think the problem is that the consumers (e.g. kairos, kcrypt etc), hijack the collector options to carry their own options like this one. Let's try to fix this on the next iteration.

jimmykarily (dismissed):

pkg/config/config_test.go
```diff
@@ -43,208 +41,4 @@ var _ = Describe("Config", func() {
 			os.RemoveAll(d)
 		}
 	})
```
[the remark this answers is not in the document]
mauromorales: Not on this PR, but I do think there should be tests here which are specific to the Kairos config

pkg/config/collector/options.go
mauromorales: Yes agree, will open a generic issue and put this one as an example

mauromorales: https://github.com/kairos-io/kairos/issues/1187

pkg/config/config_test.go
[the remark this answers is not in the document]
mauromorales: https://github.com/kairos-io/kairos/issues/1189

pkg/config/config.go
```diff
@@ -97,123 +90,43 @@ func (b Bundles) Options() (res [][]bundles.BundleOption) {
...
-		}
-	}
+	genericConfig, err := collector.Scan(o)
+	if err != nil {
+		return result, err
 
-	if c.HasConfigURL() {
-		err = c.fetchRemoteConfig()
-		if !o.NoLogs && err != nil {
-			fmt.Printf("WARNING: Couldn't fetch config_url: %s\n", err.Error())
```
jimmykarily, line 196: This is no longer a soft error (we used to ignore non-accessible remote configurations). This is why we have a failed test in CI.

I wonder if it's good practice to not fail when configuration is missing. Let's say, a user sets up something really important security-wise, in a remote configuration. It's possible that by ignoring the error, we end up with a less secure system (or even one that doesn't even work). Ignored errors are rarely a good thing and result in broken systems that are hard to debug.

My opinion is we should keep it as it is in this PR and change the test. @mudler thoughts?

pkg/config/collector/collector.go
mauromorales: removed it

jimmykarily (approved):

PR #1194, opened 2023-03-26
:bug: Fix validator on long strings

### LINKED ISSUE #994: :bug: WebUI validation breaks with a network_token
image

Using Kairos 1.6.0

Doing the validation via the CLI passes correctly. The experience should be the same in the Web UI

Itxaka:
Validator was mistakenly identifying a long yaml as a file and trying to open it, which failed with an error of filename too long.

This was not catched in order to identify that the source is not a file but a yaml, so it was directly returning the error.

This patch adds that error to the list in order ot identify the source to validate as yaml. Also adds a couple of tests for this functionality.

Fixes #994

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -66,7 +66,7 @@ test:
     WORKDIR /build
     COPY +luet/luet /usr/bin/luet
     COPY . .
-    RUN go run github.com/onsi/ginkgo/v2/ginkgo --fail-fast --slow-spec-threshold 30s --covermode=atomic --coverprofile=coverage.out -p -r ./pkg ./internal ./cmd ./sdk
+    RUN go run github.com/onsi/ginkgo/v2/ginkgo --fail-fast --covermode=atomic --coverprofile=coverage.out -p -r ./pkg ./internal ./cmd ./sdk
     SAVE ARTIFACT coverage.out AS LOCAL coverage.out
 
 OSRELEASE:
--- a/Earthfile
+++ b/Earthfile
@@ -341,25 +341,20 @@ base-image:
     IF [ "$FLAVOR" = "debian" ]
 	    RUN rm -rf /boot/initrd.img-*
     END
-    # Regenerate initrd if necessary
-    IF [ "$FLAVOR" = "opensuse-leap" ] || [ "$FLAVOR" = "opensuse-leap-arm-rpi" ] || [ "$FLAVOR" = "opensuse-tumbleweed-arm-rpi" ] || [ "$FLAVOR" = "opensuse-tumbleweed" ]
-     RUN mkinitrd
-    ELSE IF [ "$FLAVOR" = "fedora" ] || [ "$FLAVOR" = "rockylinux" ]
-     RUN kernel=$(ls /boot/vmlinuz-* | head -n1) && \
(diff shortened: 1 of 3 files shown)


internal/agent/validate.go
```diff
@@ -31,16 +31,17 @@ func Validate(source string) error {
...
 		}
-		body, err := ioutil.ReadAll(resp.Body)
+		body, err := io.ReadAll(resp.Body)
 		if err != nil {
 			return err
 		}
 		//Convert the body to type string
 		yaml = string(body)
 	} else {
+		// Maybe we should just try to read the string for the normal headers? That would identify a full yaml vs a file
```
jimmykarily, line 41: This function is called `Validate` and most of the code is trying to figure out the type of source we are dealing with. The only validation we do is on line 63 below where we call `IsValid`.

Also, we rely on the error message to decide whether the string is a file or the source itself and I worry that different OSes may return different messages. Since we only run this on Linux, we are probably safe but we could probably use `os.Stat` to see if the file exists as a safer way.

Anyway, none of these problems was introduced in this PR so at least we don't make things worse than before.

Earthfile
```diff
@@ -66,7 +66,7 @@ test:
     WORKDIR /build
     COPY +luet/luet /usr/bin/luet
     COPY . .
-    RUN go run github.com/onsi/ginkgo/v2/ginkgo --fail-fast --slow-spec-threshold 30s --covermode=atomic --coverprofile=coverage.out -p -r ./pkg ./internal ./cmd ./sdk
```
jimmykarily, line 69: Sticking to the default value? What would that be? Any reason we don't want to be explicit here?

internal/agent/validate_test.go
```diff
@@ -24,6 +24,21 @@ var _ = Describe("Validate", func() {
...
+				yaml = `#cloud-config
+users:
+  - name: kairos
+    passwd: kairos
+vpn:
+  network_token: "dssdnfjkldashfkjhasdkhfkasjdhfkjhasdjkfhaksjdhfkjashjdkfhioreqwhfuihqweruifhuewrbfhuewrfuyequfhuiehuifheqrihfuiqrehfuirqheiufhreqiuhfuiqheiufhqeuihfuiqrehfiuhqreuifrhiuqehfiuhqeirhfiuewhrfhqwehfriuewhfuihewiuhfruewhrifhwiuehrfiuhweiurfhwueihrfuiwehufhweuihrfuiwerhfuihewruifhewuihfiouwehrfiouhwei"
+`
+			})
+			It("validates", func() {
+				Expect(Validate(yaml)).ToNot(HaveOccurred())
```
jimmykarily, line 38: Maybe we should also check that it was treated as the actual sources? Just in case a different path was followed that also didn't produce an error (but did the wrong thing).

jimmykarily (approved):

Earthfile
Itxaka: The flag has been deprecated since ginkgo 2.5, so it's not doing anything anymore and won't be valid soon

mauromorales (dismissed):
ah good catch, this fix might already fix https://github.com/kairos-io/kairos/issues/994 :raised_hands:

mauromorales (approved):

Issue #1200, opened 2023-03-28
Fix CI and release 2.0.0-alpha3


jimmykarily:

Itxaka:
one of the problems is oepnsuse-leap not finding mkinitrd (its now deprecated in favour of dracut)

Im trying to fix it as part of https://github.com/kairos-io/kairos/pull/1194

Itxaka:
Oh, I'm dumb. I added "partially fixes" to the pr and that auto closed this, good catch @mauromorales

Itxaka:
all fixes are done:
 - https://github.com/kairos-io/kairos/pull/1205
 - https://github.com/kairos-io/kairos/pull/1203
 - https://github.com/kairos-io/kairos/pull/1194

Pushing alpha3!

Itxaka:
alpha3 cut

mudler:
After we have it green we need to sync the provider repo too

mudler:
well... 2.0 is targeted for end of month, so this is more like a RC than at alpha as for now :)

mudler:
Let's close this once we have the alpha in and have a new one for the RC, I didn't noticed the release I'm between.

Itxaka:
small hiccup with the arm img pushing as the creds are probably not valid for those repos so it could not push the images:https://github.com/kairos-io/kairos/actions/runs/4551323808

Itxaka:
All stuff done, release...uh...released. Closing!

PR #1203, opened 2023-03-28
:robot: Copy sarif files from proper folder

Itxaka:
Fixes the sarif dir under release workflow

Partially fixes https://github.com/kairos-io/kairos/issues/1200

### DIFF
--- a/.github/workflows/release-arm.yaml
+++ b/.github/workflows/release-arm.yaml
@@ -102,7 +102,7 @@ jobs:
           MODEL: ${{ matrix.model }}
         run: |
           mkdir sarif
-          sudo mv release/*.sarif sarif/
+          sudo mv build/*.sarif sarif/
       - name: Upload Trivy scan results to GitHub Security tab
         uses: github/codeql-action/upload-sarif@v2
         with:


mauromorales (approved):
LGTM

.github/workflows/release-arm.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
           MODEL: ${{ matrix.model }}
         run: |
           mkdir sarif
-          sudo mv release/*.sarif sarif/
+          sudo mv build/*.sarif sarif/
```
jimmykarily, line 105: doesn't `release.yaml` also need the same fix?

Itxaka: weirdly enough no, it seems to work in that job: https://github.com/kairos-io/kairos/actions/runs/4532595843/jobs/8008458849

Itxaka: oh I see, the release job moves the files from /build to /release....

Itxaka: not sure why on arm we dont directly do that or why we move the files into release/ and not work with the build directory directly. Probably due to permission stuff?

PR #1207, opened 2023-03-29
:robot: Fix schema job to get tags and include missing schemas from alpha releases

### LINKED ISSUE #1131: :bug: Update schema failed to pick up alpha releases
A PR was opened (https://github.com/kairos-io/kairos/pull/1106) but it didn't pick the right version [v2.0.0-alpha1](https://github.com/kairos-io/kairos/releases/tag/v2.0.0-alpha1) related to the original commit https://github.com/kairos-io/kairos/commit/ef2fe0345426654c33c79271b4b9db2919396b1c

also add this release schema manually

mauromorales:
It fixes the schema job and includes the two currently missing schemas for the alpha releases.

Fixes #1131

### DIFF
--- a/.github/workflows/schema.yaml
+++ b/.github/workflows/schema.yaml
@@ -15,6 +15,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: master
+          fetch-depth: 0
       - name: setup-docker
         uses: docker-practice/actions-setup-docker@master
       - name: Install Go
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ node_modules/
 
 # editor and IDE paraphernalia
 .idea
+.envrc
+.go-version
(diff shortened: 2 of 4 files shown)


Itxaka:
Isnt there a better way to do this? Cant we have like a base one and then the other superseed, overwrite, substitute the base one with whatever is added in that version? Or that is not possible? 

It would be nice to just a base one and then the versioned ones were kind of like git diff, marking what was added or removed from the base to get to that version. Or even a system like in django databases where you have the base models and there migrations that transform that base to whatever version you are in, so you can add or remove tables and so on and on each version they add more migrations.

That would make it easier to maintain I think? Maybe not.

mauromorales:
@Itxaka indeed I was thinking maybe link to the previous one if there are no changes ... will create a ticket to address

mauromorales:
Added https://github.com/kairos-io/kairos/issues/1209

Issue #1199, opened 2023-03-28
upgrade golang version for build time


mudler:
Let's bump to golang 1.20.2

Components:

- [x] https://github.com/kairos-io/kairos/pull/1213
- [x] https://github.com/kairos-io/provider-kairos/pull/299
- [x] https://github.com/kairos-io/kcrypt-challenger/pull/13
- [x]  https://github.com/kairos-io/kcrypt/pull/15
- [x] elemental-cli (needs bump from packages)
- [x] https://github.com/kairos-io/immucore/pull/107
- [x] https://github.com/kairos-io/packages/pull/162

 ( and maybe others?)

Itxaka:
- https://github.com/kairos-io/immucore/pull/107
 - https://github.com/kairos-io/packages/pull/162

mauromorales:
@Itxaka I think https://github.com/kairos-io/elemental-cli is already on 1.20, so we can scratch that out. Can you confirm?

Itxaka:
@mauromorales Indeed! But the package being used on kairos is not, its part of the https://github.com/kairos-io/packages/pull/162 PR, a bit confusing :D

PR #1213, opened 2023-03-29
:arrow_up: Bump go in Earthly and workflows to ^1.20

mauromorales:
Relates to #1199


.github/workflows/release_bin.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.18
+          go-version: 1.20
```
Itxaka, line 19: Missing the carat so this will be always 1.20

.github/workflows/image.yaml
```diff
@@ -244,7 +244,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v4
         with:
-          go-version: ^1.16
```
Itxaka, line 247: I wonder if we really need to do this, I mean currently it just means higher than, so this should be picking the latest version anyway...

mauromorales: @Itxaka good question, I wonder if it's better to let it take the latest one or pin it to the one in the `go.mod` file using `go-version-file`

Earthfile
```diff
@@ -22,9 +22,9 @@ ARG CGO_ENABLED=0
 # renovate: datasource=docker depName=quay.io/kairos/osbuilder-tools versioning=semver-coerced
 ARG OSBUILDER_VERSION=v0.5.3
 ARG OSBUILDER_IMAGE=quay.io/kairos/osbuilder-tools:$OSBUILDER_VERSION
-ARG GOLINT_VERSION=1.47.3
+ARG GOLINT_VERSION=1.52.2
 # renovate: datasource=docker depName=golang
-ARG GO_VERSION=1.18
+ARG GO_VERSION=1.20.2
```
Itxaka, line 27: I guess it should be 1.20 as that will get updated with minor version automatically...unless we want to manually update it on minor versions in case something breaks?

Itxaka (approved):

mudler (approved):
looking good here!

PR #1220, opened 2023-03-30
Change module path according to Go docs

jimmykarily:
https://go.dev/doc/modules/major-version

This way we can bump the kairos dependency on the provider-kairos repo

which otherwise produced the error:

```
~/workspace/kairos/provider-kairos (main)*$ go get -u github.com/kairos-io/kairos@v2.0.0-alpha3
go: github.com/kairos-io/kairos@v2.0.0-alpha3: invalid version: module contains a go.mod file, so module path must match major version ("github.com/kairos-io/kairos/v2")
```


jimmykarily:
> I guess that if it builds, then its correct. Nice!

That was my assumption too :D

mauromorales (approved):
oh wow, and I guess this will be so for every major version upgrade :/ ... saw this when updating urfave/cli and wasn't very impressed

LGTM

Issue #1223, opened 2023-03-30
:bug: Remote encryption broken in 2.0.0


Itxaka:
**Kairos version:**
master

**CPU architecture, OS, and Version:**
all

**Describe the bug**
remote encryption (via challenger server) is currently broken under 2.0.0.
Machine boots but immucore hasnt been able to mount COS_PERSISTENT and anything dependant on it, also no initramfs stage has been run.

**To Reproduce**
Create an encrypted COS_PERSISTENT that unlock via a server i.e. :

```
install:
  encrypted_partitions:
  - COS_PERSISTENT
  grub_options:
    extra_cmdline: "rd.neednet=1"
kcrypt:
  challenger:
    challenger_server: "http://ADDRESS"
    nv_index: ""
    c_index: ""
    tpm_device: ""
```

**Expected behavior**
Machine to boot correctly

**Logs**
```
[  OK  ] Found device /dev/disk/by-label/COS_ACTIVE.
[  OK  ] Reached target Initrd Root Device.
[    6.931109] e1000: ens3 NIC Link is Up 1000 Mbps Full Duplex, Flow Control: RX
[    6.932521] IPv6: ADDRCONF(NETDEV_CHANGE): ens3: link becomes ready
[    6.984757] EXT4-fs (loop0): mounted filesystem without journal. Quota mode: none.
         Starting kcrypt online mount...
[    6.951277] immucore[630]: <nil> INF mount done options=["ro","suid","dev","exec","async"] type=ext4 what=/dev/disk/by-label/COS_ACTIVE where=/sysroot
[    7.994804] EXT4-fs (vda2): mounted filesystem with ordered data mode. Quota mode: none.
[    7.959826] immucore[630]: <nil> INF mount done options=["rw","suid","dev","exec","async"] type=ext4 what=/dev/disk/by-label/COS_OEM where=/sysroot/oem
[    7.960855] immucore[630]: <nil> INF Running rootfs stage
[    8.074810] I/O error, dev fd0, sector 0 op 0x0:(READ) flags 0x0 phys_seg 1 prio class 0
[    8.076024] floppy: error 10 while reading block 0
[    8.207523] immucore[630]: 2023/03/30 08:46:33 Scaleway: Could not contact metadata service: Get "http://169.254.42.42/conf": dial tcp 169.254.42.42:80: connect: network is unreachable
[    8.208306] immucore[630]: 2023/03/30 08:46:33 Metaldata: Probing...
[...]

PR #1224, opened 2023-03-30
:seedling: :bug: Run kcrypt in rootfs stage

Itxaka:
Instead of having it as a serviec we can run it under the rootfs stage. This gives us the assurance that /oem is mounted by that time and that we run it on all boot modes except live mode

Fixes #

### DIFF
--- a/overlay/files/system/oem/21_kcrypt.yaml
+++ b/overlay/files/system/oem/21_kcrypt.yaml
@@ -1,5 +1,10 @@
-name: "Update discovery plugins"
+name: "Kcrypt"
 stages:
+  rootfs:
+    - name: "Unlock encrypted volumes"
+      if: '[ ! -f "/run/cos/live_mode" ]'
+      commands:
+        - kcrypt unlock-all
   after-upgrade:
     - name: "Update plugins"
       if: "[ $(kairos-agent state get oem.found) == 'true' ]"


Itxaka:
May fix https://github.com/kairos-io/kairos/issues/1223

mudler (dismissed):
this is looking good, at this point we can drop also the dracut package from the framework, right?

Itxaka:
> this is looking good, at this point we can drop also the dracut package from the framework, right?

yyyeeeeee....maybe.

Let me test if this actually works :D

overlay/files/system/oem/21_kcrypt.yaml
```diff
@@ -1,5 +1,10 @@
...
+name: "Kcrypt"
 stages:
+  rootfs:
+    - name: "Unlock encrypted volumes"
+      if: '[ ! -f "/run/cos/live_mode" ]'
+      commands:
+        - kcrypt unlock-all
   after-upgrade:
     - name: "Update plugins"
       if: "[ $(kairos-agent state get oem.found) == 'true' ]"
```
mudler, line 10: not part of this PR, I know, but I think the code right after it can be dropped too

mudler: we don't need to copy the plugins anymore into oem since .. a while (can't find the PR now)

Itxaka:
Yup, this fixes the issue, yay!

overlay/files/system/oem/21_kcrypt.yaml
Itxaka: will drop it in a following PR, good to know!

Itxaka:
> this is looking good, at this point we can drop also the dracut package from the framework, right?

Nope! Because we need the kcrypt binary in initramfs and the dracut module is the one inserting it. If we drop the package there is no kcrypt binary and the rootfs stage fails so we need a bit more changes before

mudler:
> > this is looking good, at this point we can drop also the dracut package from the framework, right?
> 
> Nope! Because we need the kcrypt binary in initramfs and the dracut module is the one inserting it. If we drop the package there is no kcrypt binary and the rootfs stage fails so we need a bit more changes before

ah right! but we can drop the dracut-module logics to call it :)

mudler (approved):

Issue #1228, opened 2023-03-30
:bug: :robot: Sbom is not being attached


Itxaka:
**Kairos version:**
master

**CPU architecture, OS, and Version:**
all

**Describe the bug**
sbom is supposed to be attached but a wrong flag in the sbom attachment command is breaking it

**To Reproduce**
check ci

https://github.com/kairos-io/kairos/actions/runs/4567186343/jobs/8060675440
https://github.com/kairos-io/kairos/actions/runs/4567186343/jobs/8060674917
etc...

**Expected behavior**

**Logs**

**Additional context**
probably `--output` was a valid flag on cosign 1.x but not on 2.x

mudler:
you are so ff fast :D

mudler:
temporarly reverted to avoid having a red master blocking us https://github.com/kairos-io/kairos/pull/1229

Itxaka:
> temporarly reverted to avoid having a red master blocking us #1229

it only really blocks the sbom and testing images on master. Images are still pushed :P

Issue #1055, opened 2023-03-07
:seedling: Attach and sign SBOM to images


mudler:
**Is your feature request related to a problem? Please describe.**
Now that we have a SBOM file, we can attach and sign that with cosign so it is signed along the process, and easily verifiable with the cosign CLI

**Describe the solution you'd like**
The CI to sign and attach sbom file automatically when signing images.
When we do this, also create a blog post for this explaining how users can see the SBOM of the images, validate the image and the SBOM. Also show how to use `grype` to generate a security report from it.

**Describe alternatives you've considered**

**Additional context**
https://docs.sigstore.dev/cosign/other_types/

Itxaka:
Before signing: `cosign attach sbom --sbom SBOM.spdx IMAGE:TAG` should make it

mudler:
that attaches it, we need also to sign it after it was attached:

e.g.
```
$ cosign sign --key cosign.key gcr.io/$(gcloud config get-value project)/alpine:sha256-e1c082e3d3c45cccac829840a25941e679c25d438cc8412c2fa221cf1a824e6a.sbom
```

mudler:
> > that attaches it, we need also to sign it after it was attached:
> > e.g.
> > ```
> > $ cosign sign --key cosign.key gcr.io/$(gcloud config get-value project)/alpine:sha256-e1c082e3d3c45cccac829840a25941e679c25d438cc8412c2fa221cf1a824e6a.sbom
> > ```
> 
> yes, sorry, thats what I (tried to) mean, that we need to attach _before_ signing :D
> 
> Bit slow this morning coffee

yup no worries, is just for reference for who is picking up this card

PR #1235, opened 2023-03-31
:robot: Attach and sign SBOM

mudler:
Attempt to fix: #1228 and #1055

I don't really like this bashism, but there seems to be no better way at the moment. Cosign attach sbom generates a new image - and the only way to get it's digest is parsing cosign output :shrug: 

Open to other solutions, but so far didn't had luck with other approaches, this is the output of `cosign attach sbom`:

```
WARNING: Attaching SBOMs this way does not sign them. If you want to sign them, use 'cosign attest --predicate core-alpine-arm-rpi-v2.0.0-alpha3-sbom.spdx.json --key <key path>' or 'cosign s
ign --key <key path> --attachment sbom <image uri>'.                                                                                                                                          
Uploading SBOM file for [quay.io/mudler/tests:sbom] to [quay.io/mudler/tests:sha256-b6ca290b6b4cdcca5b3db3ffa338ee0285c11744b4a6abaa9627746ee3291d8d.sbom] with mediaType [text/spdx+json].                                                                                                                                                    
```

which is, by the way, printed to STDERR (hence why the redirection).

I refrained to inspect docker, as there is no clear way to distinguish the images generated at this point ( we would have the signatures and the sbom, that are two sep. images)

### DIFF
--- a/.github/workflows/image-arm.yaml
+++ b/.github/workflows/image-arm.yaml
@@ -105,10 +105,9 @@ jobs:
         run: |
           docker push "$IMAGE:$TAG"
           image_ref=$(docker image inspect --format='{{index .RepoDigests 0}}' "$IMAGE:$TAG")
-          cosign sign $image_ref
           spdx=$(ls build/*.spdx.json)
           cosign attach sbom --sbom $spdx $image_ref
-          cosign sign ${image_ref}.sbom
+          cosign sign $image_ref --attachment sbom
           # in-toto attestation
           cosign attest --type spdx --predicate $spdx $image_ref
       - name: Upload results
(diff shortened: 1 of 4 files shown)


mudler:
this also adds in-toto attestations

.github/workflows/image-arm.yaml
```diff
@@ -103,8 +103,14 @@ jobs:
           TAG: "latest"
           COSIGN_YES: true
         run: |
-          docker push "$IMAGE:$TAG" # Otherwise .RepoDigests will be empty for some reason
-          cosign sign $(docker image inspect --format='{{index .RepoDigests 0}}' "$IMAGE:$TAG")
+          docker push "$IMAGE:$TAG"
+          image_ref=$(docker image inspect --format='{{index .RepoDigests 0}}' "$IMAGE:$TAG")
+          cosign sign $image_ref
```
Itxaka, line 108: I woudl remove this and `cosign sign ${image_ref}.sbom` and just use `cosign sign --attachment sbom $image_ref` to see if it signs the whole thing, so we only have to sign once

mudler:
ok small enhancements here and there, thanks @Itxaka!

Itxaka (approved):

mudler:
This can be tested only on master pipelines.. so..

Issue #1249, opened 2023-04-04
:bug: kairos-cli version is missing and register command has some dirty output


mauromorales:
missing version:

```
zeno% ./kairos-cli
NAME:
   kairos-agent-provider - kairos CLI to bootstrap, upgrade, connect and manage a kairos network

USAGE:
   kairos-agent-provider [global options] command [command options] [arguments...]

VERSION:
   0.0.0
```

Dirty outpu

```
zeno% ./kairos-cli register --config ./config.yaml --device auto ~/Pictures/kairos3.png                              
Sending registration payload, please wait
2023/04/04 23:33:48 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Receive-Buffer-Size for details.
Not enough nodes
map[]
Not enough nodes
map[healthcheck:map[12D3KooWPa3uoaFDVszwPrELAmriHtxEqfeGEqK2ox71gBuhwoQD:"2023-04-04T21:33:54Z" 12D3KooWRFhnDq5EeCzRiHKEKGSp2cAnkSCjXPxvnVw54KVmFnuw:"2023-04-04T21:33:51Z"] pairing:map[12D3KooWRFhnDq5EeCzRiHKEKGSp2cAnkSCjXPxvnVw54KVmFnuw:"ok" data:{"cc":"#cloud-config\nusers:\n  - name: kairos\n    passwd: kairos\n","device":"auto"}] presence:map[12D3KooWPa3uoaFDVszwPrELAmriHtxEqfeGEqK2ox71gBuhwoQD:"" 12D3KooWRFhnDq5EeCzRiHKEKGSp2cAnkSCjXPxvnVw54KVmFnuw:""]]
Pairing in progress
[...]

PR #1252, opened 2023-04-05
:robot: Add version to goreleaser

mauromorales:
Pass the version to goreleaser, so the artifact that can be dowloaded includes it

relates to #1249

### DIFF
--- a/.github/workflows/release_bin.yaml
+++ b/.github/workflows/release_bin.yaml
@@ -9,10 +9,11 @@ jobs:
   goreleaser:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release_bin.yaml
```diff
@@ -23,4 +24,4 @@ jobs:
           version: latest
           args: release --rm-dist
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mudler, line 26: is this an oversight? not sure if this is required?

mauromorales: oopsie yes, my bad

mudler (approved):
Nais! I guess we need the same on the provider?

PR #1256, opened 2023-04-05
:bug: Do not merge all cmdline arguments to generic config

### LINKED ISSUE #1255: :bug: Invalid configs with v2.0.0-rc*
**Kairos version:**
v2.0.0-rc*

**Describe the bug**
Seems installation with webui are broken (maybe not only affecting webui).

The collector seems to not write down correctly the config supplied by the user.

Input config for installation:

```yaml
#cloud-config
hostname: nuc-{{ trunc 4 .MachineID }}
k3s:
    enabled: true
users:
    - name: kairos
      ssh_authorized_keys:
        - github:mudler
 ```

The config file after install was the following (in `/oem`):

```yaml
#cloud-config

BOOT_IMAGE: (hd0)/boot/kernel
cdroot: true
console: ttyS0
net:
  ifnames: "1"
nodepair:
  enable: true
nomodeset: true
rd:
  cos:
    disable: true
  live:
    dir: /
    squashimg: rootfs.squashfs
root: live:CDLABEL=COS_LIVE
selinux: "0"
vga: "795"
```

few things to note here:
- the previous code (before the collector was introduced) was saving *only* what was mapped to the config structure from the cmdline, so you don't end up mixing bootargs in the cloud config file that are not in any structure (we can't set there k3s settings for instance, but it is enough specifying a remote `config_url` to fetch)
- there are ONLY cmdline configs in there now - the user config is completely discarded
- there are mixtures of extra cmdlines which are not pertinent to the installation

**To Reproduce**
Get an ISO (I've tested with  kairos-opensuse-leap-v2.0.0-rc4-k3sv1.23.17+k3s1.iso ) and try to install from the WebUI

**Expected behavior**
To have

```yaml
#cloud-config
hostname: nuc-{{ trunc 4 .MachineID }}
k3s:
    enabled: true
users:
    - name: kairos
      ssh_authorized_keys:
        - github:mudler
 ```

In the /oem files

**Logs**

**Additional context**
This is high prio, as currently blocks https://github.com/kairos-io/kairos/issues/754

### PR DESCRIPTION
**Pull request by mauromorales**
**What this PR does / why we need it**:R

White list options that can be added in the cmdline

relates to #1255

### DIFF
--- a/pkg/config/collector/collector.go
+++ b/pkg/config/collector/collector.go
@@ -4,6 +4,7 @@ package collector
 
 import (
 	"fmt"
+	"github.com/kairos-io/kairos-sdk/machine"
 	"io"
 	"net/http"
 	"os"
@@ -13,10 +14,8 @@ import (
 	"unicode"
 
 	"github.com/avast/retry-go"
-	"github.com/google/shlex"
 	"github.com/imdario/mergo"
 	"github.com/itchyny/gojq"
-	"github.com/kairos-io/kairos-sdk/unstructured"
 	"gopkg.in/yaml.v1"
 )
 
@@ -202,40 +201,22 @@ func listFiles(dir string) ([]string, error) {
 // ParseCmdLine reads options from the kernel cmdline and returns the equivalent
 // Config.
 func ParseCmdLine(file string) (*Config, error) {
-	result := &Config{}
+	result := Config{}
+	d, _ := machine.DotToYAML(file)
+	fmt.Println(string(d))
+	cmdLineConfig := map[string]interface{}{}
+	_ = yaml.Unmarshal(d, &cmdLineConfig)
 
-	if file == "" {
-		file = "/proc/cmdline"
+	configUrl, ok := cmdLineConfig["config_url"]
+	if ok {
+		result["config_url"] = configUrl
 	}
-	dat, err := os.ReadFile(file)
-	if err != nil {
-		return result, err
-	}
-
-	d, err := unstructured.ToYAML(stringToConfig(string(dat)))
-	if err != nil {
-		return result, err
-	}
-	err = yaml.Unmarshal(d, &result)
-
-	return result, err
-}
-
-func stringToConfig(s string) Config {
-	v := Config{}
-
-	splitted, _ := shlex.Split(s)
-	for _, item := range splitted {
-		parts := strings.SplitN(item, "=", 2)
-		value := "true"
-		if len(parts) > 1 {
-			value = strings.Trim(parts[1], `"`)
-		}
-		key := strings.Trim(parts[0], `"`)
-		v[key] = value
+	options, ok := cmdLineConfig["options"]
+	if ok {
+		result["options"] = options
 	}
 
-	return v
+	return &result, nil
 }
 
 // ConfigURL returns the value of config_url if set or empty string otherwise.
--- a/pkg/config/collector/collector.go
+++ b/pkg/config/collector/collector.go
@@ -93,13 +93,13 @@ func (cs Configs) Merge() (*Config, error) {
 	return result, nil
 }
 
-func Scan(o *Options) (*Config, error) {
+func Scan(o *Options, cmdLineFilter interface{}) (*Config, error) {
 	configs := Configs{}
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by jimmykarily — APPROVED** [review]

**Comment by mudler** [comment]
I think there is something still off, the old implementation relied on unmarshalling to the config, so it supported all the field of the config (and no specifics around `config_url`) see the comment here: https://github.com/kairos-io/kairos/commit/b6ae4b58c07d2a2e6e7aee056aff973e02d43026#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5L184

**Comment by jimmykarily** [comment]
> I think there is something still off, the old implementation relied on unmarshalling to the config, so it supported all the field of the config (and no specifics around `config_url`) see the comment here: [b6ae4b5#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5L184](https://github.com/kairos-io/kairos/commit/b6ae4b58c07d2a2e6e7aee056aff973e02d43026#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5L184)

I thought you said, we only read the `config_url` from the cmdline. Otherwise, what is so special about kairos config that we permit is through cmdline but we don't permit any other configuration (e.g. kcrypt)? The collector is a generic package and should not have to know about kairos specific configuration.

**Comment by mudler** [comment]
> > I think there is something still off, the old implementation relied on unmarshalling to the config, so it supported all the field of the config (and no specifics around `config_url`) see the comment here: [b6ae4b5#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5L184](https://github.com/kairos-io/kairos/commit/b6ae4b58c07d2a2e6e7aee056aff973e02d43026#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5L184)
> 
> I thought you said, we only read the `config_url` from the cmdline. 

Right, but looking at the old code is just smarter - it allows not only to read `config_url`, but for instance also pass by install options to it, which are generic to the agent. 

> Otherwise, what is so special about kairos config that we permit it through cmdline but we don't permit any other configuration (e.g. kcrypt)? The collector is a generic package and should not have to know about kairos specific configuration.

It's the agent configuration in this case, which is generic and then extends as well as other providers. We can't read cmdline from the agent for the providers, as those could have their own options, but, we can read all the agent configs as the agent is shared by all the providers

**Comment by mauromorales** [comment]
Yeah I wanted to make it like the precious code but thought it would be tricky to properly separate concerns of collection and configuration if collector needs to know about kairos config specifics. But will make it work like that and we can figure out a better way to split responsibly after the release

**Comment by mauromorales** [comment]
I thought this was fixed by passing an interface but turns out then the config.Config get's lost into a map[interface{}]interface{} looking into it and changing the tests so this doesn't get unnoticed

**Comment by mauromorales** [comment]
sorry somehow broke unit tests, on it

**Comment by mauromorales** [comment]
oki should be good now

File: pkg/config/collector/collector.go [hunk 1]
```diff
@@ -201,41 +201,24 @@ func listFiles(dir string) ([]string, error) {
...
-func ParseCmdLine(file string) (*Config, error) {
-	result := &Config{}
-
-	if file == "" {
-		file = "/proc/cmdline"
-	}
-	dat, err := os.ReadFile(file)
+func ParseCmdLine(file string, filter func(d []byte) ([]byte, error)) (*Config, error) {
+	result := Config{}
+	dotToYAML, err := machine.DotToYAML(file)
```
mudler, line 206: not related to the change at hand, but `DotToYAML` doesn't sound something belonging to "machine" (SDK)

pkg/config/config.go
```diff
@@ -103,7 +119,7 @@ func Scan(opts ...collector.Option) (c *Config, err error) {
 		return result, err
 	}
 
-	genericConfig, err := collector.Scan(o)
+	genericConfig, err := collector.Scan(o, FilterKeys)
```
mudler, line 122: wouldn't that skip everything else? what about providers?

mudler (approved):
looks good!

PR #1260, opened 2023-04-06
:bug: Add yaml extension to temp file created by webui

mauromorales:

The collector is stricter with the files it reads and requires that they have proper extension to be considered. This PR adds the yaml extension to the temporary file created by the webUI

relates to #1255


mudler (approved):
Looking good! good catch!

Issue #1259, opened 2023-04-06
:seedling: Keep OS_ID and OS_VERSION untouched


mudler:
**Is your feature request related to a problem? Please describe.**
Operators that install drivers on the machines as https://github.com/NVIDIA/gpu-operator, might just read the `/etc/os-release` to identify the system. As we modify the OS_ID and OS_VERSION to annotate the flavor and version, operators can't identify anymore clearly the system.

**Describe the solution you'd like**
We can just use our own variables, augmenting the /etc/os-release file.

**Describe alternatives you've considered**
N/A

**Additional context**
This likely will have _some_ impact on parts of the code that relies on reading OS_ID and OS_VERSION, but that should be def. possible

PR #1263, opened 2023-04-06
:seedling: Prefix variable names with distro name in os-release

mauromorales:
Relates to #1259


overlay/files/usr/lib/os-release.tmpl
```diff
@@ -1,14 +1,13 @@
...
-ANSI_COLOR="0;32"
-BUG_REPORT_URL="${BUG_REPORT_URL}"
-HOME_URL="${HOME_URL}"
-IMAGE_REPO="${OS_REPO}"
-IMAGE_LABEL="${OS_LABEL}"
-GITHUB_REPO="${GITHUB_REPO}"
-VARIANT="${VARIANT}"
-FLAVOR="${FLAVOR}"
\ No newline at end of file
+KAIROS_NAME="${OS_NAME}"
```
mauromorales, line 1: @mudler  Would it make sense to prefix all variables with KAIROS? Seems to be the recommendation, but maybe I'm missing something

> The file is extensible? Awesome! I want a new field XYZ= in it! Sure, it's extensible, and we are happy if distributions extend it. Please prefix your keys with your distribution's name however. Or even better: talk to us and we might be able update the documentation and make your field standard, if you convince us that it makes sense.

http://0pointer.de/blog/projects/os-release

mudler: yep sounds about right. but we should be careful, as I remind we read from those in several places down the line, for instance to get the current flavor  https://github.com/kairos-io/kairos-sdk/blob/77ed7ba29815f03090ad63bf0ec3e56dfa83a229/utils/utils.go#L98

saamalik:
@mauromorales @mudler - what are your thoughts on adding variables to os-release vs using a completely separate file, e.g: kairos-release ? Any best practices here?

mauromorales:
> @mauromorales @mudler - what are your thoughts on adding variables to os-release vs using a completely separate file, e.g: kairos-release ? Any best practices here?

I think the recommendation from systemd is for distros to move away from `/etc/distro_name-release` files, but for compatibility reasons those are still in place. I'd say that it would be better to stick with `/etc/os-release` and just extend it, since we can have our variable names there without touching the original variables from the underlying distro.

From what I can see there's also a way to extend via `/usr/lib/extension-release.d/extension-release.IMAGE` but maybe that's a bit too much at the moment? Or perhaps I'm just misunderstanding what it's meant for. See: https://www.man7.org/linux/man-pages/man5/os-release.5.html

mudler:
@mauromorales I suspect we need to update the sdk first and bump it here? or am I missing something? https://github.com/kairos-io/kairos-sdk/blob/77ed7ba29815f03090ad63bf0ec3e56dfa83a229/utils/utils.go#L98

mudler:
another place where it is used: https://github.com/kairos-io/kairos-sdk/blob/77ed7ba29815f03090ad63bf0ec3e56dfa83a229/state/state.go#L157

mauromorales:
@mudler yeah indeed, this one needs to be the first to be merged before all other os-release related PRs https://github.com/kairos-io/kairos-sdk/pull/12

docs/content/en/docs/Architecture/container.md
```diff
@@ -51,22 +51,22 @@ $ docker pull quay.io/kairos/core-alpine:v1.1.0
 We can run it locally with docker as a container to inspect it, as it is runnable:
 
 ```
-$ docker run -ti --rm quay.io/kairos/core-alpine:v1.1.0
+$ docker run -ti --rm quay.io/kairos/core-alpine:v2.0.0
```
mudler, line 54: I think this is wrong, we don't have core-alpine anymore, but core-alpine-ubuntu, etc. 

Maybe just use debian? should be the same I guess

docs/content/en/docs/Architecture/container.md
```diff
@@ -51,22 +51,22 @@ $ docker pull quay.io/kairos/core-alpine:v1.1.0
...
-BUG_REPORT_URL="https://github.com/kairos-io/kairos/issues"
-HOME_URL="https://github.com/kairos-io/kairos"
-IMAGE_REPO="quay.io/kairos/core-alpine"
-IMAGE_LABEL="latest"
-GITHUB_REPO="kairos-io/kairos"
-VARIANT="core"
-FLAVOR="alpine"/ #
+...
+KAIROS_NAME="kairos-core-alpine"
+KAIROS_VERSION="v1.1.0"
```
mudler, line 58: pulling 2.0, but 1.0 is displayed here

mudler (approved):
Just few nitpicks, but looking good here!

docs/content/en/docs/Architecture/container.md
mauromorales: yup good idea, decided to go with a site variable for these, won't replace everywhere to avoid making the PR too big

docs/content/en/docs/Architecture/container.md
mauromorales: good point, using variables now instead

mudler:
mmm upgrade tests seems to fail as they compare to /etc/os-release, but probably looking at the old vars:

```
      +run-qemu-test *failed* | Cleaning /tmp/342500181
      +run-qemu-test *failed* | • [FAILED] [2383.137 seconds]
      +run-qemu-test *failed* | k3s upgrade manual test upgrades [It] can upgrade to current image [upgrade-with-cli]
      +run-qemu-test *failed* | /build/tests/upgrade_cli_test.go:47

      +run-qemu-test *failed* |   [FAILED] Timed out after 600.000s.

      +run-qemu-test *failed* |   Expected
      +run-qemu-test *failed* |       <string>: 15.5 Beta
      +run-qemu-test *failed* |       
      +run-qemu-test *failed* |   to contain substring
      +run-qemu-test *failed* |       <string>: v
      +run-qemu-test *failed* |   In [It] at: /build/tests/upgrade_cli_test.go:81 @ 04/14/23 15:45:27.524
      +run-qemu-test *failed* | ------------------------------
```

examples/bundle/Dockerfile
```diff
@@ -7,8 +7,8 @@ RUN wget https://github.com/ipfs/kubo/releases/download/v0.15.0/kubo_v0.15.0_lin
 RUN tar xvf kubo.tar.gz
 RUN mv kubo/ipfs /usr/bin/ipfs
 RUN mkdir -p /usr/lib/extension-release.d/
-RUN echo ID=kairos > /usr/lib/extension-release.d/extension-release.kubo
-RUN echo VERSION_ID=$VERSION >> /usr/lib/extension-release.d/extension-release.kubo
+RUN echo KAIROS_ID=kairos > /usr/lib/extension-release.d/extension-release.kubo
+RUN echo KAIROS_VERSION_ID=$VERSION >> /usr/lib/extension-release.d/extension-release.kubo
```
mudler, line 11: I think it's other way around systemd-sysext will look into ID.

See docs: https://www.freedesktop.org/software/systemd/man/systemd-sysext.html

However, seems we can just put "_any":
```
A simple mechanism for version compatibility is enforced: a system extension image must carry a /usr/lib/extension-release.d/extension-release.$name file, which must match its image name, that is compared with the host os-release file: the contained ID= fields have to match unless "_any" is set for the extension. If the extension ID= is not "_any", the SYSEXT_LEVEL= field (if defined) has to match. If the latter is not defined, the VERSION_ID= field has to match instead. If the extension defines the ARCHITECTURE= field and the value is not "_any" it has to match the kernel's architecture reported by [uname(2)](https://man7.org/linux/man-pages/man2/uname.2.html) but the used architecture identifiers are the same as for ConditionArchitecture= described in [systemd.unit(5)](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#). System extensions should not ship a /usr/lib/os-release file (as that would be merged into the host /usr/ tree, overriding the host OS version data, which is not desirable). The extension-release file follows the same format and semantics, and carries the same content, as the os-release file of the OS, but it describes the resources carried in the extension image.
```

mauromorales: gotcha, thanks!

mudler (approved):
Looking good here, nice @mauromorales !

docs/content/en/docs/Architecture/container.md
```diff
@@ -45,28 +45,28 @@ The Image support matrix in [here](/docs/reference/image_matrix) lists all the c
...
+KAIROS_VERSION="{{< kairosVersion >}}"
+KAIROS_ID="kairos"
+KAIROS_ID_LIKE="kairos-core-{{< flavor >}}"
+KAIROS_VERSION_ID="{{< kairosVersion >}}"
+KAIROS_PRETTY_NAME="kairos-core-{{< flavor >}} {{< kairosVersion >}}"
+KAIROS_BUG_REPORT_URL="https://github.com/kairos-io/kairos/issues"
+KAIROS_HOME_URL="https://github.com/kairos-io/kairos"
+KAIROS_IMAGE_REPO="{{< registryURL >}}/core-{{< flavor >}}"
+KAIROS_IMAGE_LABEL="latest"
+KAIROS_GITHUB_REPO="kairos-io/kairos"
```
mudler, line 67: @mauromorales last one: https://github.com/kairos-io/kairos/blob/d7d9478f2320956b83f566a2f583ae388b8baf09/internal/agent/upgrade.go#L35

mudler: oh, right, we care of that automatically. nevermind, ignore me!

examples/bundle/Dockerfile
mauromorales: even with `ID="_any"`, this one is still causing trouble  https://github.com/kairos-io/kairos/actions/runs/4732001891/jobs/8401102181#step:7:2560

mauromorales: @mudler unfortunately the `_any` special value, was only introduced in systemd 252 https://github.com/systemd/systemd/blob/main/NEWS#L1116 while the leap box where we run tests against has only 249 atm. I'm going to switch this test to use tumbleweed which comes with 253 already and works properly.

mudler (approved):
Looks good!

mudler:
since we are at it, care to update the docs mentioning that _any can be used for systemd-XXX versions? I think that can be helpful for others as well if we make it a bit more visible in our docs - there are cases, like static golang binaries where specifying a version might be just redundant

docs/content/en/docs/Advanced/livelayering.md
```diff
@@ -68,31 +68,29 @@ For example, a bundle can be defined as a naked container image containing only
 
 Consider the following Dockerfile to create an extension which adds `/usr/bin/ipfs` to the system:
 
+
+{{% alert title="Note" %}}
+Note that systemd extensions require an extension-release file, which can be used to validate different aspects of the system being run.
```
mauromorales, line 73: @mudler yeah makes sense. How about this?

mudler: perfect!

mudler (approved):

mauromorales:
@mudler could you merge and change the required test from "leap" to "tumbleweed"?

image

mudler (approved):

PR #1269, opened 2023-04-11
Add cypress tests for web UI

Itxaka:

This adds basic webUI tests to our workflows based of cypress.
Currently the deployment and test is simplified, as it runs a local webui agent inside earthly and tests several parts on the UI and some of the functionality (validation)

Hopefully in the near future this is expanded to use a full qemu VM with the webui running and we can trigger the full install from there although its  a bit nastey as it would be earthly -> ginkgo -> qemu -> cypress

Fixes #

### DIFF
--- a/internal/webui/public/cypress/e2e/spec.cy.js
+++ b/internal/webui/public/cypress/e2e/spec.cy.js
@@ -1,0 +1,49 @@
+describe('Basic Tests for webui', () => {
+  beforeEach(() => {
+    cy.visit('/')
+    cy.intercept({
+      method: 'POST',
+      url: '/validate',
+    }, {log: false}).as('validate')
+  })
+  it("basic items on the ui exist", () => {
+    cy.contains('#cloud-config').should("exist").should("be.visible")
+    cy.contains('Welcome to the Installer!').should("exist").should("be.visible")
+    cy.contains("p a", "cloud-config config configuration file")
+        .should("have.attr", "href", "/local/docs/reference/configuration/")
+    cy.get("#cloud-config-help a")
+        .should("have.attr", "href", "/local/docs/examples/")
+    cy.get("#installation-device").should("have.value", "auto")
+
+    // footer
+    cy.get("a .fa-github").should("exist").parent().should("have.attr", "href", "https://github.com/kairos-io/kairos")
+    cy.get("a .fa-book").should("exist").parent().should("have.attr", "href", "https://kairos.io/docs")
+    cy.get("#reboot-checkbox").should("exist").should("not.be.checked")
+    cy.get("#poweroff-checkbox").should("exist").should("not.be.checked")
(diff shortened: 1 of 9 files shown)


mauromorales (approved):
Noiiice! :clap:

mudler (dismissed):
Lovely!!!

internal/webui/public/cypress/e2e/spec.cy.js
```diff
@@ -0,0 +1,49 @@
...
+    cy.get("button").should("exist").invoke("text").should("equal", "Install")
+  })
+  it('validation works', () => {
+    cy.get('.CodeMirror')
+        .first()
+        .then((editor) => {
+          editor[0].CodeMirror.setValue('');
+        });
+
+    cy.get(".CodeMirror textarea").type("#cloud-config{enter}users:{enter}  - name: caca", {force: true})
```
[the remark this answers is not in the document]
Itxaka: nooooo, crap. It was a placeholder lmao

internal/webui/public/index.html
```diff
@@ -105,21 +105,21 @@ <h1>Welcome to the Installer!</h1>
 
               codeMirrorEditor.on('change', () => content = codeMirrorEditor.getValue())
             " 
-            x-on:keydown="validateYAML"
+            x-on:keydown.debounce="validateYAML"
```
Itxaka, line 108: This is so there is a delay after keys of 250ms. Otherwise, this keeps firing while you write and can send a lot of data to the validation endpoint which is not nice. Having this means that after the last keydown has been sent and no other one key has been pressed in 250ms then event fires. This was causing a lot of calls to the validate endpoint while writing (one per key!)

internal/webui/public/cypress/e2e/spec.cy.js
```diff
@@ -0,0 +1,49 @@
...
+    cy.get('.CodeMirror')
+        .first()
+        .then((editor) => {
+          editor[0].CodeMirror.setValue('');
+        });
+    cy.get(".CodeMirror textarea").type("blablabla", {force: true})
+    cy.get("#validator-alert").invoke("text").should("match", /Failed validating syntax/)
+
+  })
+  it('should install', function () {
```
Itxaka, line 46: This doesnt really work, it fails to find the agent but we should probably hijack the call and check that all data was passed correctly.

mauromorales (commented):
Doesn't have to be in this PR, but I think it might make sense to somehow keep track of the report. So we can give an idea of how much the coverage went up/down

Itxaka:
> Doesn't have to be in this PR, but I think it might make sense to somehow keep track of the report. So we can give an idea of how much the coverage went up/down

that is a really good point! Gonna have a look to see if we can also use codecov for this or it will get confused and break :D

PR #1273, opened 2023-04-11
:bug: Replace kairos to core in arm build and release workflows

mauromorales:
I am uncertain if I'm missing something or if there was a confusion here with the naming


mauromorales:
@mudler oki thanks for the clarification :)

PR #1290, opened 2023-04-13
Pass VERSION flag including v2 in path

### LINKED ISSUE #1287: :bug: Kairos agent version in v2 shows as 0.0.0
I'm pretty sure this was fixed at some point but in RC6 and v2 it shows as 0.0.0.

This issue applies to both the agent delivered within the image, and the one released by the goreleaser, so make sure all the following include the version

- [x] core images agent
- [x] standard images agent
- [x] kairos artifact (was ok)
- [x] provider kairos artifact (was ok)

mauromorales:
Since the code is built on v2, the VERSION to build it needs to also have v2 in the path

fixes #1287

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -2,7 +2,7 @@
 project_name: kairos-agent
 builds:
   - ldflags:
-      - -w -s -X "github.com/kairos-io/kairos/internal/common.VERSION={{.Env.VERSION}}"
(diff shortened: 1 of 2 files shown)


Itxaka (approved):
Oooh good catch!

mudler (approved):
Indeed! Good catch!

PR #1309, opened 2023-04-17
:bug: Fix rpi initrd not being linked

### LINKED ISSUE #1300: :bug: v2.0.0 does not boot anymore on rpi
**Kairos version:**

kairos-opensuse-leap-arm-rpi:v2.0.0-k3sv1.26.3-k3s1

**Describe the bug**

image
image

build with:
```
docker run -v $PWD:/HERE -v /var/run/docker.sock:/var/run/docker.sock --privileged -i --rm --entrypoint=/build-arm-image.sh quay.io/kairos/osbuilder-tools:latest \
 --model rpi64 \
 --local \
 --config /HERE/cloud-config.yaml \
 --efi-dir /HERE/boot \
 --docker-image kairos-rpi-ogi /HERE/build/out.img
```

**To Reproduce**

**Expected behavior**

**Logs**

**Additional context**

Itxaka:
Looks like I added the skip to the rpi images but those have dracut and we should use that to generate the initrd and link it

Fixes #1300

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -350,7 +350,7 @@ base-image:
     END
 
 
-    IF [[ "$FLAVOR" =~ ^alpine.* ]] || [[ "$FLAVOR" =~ .*-arm-rpi$ ]]
+    IF [[ "$FLAVOR" =~ ^alpine.* ]]
         # no dracut on those flavors, do nothing
     ELSE
         # Regenerate initrd if necessary


Earthfile
```diff
@@ -350,7 +350,7 @@ base-image:
     END
 
 
-    IF [[ "$FLAVOR" =~ ^alpine.* ]] || [[ "$FLAVOR" =~ .*-arm-rpi$ ]]
+    IF [[ "$FLAVOR" =~ ^alpine.* ]]
```
mauromorales, line 353: I wonder how we could document such kind of information ... like it would be nice if instead of checking here on the flavor, there could be some sort of file where we check "hasDracut?" and if the current flavor happens to be listed there it's true

mauromorales (approved):
lgtm

Earthfile
Itxaka: not only that but all the differences between distros on if they auto-lin the Image or we have to do it, if its a symlink or a regular file. pita!

Itxaka: in any case, alpine wont have dracut unless you install it for testing so maybe we can end up with ALL of them just running dracut somehow :D

PR #1310, opened 2023-04-17
:bug: Ensure data source is ready during reset

### LINKED ISSUE #1178: :bug: Reset does not work on physical hardware (INTEL NUC as Example)
**Kairos version:**
1.5,1.6,2.0

x86 12 Generation Intel.  OS version does not matter.  The result is the same.

**Describe the bug**
When sending a node to reset, the node never comes back online.  It simply sits at a blank screen.  The node boots into recovery, but the reset command is never applied.

**To Reproduce**
Load an Intel NUC and send it to reset.

**Expected behavior**
The Node should reset back to the original state.

mudler:

Fixes #1178

### DIFF
--- a/internal/agent/reset.go
+++ b/internal/agent/reset.go
@@ -64,6 +64,8 @@ func Reset(dir ...string) error {
 	lock.Lock()
 	args := []string{"reset"}
 
+	ensureDataSourceReady()
+
 	bus.Manager.Publish(sdk.EventBeforeReset, sdk.EventPayload{}) //nolint:errcheck
 
 	optsArgs := optsToArgs(options)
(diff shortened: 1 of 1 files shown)


mauromorales (approved):
lgtm

Itxaka (approved):
does this make sense? I mean, 5 minutes waiting for source+60 seconds for reset is quite a lot...we need to fix the datasource wait!

mudler:
> does this make sense? I mean, 5 minutes waiting for source+60 seconds for reset is quite a lot...we need to fix the datasource wait!

yep agreed, data source pulling code is rather suboptimal :/

mudler:
Ok all green, merging!

PR #1332, opened 2023-04-21
:robot: Fix release tests

Itxaka:
Looks like we only get a small subset of the releases so we cant check for a specific number in the list that we get

Fixes #


Itxaka:
This started failing once we added a new test as it changed the list. While Im fixing the release function to get the full list, this unblocks the unit-test job.


No need to wait for all jobs to run, will cancel it and leave the unit-test job only.

mauromorales (approved):

PR #1333, opened 2023-04-21
:sparkles: Allow showing pre-releases on upgrade options

Itxaka:
By default get 30 releases and skip any pre-releases found. Allow toggling showing pre-releases both for upgrade and for listing them

Fixes #

### DIFF
--- a/pkg/github/releases_test.go
+++ b/pkg/github/releases_test.go
@@ -16,7 +16,16 @@ func TestReleases(t *testing.T) {
 
 var _ = Describe("Releases", func() {
 	It("can find the proper releases in order", func() {
-		releases, err := github.FindReleases(context.Background(), "", "kairos-io/kairos")
+		releases, err := github.FindReleases(context.Background(), "", "kairos-io/kairos", false)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(releases)).To(BeNumerically(">", 0))
+		// Expect the one at the bottom to be the first "real" release of kairos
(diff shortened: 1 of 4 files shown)


mauromorales (approved):
lgtm

pkg/github/releases_test.go
```diff
@@ -16,7 +16,16 @@ func TestReleases(t *testing.T) {
...
-		releases, err := github.FindReleases(context.Background(), "", "kairos-io/kairos")
+		releases, err := github.FindReleases(context.Background(), "", "kairos-io/kairos", false)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(len(releases)).To(BeNumerically(">", 0))
+		// Expect the one at the bottom to be the first "real" release of kairos
+		Expect(releases[len(releases)-1].Original()).To(Equal("v1.0.0"))
+		// Expect the first one to be greater than the last one
+		Expect(releases[0].GreaterThan(releases[len(releases)-1]))
+	})
+	It("can find the proper releases in order with prereleases", func() {
```
jimmykarily, line 27: There is no check that pre-releases were actually fetched. The boolean could as well be ignored.

Itxaka: nope, it only confirms that in both modes it can get a list of releases correctly. I would need a test repo with fixed releases to test this properly or be able to mock the response somehow, otherwise our releases keep changing and there is no way of having a test that wont fail in the near future (hint, this already failed as soon as we published a new release lol)

PR #1338, opened 2023-04-21
:robot: Download only the x86 iso for testing

Itxaka:
As we now have also arm generic images, the download-latest job is now downsloading both isos as they match in name.

Adds a tighter regex to download the proper images

Fixes #


mauromorales (approved):
lgtm

mudler (approved):

PR #1344, opened 2023-04-24
:robot: Fix releases test

### LINKED ISSUE #1345: :robot: CI test relies on the results of GitHub releases, making it brittle
This test https://github.com/kairos-io/kairos/blob/fdff7517834e0ff24702b68a75ab86ffcc2f8438/pkg/github/releases_test.go#L18

Uses the result of a call to GitHub list of releases, but these are changing with time. I think it would be better to record this call, so a) it does not make the call live for every run of the test and b) so we can reliably expect a certain release number to be fixed all the time

Itxaka:
Leftover broken tests for releases functions

Fixes #1345


mauromorales (approved):
Nice, I opened an issue but you're too fast :)

LGTM

PR #1359, opened 2023-04-26
:sparkles: Compress firmware and kernel modules on ubuntu

Itxaka:

Fixes #


Itxaka:
Need to compare sizes but the firmware goes from 962M to 377M and modules from 584M to 146M but the final iso doesnt really translate into the same gain...

Earthfile
```diff
@@ -348,6 +348,13 @@ base-image:
         COPY github.com/kairos-io/immucore:$IMMUCORE_DEV_BRANCH+dracut-artifacts/10-immucore.conf /etc/dracut.conf.d/10-immucore.conf
     END
     # END
+    IF [ "$FLAVOR" = "ubuntu-20-lts" ] || [ "$FLAVOR" = "ubuntu" ] || [ "$FLAVOR" = "ubuntu-22-lts" ]
```
mauromorales, line 351: I guess we need to remember to also do this with the ubuntu RPi images once we have them

mauromorales (dismissed):
lgtm

Earthfile
Itxaka: Good point. I tried a regex here but somehow I could not get it to work :sob:

Itxaka:
I dunno, this doesn't seem alrigth. While isos boot and I can confirm that the modules are indeed compressed, the size is not that big. A mere 100Mb as maximum...

Itxaka:
ok, in teh final system there is a 800Mb difference between a compressed and uncompressed image installed. That makes more sense, the active.img is used 1.2Gb vs 2.0Gb between them.

mauromorales (approved):
lgtm

PR #1362, opened 2023-04-26
:robot: run install test under earthly

Itxaka:

Fixes #

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -225,7 +225,7 @@ jobs:
   install-test:
     needs:
       - build
-    runs-on: macos-12
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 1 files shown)


Itxaka:
I see now what @jimmykarily meant, this is slow as heck.

jimmykarily:
But running it outside earthly doesn't make it automatically faster. We need to enable KVM the here are lost of hoops to jump through to achieve that.

Itxaka:
> But running it outside earthly doesn't make it automatically faster. We need to enable KVM the here are lost of hoops to jump through to achieve that.

kind of does lol. Im guessing the extra layer of dockering

jimmykarily:
I managed to get tests to run with KVM here: https://github.com/kairos-io/provider-kairos/pull/306 (self-hosted runners). When that gets merged we can plan how to do that for this repo too. We may need more resources for that.

PR #1369, opened 2023-05-01
:robot: Allow building agent from branches

Itxaka:

Fixes #

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -267,6 +267,15 @@ base-image:
           systemctl enable cos-setup-network.service
     END
 
+    # TEST KAIROS-AGENT FROM BRANCH
+    ARG KAIROS_AGENT_DEV
+    ARG KAIROS_AGENT_DEV_BRANCH=main
+    IF [ "$KAIROS_AGENT_DEV" = "true" ]
+        RUN rm -rf /usr/bin/kairos-agent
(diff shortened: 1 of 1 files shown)


Earthfile
```diff
@@ -267,6 +267,15 @@ base-image:
           systemctl enable cos-setup-network.service
     END
 
+    # TEST KAIROS-AGENT FROM BRANCH
+    ARG KAIROS_AGENT_DEV
+    ARG KAIROS_AGENT_DEV_BRANCH=master
```
mauromorales, line 272: Not a strong opinion but can we use main ... most other repos I think are using main

Earthfile
```diff
@@ -267,6 +267,15 @@ base-image:
...
+    IF [ "$KAIROS_AGENT_DEV" = "true" ]
+        RUN rm -rf /usr/bin/kairos-agent
```
mudler, line 274: shouldn't this be part of the framework instead?

Earthfile
mauromorales: actually I just double checked and the agent branch is main so I think it will need to be the default

Earthfile
Itxaka: Was gonna add it there, but this is mainly for dev... and Im guessing for building full isos to try them manually with changes. Not sure if needed on the framework? Im worried that it migth end up in a pushed image or something :/

mauromorales (approved):
lgtm

Issue #1234, opened 2023-03-31
:robot: Have linting and make it required on every one of Kairos repositories


mauromorales:
Standards applied on the Kairos main repo should also be required on every other repository. There will be linting steps that might not make sense, those can be ignored, but we should to have this as a standard of our different codebases.

- [x] https://github.com/kairos-io/kairos/
- [ ] https://github.com/kairos-io/provider-kairos
- https://github.com/kairos-io/elemental-cli will be merged into kairos so no need to
- [ ] https://github.com/kairos-io/immucore
- [ ] https://github.com/kairos-io/kcrypt-challenger
- [ ] https://github.com/kairos-io/packages.kairos.io
- [ ] https://github.com/kairos-io/packages
- [ ] https://github.com/kairos-io/kcrypt
- [ ] https://github.com/kairos-io/provider-rke2
- [ ] https://github.com/kairos-io/provider-kubeadm
- [ ] https://github.com/kairos-io/kairos-sdk
- [ ] https://github.com/kairos-io/cluster-api-provider-kairos
- [ ] https://github.com/kairos-io/AuroraBoot
- [ ] https://github.com/kairos-io/community-bundles
- [ ] https://github.com/kairos-io/provider-k3s
- [ ] https://github.com/kairos-io/helm-charts
- [ ] https://github.com/kairos-io/provider-amt
- [ ] https://github.com/kairos-io/tpm-helpers
- [ ] https://github.com/kairos-io/entangle
- [ ] https://github.com/kairos-io/osbuilder
- more?

Itxaka:
Maybe we could have an action that runs our generic linters so it can be reused in all repos?
That would make everything more centralized and kept all repos in sync?

mudler:
A small sample of a composite action (that we already use) https://github.com/Luet-lab/luet-install-action

mauromorales:
ohh I like the idea, thanks for sharing guys

jimmykarily:
Created a new repo for the action here: https://github.com/kairos-io/linting-composite-action

mauromorales:
@jimmykarily there's already https://github.com/kairos-io/kairos-linter sorry that I didn't put it here in the ticket. We can use either or, no preference for me but I had already done some progress on the one I mention

jimmykarily:
Oops :man_facepalming: . I started with the same approach by using an Earthfile but then I couldn't find a way to tell Earthly "here is the Earthfile, here is the context dir" like you can do with docker. 
[...]

PR #1371, opened 2023-05-03
Consume reusable linting workflow

jimmykarily:
part of: https://github.com/kairos-io/kairos/issues/1234


jimmykarily:
The hadolint errors were there before but the earthly target has a mistake and didn't catch them. There are legitimate issues we should fix.

jimmykarily:
I'm not sure I agree with the hadolint suggestions:

```
./images/Dockerfile.opensuse-leap-arm-rpi:9 DL3037 warning: Specify version with `zypper install -y <package>=<version>`.
```
from [here](https://github.com/kairos-io/kairos/actions/runs/4873588267/jobs/8693376161?pr=1371). As far as I know, not every version is available for installation at any given image. It's the responsibility of the distribution to include package versions that "work". I understand builds should be reproducible but I'm not sure pinning the version of every package is doable.

jimmykarily:
I created a `.hadolint.yaml` file to assign them the severity "info" and then ignore them (not fail on them).

mauromorales (approved):
lgtm

PR #1422, opened 2023-05-17
Remove elemental-cli from kairos

Itxaka:

Fixes #


Itxaka:
Needs a repository bump to get the newer packages which include kairos-agent run stage and the new services using kairos-agent instead of elemental

Itxaka:
Needs a repository bump to get the newer packages which include kairos-agent run stage and the new services using kairos-agent instead of elemental

scripts/build_vm.sh
```diff
@@ -10,7 +10,7 @@ SSH_USER="${SSH_USER:-kairos}"
 SSHPASS="${SSH_PASS:-kairos}"
 export SSHPASS
 HD_SIZE="${HD_SIZE:-50000}"
-INSTALL_COMMAND="${INSTALL_COMMAND:-sudo /bin/sh -c 'elemental install /dev/sda && sync'}"
+INSTALL_COMMAND="${INSTALL_COMMAND:-sudo /bin/sh -c 'kairos-agent install /dev/sda && sync'}"
```
jimmykarily, line 13: Isn't it true that `kairos-agent install` and `kairos-agent manual-install` can be both used to achieve the same things (are interchangeable) ? If yes, maybe we should get rid of one of the 2.

jimmykarily: Nothing to do with this PR though, I just had that thought.

jimmykarily (dismissed):

scripts/build_vm.sh
Itxaka: Basically yes, but manual install accepts a config file while install does not and expects a config to be ready in the live CD or config dirs.

We should merge both and just add a flag to override the config file.

mauromorales (approved):

Itxaka:
moved to https://github.com/kairos-io/kairos/pull/1441

Issue #1395, opened 2023-05-09
support for NVIDIA Jetson AGX Orin (Developer Kit)


mudler:
**Is your feature request related to a problem? Please describe.**
A tracker issue for Nvidia jetson AGX Orin support

## Action items
- [x] #1440 
- [ ] QA: Try upgrades
- [ ] Create equivalent image in provider-kairos
- [ ] https://github.com/kairos-io/kairos-docs/pull/29
- [ ] See what we can do about resets https://github.com/kairos-io/kairos/issues/622

jimmykarily:
It works and we have docs. Let's close this.

Issue #622, opened 2023-01-09
Factory reset is not working as described on raspberrypi


c0ffee:
**Kairos version:**
NAME="kairos-opensuse-arm-rpi"
VERSION="v1.4.0-k3sv1.24.9+k3s1"
ID="kairos"
ID_LIKE="kairos-opensuse-arm-rpi"
VERSION_ID="v1.4.0-k3sv1.24.9+k3s1"
PRETTY_NAME="kairos-opensuse-arm-rpi v1.4.0-k3sv1.24.9+k3s1"
ANSI_COLOR="0;32"
BUG_REPORT_URL="https://github.com/kairos-io/kairos/issues/new/choose"
HOME_URL="https://github.com/kairos-io/provider-kairos"
IMAGE_REPO="quay.io/kairos/kairos-opensuse-arm-rpi"
IMAGE_LABEL="latest"
GITHUB_REPO="kairos-io/provider-kairos"
VARIANT="core"
FLAVOR="opensuse"

**CPU architecture, OS, and Version:**
Linux k8s-rpi3-1 5.14.21-150400.24.38-default #1 SMP PREEMPT_DYNAMIC Fri Dec 9 09:29:22 UTC 2022 (e9c5676) aarch64 aarch64 aarch64 GNU/Linux

**Describe the bug**
Factory reset from userspace is not working as described in the documentation. /oem is actually RO and is missing the grubenv file. Actually even the grub entry is missing at all too.

**To Reproduce**
grub2-editenv /oem/grubenv set next_entry=statereset

**Expected behavior**
Device does a factory reset :)

**Logs**

**Additional context**

mudler:
Thanks @c0ffee for filing an issue! 

That's a good point - by default ARM images don't come with an OEM partition so docs aren't correct in such case. We should move to LVM, or either document how to set an OEM partition externally.

c0ffee:
The reset option is missing in the grub menu too

What would be then the way to do a reset on Wich a device?

In general I think LVM would be the best approach to have the same lock and feel

The limitation is probably the partition table at the moment right?

mudler:
> The reset option is missing in the grub menu too
> 
> What would be then the way to do a reset on Wich a device?

Once a partition labeled "COS_OEM" is found it is automatically loaded, however that needs further tweaking. Indeed the best approach is to move over something more flexible, like LVM.

> The limitation is probably the partition table at the moment right?

Correct

jimmykarily:
[...]

PR #1440, opened 2023-05-22
:penguin: Add Jetson agx orin image

mudler:
**What this PR does / why we need it**: Adds initial support for Nvidia AGX Orin. Will follow up and iterate with the image hooked from the CI.

The difference with the other flavors is that instead of a flashable SD image, we prepare the partitions .img files that are ready to be flashed to the eeprom, and we just pack that as part of the releases so others don't have to re-create those images again.

Next:
- Try upgrades
- Create equivalent image in provider-kairos
- Add documentation on how to flash the board with the image files
- See what we can do about resets https://github.com/kairos-io/kairos/issues/622 

Related to #1395

### DIFF
--- a/.github/workflows/image-arm.yaml
+++ b/.github/workflows/image-arm.yaml
@@ -86,7 +86,7 @@ jobs:
                 insecure = true
                 http = true
           EOF
-          earthly -P +all-arm --IMAGE_NAME=kairos-$FLAVOR-latest.img --IMAGE=quay.io/kairos/core-$FLAVOR:latest --MODEL=$MODEL --FLAVOR=$FLAVOR
+          docker run --privileged -v $HOME/.earthly/config.yml:/etc/.earthly/config.yml -v /var/run/docker.sock:/var/run/docker.sock --rm --env EARTHLY_BUILD_ARGS -t -v "$(pwd)":/workspace -v earthly-tmp:/tmp/earthly:rw earthly/earthly:v0.7.5 --allow-privileged +all-arm --IMAGE_NAME=kairos-$FLAVOR-latest.img --IMAGE=quay.io/kairos/core-$FLAVOR:latest --MODEL=$MODEL --FLAVOR=$FLAVOR
       - name: Push  🔧
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
         env:
--- a/.github/workflows/image-arm.yaml
+++ b/.github/workflows/image-arm.yaml
@@ -65,12 +65,19 @@ jobs:
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@master
+      - name: Install earthly
+        uses: Luet-lab/luet-install-action@v1
+        with:
+          repository: quay.io/kairos/packages
+          packages: utils/earthly
       - name: Build  🔧
         env:
           FLAVOR: ${{ matrix.flavor }}
           MODEL: ${{ matrix.model }}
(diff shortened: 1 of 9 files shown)


overlay/files-nvidia/etc/cos/bootargs.cfg
```diff
@@ -0,0 +1,9 @@
+set kernel=/boot/vmlinuz
+
+if [ -n "$recoverylabel" ]; then
+    set kernelcmd="console=tty1 console=ttyTCU0,115200 root=live:LABEL=$recoverylabel net.ifnames=1 rd.live.dir=/ rd.live.squashimg=$img panic=5 rd.cos.oemtimeout=10"
```
Itxaka, line 4: we need a template for this file asap.

Earthfile
```diff
@@ -477,6 +491,39 @@ arm-image:
   END
   SAVE ARTIFACT /build/$IMAGE_NAME.sha256 img-sha256 AS LOCAL build/$IMAGE_NAME.sha256
 
+prepare-arm-image:
```
Itxaka, line 494: cant we reuse the existing arm image stuff? What is the difference from other arm images?

Earthfile
```diff
@@ -55,7 +55,17 @@ all-arm:
       BUILD --platform=linux/arm64 +trivy-scan --MODEL=rpi64
       BUILD --platform=linux/arm64 +grype-scan --MODEL=rpi64
   END
-  BUILD +arm-image --MODEL=rpi64
+  
+  IF [[ "$FLAVOR" = "ubuntu-20-lts-arm-nvidia-jetson-agx-orin" ]]
```
Itxaka, line 59: do we expect other flavors for this boards or ubuntu would be the only one?

.github/workflows/image-arm.yaml
```diff
@@ -65,12 +65,19 @@ jobs:
...
+        with:
+          repository: quay.io/kairos/packages
+          packages: utils/earthly
       - name: Build  🔧
         env:
           FLAVOR: ${{ matrix.flavor }}
           MODEL: ${{ matrix.model }}
         run: |
-          ./earthly.sh +all-arm --IMAGE_NAME=kairos-$FLAVOR-latest.img --IMAGE=quay.io/kairos/core-$FLAVOR:latest --MODEL=$MODEL --FLAVOR=$FLAVOR
+          earthly config "global.conversion_parallelism" "1"
```
Itxaka, line 78: comment with reason?

Itxaka (dismissed):

Earthfile
mudler: we don't actually create an SD card here, we just prepare the partitions

Earthfile
mudler: can't tell, probably this one will be the only one for a long time.. until nvidia releases another version or we have some reques

.github/workflows/image-arm.yaml
mudler: dunno if it works yet, debugging as it fails in transferring generated assets...

mauromorales (dismissed):
:clap:

mudler:
nice :D 
```
failed to register layer: ApplyLayer exit status 1 stdout:  stderr: write /usr/src/linux-headers-5.10.104-tegra-ubuntu20.04_aarch64/kernel-5.10/scripts/recordmcount.pl: no space left on device: exit status 1
```

mudler:
GREEEN!

PR #1441, opened 2023-05-22
Fully drop elemental from kairos

Itxaka:

Fixes #


overlay/files-alpine/etc/init.d/kairos-agent
```diff
@@ -2,15 +2,11 @@
...
-supervise_daemon_args="--stdout /var/log/kairos/agent.log --stderr /var/log/kairos/agent.log"
-pidfile="/run/kairos-agent.pid"
-respawn_delay=5
-set -o allexport
-if [ -f /etc/environment ]; then source /etc/environment; fi
-set +o allexport
\ No newline at end of file
+start() {
+    kairos-agent start
+    eend $?
```
Itxaka, line 11: not a type, I assure you :D

overlay/files/system/oem/00_datasource.yaml
```diff
@@ -5,24 +5,27 @@ stages:
       if: '[ ! -f /oem/userdata ]'
       name: "Pull data from provider"
       datasource:
-        providers: ["cdrom", "gcp", "openstack", "aws", "azure", "hetzner", "packet", "scaleway", "vultr", "digitalocean", "metaldata"]
+        providers: ["cdrom", "gcp", "openstack", "aws", "azure", "hetzner", "packet", "vultr", "digitalocean", "metaldata"]
```
Itxaka, line 8: scaleway has no timeout out of the box so it can take up to 60 seconds to timeout if there is network during initramfs. Patch upstream to fix this has been sent and merged, waiting for it to trickle down.

overlay/files/system/oem/02_agent.yaml
```diff
@@ -1,11 +1,11 @@
 name: "Start agent"
 stages:
   boot:
-    - if: '[ ! -f "/run/cos/recovery_mode" ]'
+    - if: '[ ! -f "/run/cos/recovery_mode" ] && [ -d "/usr/share/systemd" ]'
```
Itxaka, line 4: avoid running it on openrc systems. We need a better way to check this, probably via the `kairos-agent state get init` or something similar

overlay/files/system/oem/11_persistency.yaml
```diff
@@ -33,6 +33,7 @@ stages:
           /var/lib/wicked
           /var/lib/longhorn
           /var/lib/cni
+          /var/lib/dbus
```
Itxaka, line 36: alpine stores the machine-id here so it needs to be persistent

mauromorales (approved):
lgtm

overlay/files-alpine/etc/init.d/kairos-agent
```diff
@@ -2,15 +2,11 @@
...
 depend() {
   provide kairos-agent
+  after cos-setup-network
+  use net
 }
 
-supervisor=supervise-daemon
-name="kairos-agent"
-command="kairos-agent start"
-command_args="--restart"
```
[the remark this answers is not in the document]
Itxaka: Found out that leaving like that it would exit as crashed all the time.

Openrc is a bit weird, I could only get it to work as a one shot script that exits correctly like this

PR #1453, opened 2023-05-25
Yaml cleanup

Itxaka:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1458, opened 2023-05-28
:bug: Fix yip dependency

Itxaka:
Hardening depends on the accounting file to have run in order to have the admin group available.

With yip 1.x this runs in parallel so it could fail to find the group. Merge both files so the hardening occurs after the group has been created

Fixes #


Itxaka:
integrated with https://github.com/kairos-io/kairos/pull/1453/

PR #1461, opened 2023-05-29
:robot: disable parallelism on arm self-hosted release builds

mudler:

This is an attempt to fix release pipelines failures: https://github.com/kairos-io/kairos/actions/runs/5089099753/jobs/9149531881

### DIFF
--- a/.github/workflows/release-arm.yaml
+++ b/.github/workflows/release-arm.yaml
@@ -47,6 +47,11 @@ jobs:
           platforms: all
       - name: Install Cosign
         uses: sigstore/cosign-installer@main
+      - name: Install earthly
+        uses: Luet-lab/luet-install-action@v1
+        with:
+          repository: quay.io/kairos/packages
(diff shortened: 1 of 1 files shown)


Itxaka (commented):
Doesn't this slows it down? Does this means that runs everything serially instead of doing several things in parallel?

Doesn't earthly has a max 20 job queue for build kit? What is killing jobs exactly?

mudler:
> Doesn't this slows it down? Does this means that runs everything serially instead of doing several things in parallel?
> 
> Doesn't earthly has a max 20 job queue for build kit? What is killing jobs exactly?

Yes, it's one job only tho - I don't have the answers to all the questions you have here, but my guess it just exhaust ram when dumping 16 files in parallel

Itxaka (approved):

mudler:
Let's see... sadly this was affecting only releases, we will have to see..

PR #1462, opened 2023-05-30
:robot: run release-arm build on self-hosted

mudler:


mudler:
only testable on releases..

PR #1465, opened 2023-05-30
:warning: Add pure alpine flavor

Itxaka:
Provides a pure alpine flavor with alpine kernel and custom initrd

Requires:
 - https://github.com/kairos-io/packages/pull/253
 - https://github.com/kairos-io/packages/pull/254

Fixes #


mauromorales (approved):
:fire:

mudler (approved):
absolute love! <3

Itxaka:
fails due to kernel and initramfs being in different versions, new package with both bundled incoming.

mauromorales (approved):

Itxaka:
weird, it passes locally with no issues the autoinstall test, in fact is fast af (67 seconds)

Itxaka:
I havent tested this on rpi AT ALL. So no idea if this would work.

mauromorales:
gave it a spin on the RPi and got this issue

image

Itxaka:
> gave it a spin on the RPi and got this issue
> 
> image
> 

Thanks @mauromorales !

I think I screwed up because dracut should not be there 😭

Itxaka:
Time to get this ball running again....

Itxaka:
only thing missing for x86 is a missing cryptsetup in the initramfs, then it should pass the encryption tests as well.

For rpi, needs manual testing.

Itxaka:
This now should pass all tests, both other flavors and alpine pure, including encryption.

Only thing missing:
 - check yip configs to see why hey are not behaving (probably some checks not working as expected under initramfs)
 - check manual QA on rpi flavors

Itxaka:
This is what is autogenerated during initramfs I guess:

```
cat /etc/machine-id 
969551ab2fa82dc82a27e2f4d
```

And this is the proper format:

```
localhost:~$ dbus-uuidgen 
ae9ccad8d83b84d51bfb20fb00000092
```

Itxaka:
Needed to clean up /var/lib/dbus/machine-id as seems like its created during base image build!

Also, expansion of COS_PERSISTENT does not work, maybe missing a udev trigger somewhere? udev is run on the initramfs start and luks tests confirmt hat the /dev/disk/by-label is filled correctly, otherwise it cant work as expected... so no idea why that is happening, iut doesnt seem to find the partition correctly somehow.

Itxaka:
kairos-agent running twice on datasource -> Fixed by https://github.com/kairos-io/packages/pull/489
Expansion not working -> deps missing from initramfs, working on that currently...

jimmykarily:
> rpi boots as expected.
> 
> Minor issue with machine-id, seems to be generated during initramfs but not in the correct size somehow? So on rootfs switch dbus complains about it and cant start...
> 
> `UUID file '/etc/machine-id' should contain a hex string of length 32, not length 25, with no other text`

Maybe relevant? https://github.com/kairos-io/kairos/blob/750b0d1d00b029cea57e5913357438b80a85b2c3/Earthfile#L351

Itxaka:
> > rpi boots as expected.
> > Minor issue with machine-id, seems to be generated during initramfs but not in the correct size somehow? So on rootfs switch dbus complains about it and cant start...
> > `UUID file '/etc/machine-id' should contain a hex string of length 32, not length 25, with no other text`
> 
> Maybe relevant?
> 
> https://github.com/kairos-io/kairos/blob/750b0d1d00b029cea57e5913357438b80a85b2c3/Earthfile#L351

not reall, that would be good, but something is creating a /var/lib/dbus/machine-id which is not correct...no idea what, maybe udev, maybe busybox with its services...

Anyway, its fixed now :)

framework-profile.yaml
```diff
@@ -11,119 +11,96 @@ flavors:
...
+    - alpine-pure
+  alpine-arm-rpi:
     - common-packages
     - kairos-toolchain
-    - ubuntu-kernel
     - openrc
+    - alpine-pure-rpi
 # See https://github.com/kairos-io/packages/pull/67 for rationale
-dracut-network-legacy:
+alpine-pure:
```
mauromorales, line 90: 🤣 why not just alpine? without the `pure` don't want to start here a war between pure and impure distros at some point hahaha

mauromorales (approved):
Fantastic work @Itxaka 👏 👏

framework-profile.yaml
Itxaka: should be alpine-kernel xDD

alpine is already used up there for the distro name, this was a way to differ between the alpine-{ubuntu,opensuse} and the newer one, but that is now dropped xDDD

will change it to alpine-kernel or alpine-common or something more straightforward.

Itxaka:
only thing missing is provider-kairos with p2p enabled takes over the terminals witht he output, blocking the loging console. Something minor IMHO.

Itxaka:
Missing things:

 - (all) run edgevpn and all that
 - (rpi) fix clock on initramfs (not set, defaults to 1970) - Not urget, need to decide between using swclock or ntp if network.

PR #1475, opened 2023-06-02
:seedling: Drop tmp.mount services

Itxaka:
Should be not needed as immucore is mounting the tmp dir and systemd will make sure the mount is on once switched to the new root after initramfs

Fixes #


mauromorales (approved):
lgtm

PR #1476, opened 2023-06-02
:sparkles: Remove OverlayFS patch

mauromorales:
A flag can be passed, so the kernel enables the OverlayFS module. 

> rd.live.overlay.overlayfs=1
           Enables the use of the OverlayFS kernel module, if available,
           to provide a copy-on-write union directory for the root
           filesystem. OverlayFS overlays are directories of the files
           that have changed on the read-only base (lower) filesystem.
           The root filesystem is provided through a special overlay
           type mount that merges the lower and upper directories. If an
           OverlayFS upper directory is not present on the boot device,
           a tmpfs directory will be created at /run/overlayfs to
           provide temporary storage. Persistent storage can be provided
           on vfat or msdos formatted devices by supplying the OverlayFS
           upper directory within an embedded filesystem that supports
           the creation of trusted.* extended attributes and provides a
           valid d_type in readdir responses, such as with ext4 and xfs.
           On non-vfat-formatted devices, a persistent OverlayFS overlay
           can extend the available root filesystem storage up to the
           capacity of the LiveOS disk device.
>         If a persistent overlay is detected at the standard LiveOS
           path, the overlay & overlay type detected, whether OverlayFS
           or Device-mapper, will be used.
>         The rd.live.overlay.readonly option, which allows a
           persistent overlayfs to be mounted read-only through a higher
           level transient overlay directory, has been implemented
           through the multiple lower layers feature of OverlayFS.

I tested Tumbleweed and Debian, which are already using the patch, and Ubuntu 23.04, which will need it. Plus also tested Fedora to make sure it doesn't break anything in one flavor that is not relying on the patch.


Itxaka (approved):
Is it possible to bring the overlay module directly with the modules.conf.d dir instead?

I guess this would also affect the grub-config package due to it being used with recovery squashes but as we currently dont use it anywhere....no problem.

mauromorales:
> Is it possible to bring the overlay module directly with the modules.conf.d dir instead?

Good point, let me try it out

mauromorales:
could't find a way to do it, so will merge like this and try with a different autoloaded module and get back to this one if I find something

PR #1492, opened 2023-06-08
:robot: Build and publish ubuntu arm generic

mauromorales:
Currently, we only release the opensuse-leap, but we can already build ubuntu and maybe others too? Not sure if we want to provide all flavors also as generic, I know this adds to the build time, but my fear would be that if we never test them, they might be in a broken state and we will not notice. So let me know what you think

### DIFF
--- a/.github/workflows/image-arm.yaml
+++ b/.github/workflows/image-arm.yaml
@@ -171,13 +171,12 @@ jobs:
           if-no-files-found: error
   image_and_iso_arm64_generic:
     runs-on: ubuntu-latest
-    needs:
-      - get-matrix
     strategy:
       fail-fast: false
       matrix:
         flavor:
           - "opensuse-leap"
+          - "ubuntu"
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


Itxaka (dismissed):

jimmykarily:
Building them doesn't mean we test them, unless they are used in some test scenarios. Is there really a benefit in building more of them?

mauromorales:
yeah indeed, I meant to say testing the build at least ... but it's a good question, not sure who would be consuming these so no idea if there's little or much value from it

mauromorales:
@mudler what is the goal for the generic images? Should we have ready to download images from all flavors, or just having the existing one is ok?

jimmykarily:
We discussed this in planning. This is the outcome:

We don't have a plan on how to automate the arm testing (because emulating arm on non-arm CPUs with qemu is very slow). For this reason, it doesn't make sense to built too many flavors for this. As soon as we automate testing, we can add more.
For now, testing will be manual and one flavor should be sufficient.

mauromorales:
Makes sense closing

PR #1496, opened 2023-06-09
:robot: Use builjet arm workers for arm jobs

Itxaka:
Runs arm workers on buildjet with native arm64 workers.
Instead of compressing img files, we transform them into qcow2 which are compressed by default, speeding up the build up to 30 minutes.

Fixes #

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -614,8 +614,8 @@ arm-image:
     RUN /build-arm-image.sh --use-lvm --model $MODEL --directory "/build/image" /build/$IMAGE_NAME
   END
   IF [ "$COMPRESS_IMG" = "true" ]
-      RUN xz -v /build/$IMAGE_NAME
-      SAVE ARTIFACT /build/$IMAGE_NAME.xz img AS LOCAL build/$IMAGE_NAME.xz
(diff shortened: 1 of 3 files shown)


Earthfile
```diff
@@ -614,8 +614,8 @@ arm-image:
     RUN /build-arm-image.sh --use-lvm --model $MODEL --directory "/build/image" /build/$IMAGE_NAME
   END
   IF [ "$COMPRESS_IMG" = "true" ]
-      RUN xz -v /build/$IMAGE_NAME
-      SAVE ARTIFACT /build/$IMAGE_NAME.xz img AS LOCAL build/$IMAGE_NAME.xz
```
mauromorales, line 618: is this change intended? If yes, maybe we can have a different flag to decide whether we want a qcow2 or img

Itxaka: done!

mauromorales (dismissed):
noiiiice

mauromorales (approved):

jimmykarily:
This PR only, cost 5 dollars! This doesn't scale :D Let's try to optimize on our workers instead.

Itxaka:
Will resend the patch for not compressing arm images on PRs as separated.

Just 1 PR with 4 or 5 commits already set the usage of buildjet to an estimated 5$ according to the buildjet dashboard. I dont think this scales properly LMAO.

PR #1500, opened 2023-06-12
:bug: Alpine: Mount bpf system once we boot

Itxaka:
initramfs stage runs under a chroot.
If we do the mount in initramfs we are troublemaking the chroot mounts that the stage uses to run the config files. Nothing should be mounted in the binded paths during that stage to avoid issues as after the stage is finished, the chroot is removed fully, so those mounts would get lost.

Fixes k3s issues on provider-kairos
Fixes cgroups for alpine

### DIFF
--- a/overlay/files/system/oem/00_rootfs.yaml
+++ b/overlay/files/system/oem/00_rootfs.yaml
@@ -13,6 +13,10 @@ stages:
         providers: ["aws", "gcp", "openstack", "cdrom"]
         path: "/oem"
   rootfs:
+    - name: "Mount BPF on Alpine systems"
+      if: '[ ! -e "/sbin/systemctl" ] && [ ! -e "/usr/bin/systemctl" ] && [ ! -e "/usr/sbin/systemctl" ] && [ ! -e "/usr/bin/systemctl" ]'
+      commands:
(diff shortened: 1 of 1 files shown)


overlay/files/system/oem/00_rootfs.yaml
```diff
@@ -13,6 +13,10 @@ stages:
         providers: ["aws", "gcp", "openstack", "cdrom"]
         path: "/oem"
   rootfs:
+    - name: "Mount BPF on Alpine systems"
```
mauromorales, line 16: maybe "Mount BPF on OpenRC system" would be a better description here?

Itxaka: I want to talk with @mudler about this, because I dont get why its there. AFAIK the bpf tools in the alpine ecosystem are non-existant and this is possible by using the opensuse/ubuntu kernels. 

Not sure if this will be there for a long time :D

But yeah, maybe openrc systems would be nicer.

mauromorales: gotcha, not a srtong request from my side, if we can get rid of it even better I guess

PR #1514, opened 2023-06-14
:penguin: do not install unattended upgrades

mudler:
We do remove it in case it is present in the base image already, but there is no reason to install it in the first place if isn't there.


mauromorales (approved):
lgtm

mauromorales (approved):

PR #1531, opened 2023-06-20
:bug: Fix restoring machine-id on reboot

Itxaka:
Was missing a check for openrc on the restoring machine-id so it was restored twice and the second was picking up a new machine-id generated by dbus

This worked on openrc systems as the paths to save and restore were correct and the save overwrote the generation but on systemd it was a newly generated machine-id that we are not saving thus it never matched.

Alos drops a duplicated metadata call

Fixes #


mauromorales (dismissed):
lgtm

mauromorales (approved):

jimmykarily (approved):

PR #1537, opened 2023-06-21
POC: Use prebuilt image for most of the targets

Itxaka:
This uses a prebuilt image with all the deps already on it.
The image is tracked against the golang:VERSION dokcer image as base image
Renovate is used to autobump it when the upstream image changes
Quay auto builds both master and any tags added

This solves:

 - reuse of the same image everywhere instead of having 5 different images
 - do not waste time installing deps on each target that needs it
 - for some targets it speeds up running them as the image was already downloaded (like version already gets it so we got the image in cache at that point)
 - easily see if something is broken on the base image (just happened during the lat release, image changed, release was broken in the middle)
 - no more api limits reached, quay has "unlimited" pulls
 - easy to update, at our own cadence, easy to test that it works
 - one single place for the image configuration, also renovate bumps it if we bump it, easy to change to dockerhub/gcr or whatever.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1,5 +1,5 @@
 VERSION 0.6
-FROM alpine
+
 ARG VARIANT=core # core, lite, framework
 ARG FLAVOR=opensuse-leap
 ARG BASE_URL=quay.io/kairos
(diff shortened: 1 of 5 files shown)


Itxaka:
oh yeah this also moves the suse image to be pulled from registry.opensuse.org/ which should not have api pull limits

Itxaka:
> In some places we call "BASE_IMAGE" the image that the flavor it based on. Now we use the same name for the image on which we run build targets. Maybe we should use different names for each (e.g. "BUILD_BASE_IMAGE"?)

thats `+base-image` which is a target, but yeah, can be changed to BUILD_BASE_IMAGE or whatever, no issues there

PR #1540, opened 2023-06-21
v2.2.1 Test PR

mauromorales:
supersedes https://github.com/kairos-io/kairos/pull/1535


Itxaka (approved):
Looking good. Upgrade-latest-with-cli for alpine will fail as this branch doesn't have the fix, but I can live with it, rest should pass with no issues

PR #1542, opened 2023-06-21
:sparkles: Refactor CI target into ALL target

mauromorales:
just a small refactoring to try to reduce a bit all our targets in Earthly


Itxaka (approved):
IMO it's clearer to have different targets for different ... Well, targets. It's easier to use than having extra ARGS. But got nothing against it.

mauromorales:
oki no problem, I was looking at it more from like a code perspective, cuz atm we maintain two targets (in this case) and to reduce likelyhood of forgeting to change one without updating the other but maybe that can be achieved via a common target. Will try that and see how it feels

PR #1559, opened 2023-06-27
:robot: Test PRs only on one flavor

### LINKED ISSUE #1556: Build only one flavor on PRs

jimmykarily:
Refactor all jobs to reusable workflows so that we can construct a PR and a master pipeline my mixing and matching jobs as desired.

Fixes #1556


mudler (dismissed):

jimmykarily:
An alternative would be to run the whole workflow (build + test) for only one flavor on PRs and if that passes, we run some more. I'm only concerned that we have some code switched to identify openrc based flavors and we do things differently then (check `isFlavor` method in tests).

Otoh, maybe that's not creating issues as often as to justify delaying all PRs.

.github/workflows/image-arm-pr.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+concurrency:
+  group: ci-arm-${{ github.head_ref || github.ref }}-${{ github.repository }}
+  cancel-in-progress: true
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  docker:
+    uses: kairos-io/kairos/.github/workflows/reusable-docker-arm-build.yaml@1556-pr-one-flavor
```
jimmykarily, line 16: I'll change these to `master` right before merging. Right now if I do, it won't find the definition of that workflow.

jimmykarily:
I'll remove the mandatory checks for merging that are not going to run anymore

jimmykarily:
Switching the zfs test to run on opensuse-leap broke it because zpool is not installed: https://github.com/kairos-io/kairos/blob/0294acdde96aadbbc4268f30127bb5b28cc4bd81/tests/zfs_test.go#L41

Obviously on ubuntu (where this test used to run), it's installed.

jimmykarily:
Strange, I tried this: https://github.com/kairos-io/kairos/commit/8b5d877e962fdc9bdba51c1cd1010be40c0e56ed#diff-52ac78a83f922b3f0116bb67b1aa8100a7620a3be6d63b00ef53cb1495e4eb19R71 on a docker container and it seems to provide `zpool`.

jimmykarily:
It was removed immediately after: https://github.com/kairos-io/kairos/commit/c22b38a3d76ec3d433b0bc52d75d79e3c385389b#diff-01f2970c67918dfa463783e91789e617d04d30b4555f9b2e1c156e91435eb33fL84

jimmykarily:
I'll try switching to ubuntu for PRs

.github/workflows/image-arm-pr.yaml
```diff
@@ -0,0 +1,25 @@
...
+    with:
+      flavor: opensuse-leap-arm-rpi
+      model: rpi64
+      worker: ubuntu-latest
+
+  image_and_iso_arm64_generic:
+    uses: kairos-io/kairos/.github/workflows/reusable-image-and-iso-arm-generic.yaml@1556-pr-one-flavor
```
mudler, line 23: mmmh maybe we need a composite action?

mudler (approved):

.github/workflows/image-arm-pr.yaml
jimmykarily: [I see no benefit in using a composite action](https://dev.to/n3wt0n/composite-actions-vs-reusable-workflows-what-is-the-difference-github-actions-11kd). Re-usable workflows are a lot more flexible as far as I can tell.

mauromorales:
Reusable workflows merged in #1585

Issue #1548, opened 2023-06-22
kairos-agent upgrade on custom images fails with missing /etc/os-release keys


venkatnsrinivasan:
We built a custom image for RHEL 8 with the steps outlined here 
https://kairos.io/docs/reference/build-from-scratch/

The resulting image has an /etc/os-release as below.

```
NAME="Red Hat Enterprise Linux"
VERSION="8.7 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.7 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/8/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.7
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"
```

This image is able to be deployed successfully but when trying and upgrade via kairos-agent upgrade in Kairos 2.2.0 the upgrade fails with
`Cant find IMAGE_REPO key under /etc/os-release
`

mauromorales:
I think the problem is that the following step:

https://github.com/kairos-io/kairos/blob/9b3887a6b8e95d9eb0959f3ffbdc05723c4e19c5/Earthfile#L111

is run when building the base images. So when using BYOI it is missing, or at least I couldn't find it on AuroraBuild/osbuilder

mauromorales:
I think either osbuilder offer a mechanism how to do this (and then it can be used on the main factory as well), or it should be part of the documentation for BYOI. Having said that, it will be the job of the user doing BYOI to define all or most of the values that go there, e.g. keep their own versioning because at the end of the day, they can decide on the cadence of release and it doesn't fit that they keep the same versioning of our releases. Alternatively we could do something like what we do with the k3s version, and when building the BYOI the final version would be something along the lines of KAIROS_VERSION+OWN_VERSION

mauromorales:
[...]

PR #1561, opened 2023-06-28
Use osbuilder tools to update os-release

mauromorales:
Dog fooding what was added on the docs, plus makes our factory process less dependant on Earthly and more into Osbuilder like we want to do for BYOI

relates to #1548


mudler (approved):

mauromorales:
not sure this is possible because of the multiple platform switch (amd -> arm -> amd) will close since it's not a high prio

PR #1582, opened 2023-07-05
:robot: Clean up space on each test job

Itxaka:
Try to clean up as much space as possible before each CI job to avoid running out of space

Fixes #


mauromorales (approved):
doesn't need to be for this PR but I wonder if we can package this on a composite or reusable action, to avoid having to repeat all that code

PR #1585, opened 2023-07-05
:robot: Introduce reusable jobs

mauromorales:
First part of #1556 which introduces the reusable jobs. Doing the split because it was much easier for me to do the diffying of the files and so that I can merge this first and the actuall migration second


.github/workflows/reusable-upgrade-latest-test.yaml
```diff
@@ -0,0 +1,92 @@
...
+      - run: |
+          git fetch --prune --unshallow
+      - name: Download latest release
+        uses: actions/download-artifact@v3
+        with:
+          name: latest-release.zip
+      - name: Download iso
+        uses: actions/download-artifact@v3
+        with:
+          name: kairos-${{ inputs.flavor }}.iso.zip
```
mauromorales, line 44: do we need to unpack this? see line 83 ... before we didn't need to because it was all in the same job

mauromorales:
Theres's nothing that runs differently because of this change, so I will just merge

Issue #1557, opened 2023-06-27
Merge provider-kairos CI with kairos


jimmykarily:
We have lots of duplication between kairos and provider-kairos repositories (e.g. Earthly targets, tests etc). The provider-kairos repository should just be the provider-kairos source code with only a binary as the artifact.

The `kairos-io/kairos` repository can build provider-kairos images (and core images like it does now) by integrating the provider-kairos binary in the images.

This will allow us to find breaking changes earlier (before we even release kairos core) and also reduce how many CI resources we need.

PR #1592, opened 2023-07-07
Build provider in kairos and bring tests

Itxaka:

Brings the testing of provider-kairos into kairos CI.

This only deals with x86 testing as there is no other tests rather than building an iso/image for arm64.

Fixes https://github.com/kairos-io/kairos/issues/1557


Earthfile
```diff
@@ -338,6 +346,27 @@ base-image:
...
     # END
 
+    # Build with provider-kairos
+    # TODO: Allow installing from luet packages
+    ARG PROVIDER_KAIROS
+    ARG PROVIDER_KAIROS_BRANCH=main
+    IF [[ "$PROVIDER_KAIROS" != "" ]]
+        DO github.com/kairos-io/provider-kairos:main+PROVIDER_INSTALL --FLAVOR=$FLAVOR
+        # Redo os-release with override settings to point to provider-kairos stuff
+        # in earthly 0.7 we will be able to just override VARIANT here and just run the OSRELEASE once
```
mauromorales, line 356: could we just upgrade to this verstion already?

tests/assets/kubernetes_logs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
```
mauromorales, line 2: a short description could be useful here

tests/assets/kubernetes_logs.sh
```diff
@@ -0,0 +1,10 @@
...
+for n in $(k3s kubectl  get namespace -A | tr -s ' ' | cut -f1 -d' ' | tail -n +2); do
```
mauromorales, line 3:
```suggestion
for n in $(k3s kubectl get namespace -A | tr -s ' ' | cut -f1 -d' ' | tail -n +2); do
```

mauromorales (dismissed):
sweet 💪

Earthfile
Itxaka: Not easily as 0.7 introduces the concept of global vars and that could potentially screw the current earthfile. 

It would need to come in a different PR to make sure we don't break any other stuff

tests/assets/kubernetes_logs.sh
Itxaka: Good point, this was just a copy from provider so I didn't think of touching this files

Itxaka:
all jobs passed under https://github.com/kairos-io/kairos/actions/runs/5528778626/jobs/10086510302 moving them now to master only

.github/workflows/image.yaml
```diff
@@ -179,6 +179,86 @@ jobs:
...
+          head -n 30 <<< "${pkgs}"
+          echo
+          sudo rm -rfv build || true
+          df -h
+      - uses: actions/checkout@v3
+      - run: |
+          git fetch --prune --unshallow
+      - name: setup-docker
+        uses: docker-practice/actions-setup-docker@master
+      - name: Release space from worker
```
mauromorales, line 234: not a big deal, but should we move together with the other Release space?

mauromorales (dismissed):
Just two minor things, other than that lgtm

.github/workflows/image.yaml
Itxaka: woops, didnt even noticed

PR #1616, opened 2023-07-18
Test

Itxaka:

Fixes #

### DIFF
--- a/.github/workflows/reusable-build-flavor.yaml
+++ b/.github/workflows/reusable-build-flavor.yaml
@@ -73,6 +73,7 @@ jobs:
           FLAVOR: ${{ inputs.flavor }}
           IMAGE: quay.io/kairos/core-${{ inputs.flavor }}:latest
         run: |
+          exit 1
           earthly +ci --SECURITY_SCANS=false --IMAGE=$IMAGE --FLAVOR=$FLAVOR
           sudo mv build/* .
           sudo rm -rf build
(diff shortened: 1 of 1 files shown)


mauromorales (commented):
I think the best would be to add a final job that depends on all of these jobs. I haven't tried it but I think you could ensure it always runs by using `if: always()` and then in the part that does the actual notification you can validate if any of the previous jobs failed by doing `if: failure()`

https://docs.github.com/en/actions/learn-github-actions/expressions#always

https://docs.github.com/en/actions/learn-github-actions/expressions#failure

Itxaka:
> I think the best would be to add a final job that depends on all of these jobs. I haven't tried it but I think you could ensure it always runs by using `if: always()` and then in the part that does the actual notification you can validate if any of the previous jobs failed by doing `if: failure()`
> 
> [docs.github.com/en/actions/learn-github-actions/expressions#always](https://docs.github.com/en/actions/learn-github-actions/expressions#always)
> 
> [docs.github.com/en/actions/learn-github-actions/expressions#failure](https://docs.github.com/en/actions/learn-github-actions/expressions#failure)

But im wondering, as if you `depend` on a previous job, and that job fails, your dependant job would not even trigger no? Even if you have a failure() check, that is only checked one the job is started, and it wont start.

Let me create a test repo with this to see if it actually works like that.

Itxaka:
> > I think the best would be to add a final job that depends on all of these jobs. I haven't tried it but I think you could ensure it always runs by using `if: always()` and then in the part that does the actual notification you can validate if any of the previous jobs failed by doing `if: failure()`
> > [docs.github.com/en/actions/learn-github-actions/expressions#always](https://docs.github.com/en/actions/learn-github-actions/expressions#always)
> > [docs.github.com/en/actions/learn-github-actions/expressions#failure](https://docs.github.com/en/actions/learn-github-actions/expressions#failure)
> 
> But im wondering, as if you `depend` on a previous job, and that job fails, your dependant job would not even trigger no? Even if you have a failure() check, that is only checked one the job is started, and it wont start.
> 
> Let me create a test repo with this to see if it actually works like that.

shiiiiet, it seems to actually work!

https://github.com/Itxaka/test-ci-depends/actions/runs/5597106674

Itxaka:
> fantastic!

ah yes, the problem I guess its how to depend on everything

Like if i depend on core and qemu-acceptance test and core fails, will it trigger? Then that means we still can not notify when building is broken as it wont trigger the notify job as the dependent jobs havent triggered


Gonna test that as well.

Itxaka:
> > fantastic!
> 
> ah yes, the problem I guess its how to depend on everything
> 
> Like if i depend on core and qemu-acceptance test and core fails, will it trigger? Then that means we still can not notify when building is broken as it wont trigger the notify job as the dependent jobs havent triggered
> 
> Gonna test that as well.

wow, this actually works: https://github.com/Itxaka/test-ci-depends/actions/runs/5597190257

Im sure it has changed as I tried the same thing a couple of years ago and it was a no-go. NICE

.github/workflows/reusable-build-flavor.yaml
```diff
@@ -73,6 +73,7 @@ jobs:
           FLAVOR: ${{ inputs.flavor }}
           IMAGE: quay.io/kairos/core-${{ inputs.flavor }}:latest
         run: |
+          exit 1
```
mauromorales, line 76: Notification only runs on master, so I guess we can drop this ... otherwise if you want to just test it, you can copy the notification into image-pr.yaml without the branch checks

mauromorales (dismissed):
Very cool! It's only missing to alwasy put a random gif :speak_no_evil:

.github/workflows/reusable-build-flavor.yaml
Itxaka: ffffuuuuuuuuuuuu

Itxaka: fixed, good catch yikes

Itxaka:
> Very cool! It's only missing to alwasy put a random gif speak_no_evil

yeah... not too sure about "random" gifs. I dont want it to post gore or something racist into the chat....internet is evil!

mauromorales (dismissed):
lgtm

Itxaka:
Fixed the lint and pushing as it wont affect PR tests!

PR #1619, opened 2023-07-19
🤖 Add simple UKI test

Itxaka:

Fixes #

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -118,6 +118,7 @@ func startVM() (context.Context, VM) {
 
 	stateDir, err := os.MkdirTemp("", "")
 	Expect(err).ToNot(HaveOccurred())
+	fmt.Printf("State dir: %s\n", stateDir)
 
 	if os.Getenv("EMULATE_TPM") != "" {
 		emulateTPM(stateDir)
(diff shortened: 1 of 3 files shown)


tests/uki_test.go
```diff
@@ -0,0 +1,89 @@
...
+			out, err := vm.Sudo("kairos-agent state")
+			Expect(err).ToNot(HaveOccurred())
+			// TODO: make agetn report uki_mode or something?
+			Expect(out).To(ContainSubstring("boot: unknown"))
+			currentVersion, err := vm.Sudo(getVersionCmd)
+			Expect(err).ToNot(HaveOccurred(), currentVersion)
+
+			stateAssertVM(vm, "kairos.version", strings.ReplaceAll(strings.ReplaceAll(currentVersion, "\r", ""), "\n", ""))
+			stateContains(vm, "system.os.name", "alpine", "opensuse", "ubuntu", "debian", "fedora")
+			stateContains(vm, "kairos.flavor", "alpine", "opensuse", "ubuntu", "debian", "fedora")
```
mauromorales, line 86: how come we need to have here all the different flavors?

Itxaka: Just a generic test that check that is indeed on of the flavors of the lost so if we change or add extra flavors for the test in the future it doesn't need a test change

mauromorales: gotcha, I guess ideally we should pass it via an env var or pick it from the uki file name or something like that, but can be improved later on

mauromorales (approved):
lgtm

tests/uki_test.go
Itxaka: yeah, this was mainly copied fromt he existing acceptance test. Its kind of generic and could be more specific.

tests/tests_suite_test.go
```diff
@@ -186,6 +187,28 @@ func startVM() (context.Context, VM) {
...
+			if FW != "" {
+				m.Args = append(m.Args,
+					"-bios", FW)
+			}
+
+			return nil
+		},
+		// UKI boot
+		func(m *types.MachineConfig) error {
+			drive := os.Getenv("UKI_DRIVE")
```
jimmykarily, line 202: Sorry for the late comment. I think the test below can easily turn to false positive if one omits these 2 environment variables. A quick solution would be to check in the test that these 2 environment variables are set (like [we do for `CONTAINER_IMAGE` in the upgrade test](https://github.com/kairos-io/kairos/blob/224d7c0045ddd5c1d648e7f379359f175aaabbd4/tests/upgrade_cli_test.go#L16-L21))

PR #1643, opened 2023-07-25
:bug: Stop kairos services in proper time

Itxaka:
Marks the services of kairos and webui with a max stop time of 10 seconds to not block when shutting down or restarting as they get the default 1m30s and that slows the whole system.

10s should be more than enough for the services to stop.

Fixes #


mauromorales (approved):
lgtm

jimmykarily (approved):

Issue #1609, opened 2023-07-17
:warning:  Build provider-kairos in this repository


mudler:
Continuation of : https://github.com/kairos-io/kairos/issues/1557

We have moved the provider-kairos tests to kairos, this card is about moving the release process into kairos

## Action items

- [x] Have provider-kairos as a package
- [x] Build k3s as a luet package (https://github.com/kairos-io/packages/pull/335)
- [ ] Move the release jobs from provider kairos to kairos
- [ ] Merge the build process such as the core image is built, and right after we build the provider image to save CI resources
- [ ] Update the release template, documentation
- [ ] QA

Open points to check:
  - Adapt provider-kairos binary to pick up release list from the new location
  - Check if changes to Auroraboot are needed

jimmykarily:
Provider kairos has a luet package: https://github.com/kairos-io/packages/tree/main/packages/bundles/provider-kairos
and so does k3s: https://github.com/kairos-io/packages/tree/main/packages/k3s

but the k3s one doesn't do all the things described on Itxaka's comment above (e.g. it doesn't create [the services files](https://github.com/k3s-io/k3s/blob/be442433538a39c0a568516fd39a06e364d5d075/install.sh#L891)), it's [in TODO](https://github.com/kairos-io/packages/blob/fddf0ce68bf92c9dab71991ceb6d86e2c355e991/packages/k3s/build.yaml#L24))

We can try to finish the k3s package and use it but I wonder if it would be better to actually call the the k3s installer script from within the luet package instead. Are we breaking some convention if a luet package makes calls to the internet? Is it supposed to be self-contained? I'm thinking that if we don't do this, we would need to always follow the logic of the installer script and make sure we replicate the same result in the luet package.

jimmykarily:
The existing package already downloads k3s: https://github.com/kairos-io/packages/blob/fddf0ce68bf92c9dab71991ceb6d86e2c355e991/packages/k3s/build.yaml#L9

We could as well do this in the package: https://github.com/kairos-io/provider-kairos/blob/9f56a5578bfee5d0d0018d4eba7686a19939f7fb/Earthfile#L168-L172

PR #1645, opened 2023-07-25
1609 release standard images

jimmykarily:
Part of #1609


.github/kubernetes-versions.json
```diff
@@ -0,0 +1,5 @@
+[
+    "v1.25.11+k3s1",
+    "v1.26.6+k3s1",
+    "v1.27.3+k3s1"
```
jimmykarily, line 4: These should match what the repository provides: https://github.com/kairos-io/packages/blob/main/packages/k8s/k3s/collection.yaml

I wonder if we should parse the `collection.yaml` from the remote location directly :thinking: . This way, when we want to change the versions we support, we will only have to change the versions built in the packages repository and this pipeline will pick it up automatically.

Itxaka: I would say yes, otherwise we will need to do manual stuff no?

jimmykarily:
This PR needs to also migrate the release-arm workflow from provider-kairos repo

.github/kubernetes-versions.json
jimmykarily: Done (I think, we will know when the pipeline runs)

jimmykarily: https://github.com/kairos-io/kairos/pull/1645/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR40-R42

jimmykarily:
Closing this draft to save some CI resources. I still have quite some things to push.

Issue #1620, opened 2023-07-19
Add common linux utilities into the base packages


saamalik:
**Is your feature request related to a problem? Please describe.**
When troubleshooting issues, e.g: why machine isn't routable - it's frustrating not having common utilities for debugging.

**Describe the solution you'd like**
Please add `iputils-ping` and `less` to the list of packages installed (at least for Ubuntu) - but ideally all variants for consistency.

**Describe alternatives you've considered**
Use a customized image that adds these packages. For most packages, I don't think we should keep adding them to the base image (bloatware) - but common utilities used for troubleshooting like `ping` and `less` are really helpful for vanilla installs (which I think is how most users would start)

## flavors

- [x] opensuse-leap
- [x] opensuse-tumbleweed
- [ ] fedora (missing less) 
- [ ] debian (missing both)
- [x] rockylinux
- [x] almalinux
- [x] alpine
- [ ] ubuntu (missing both)

PR #1656, opened 2023-07-27
:sparkles: Refactor Dockerfiles for Ubuntu based flavors

mauromorales:
I was looking into adding the different packages requested in #1620 and noticed that we again have differences in our Dockerfiles, so I decided to do a POC to evaluate a better way to maintain them, and here's my proposal, which is what I already mentioned in https://github.com/kairos-io/kairos/discussions/1400

The major change is to go from individual Dockerfiles per flavor within the same distro, to having a single one for the distro using stage-builds. The reason why I think this makes sense is that most of the content of these files is supposed to be shared across flavors based on a single distribution, and only specific packages or scripts need to run for each different flavor within that distro.


images/Dockerfile.ubuntu
```diff
@@ -80,6 +75,61 @@ RUN apt-get update \
...
+    lldpd \
+    patch \
+    policykit-1 \
+    snmpd \
+    ubuntu-advantage-tools \
+    && apt-get remove -y unattended-upgrades && apt-get clean && rm -rf /var/lib/apt/lists/*
+# workaround https://github.com/kairos-io/kairos/issues/949
+# TODO: backport patch into packages
+COPY images/dracut-broken-iscsi-ubuntu-20.patch /
+RUN cd /usr/lib/dracut/modules.d/95iscsi && patch < /dracut-broken-iscsi-ubuntu-20.patch && rm -rf /dracut-broken-iscsi-ubuntu-20.patch
```
mauromorales, line 111: this part here is actually not required any more, that's how I spotted the inconsistency. Will remove after review, just want to use this to exemplify the benefit of having a multi-stage setup

images/Dockerfile.ubuntu
```diff
@@ -80,6 +75,61 @@ RUN apt-get update \
...
+    lldpd \
+    patch \
+    policykit-1 \
+    snmpd \
+    ubuntu-advantage-tools \
+    && apt-get remove -y unattended-upgrades && apt-get clean && rm -rf /var/lib/apt/lists/*
+# workaround https://github.com/kairos-io/kairos/issues/949
+# TODO: backport patch into packages
+# COPY images/dracut-broken-iscsi-ubuntu-20.patch /
+# RUN cd /usr/lib/dracut/modules.d/95iscsi && patch < /dracut-broken-iscsi-ubuntu-20.patch && rm -rf /dracut-broken-iscsi-ubuntu-20.patch
```
mauromorales, line 111: This part is not commented on the Ubuntu-20-lts dockerfile, that's how I spotted the inconsistency. Will remove after review, just want to use this to exemplify the benefit of having a multi-stage setup

images/Dockerfile.ubuntu
```diff
@@ -80,6 +75,61 @@ RUN apt-get update \
     zstd \
     && apt-get remove -y unattended-upgrades && apt-get clean && rm -rf /var/lib/apt/lists/*
 
+# Packages and scripts only for specific images
```
mauromorales, line 78: my guess is that most of the files that are not consistent accross the Dockerfiles on master, are not because they are specifically required but because we failed to keep them in sync. But even with those differences, you can see how this section is actually pretty small and makes it easy to eval the differences between distro versions.

mauromorales:
Merging since it was agreed during daily. Will open a new PR to address other distros

PR #1658, opened 2023-07-28
:bug: Add missing packages

### LINKED ISSUE #1620: Add common linux utilities into the base packages

mauromorales:

Fixes #1620


Itxaka (dismissed):
nice

PR #1689, opened 2023-08-07
1609 release standard images

### LINKED ISSUE #1609: :warning:  Build provider-kairos in this repository

jimmykarily:
Fixes #1609


Earthfile
```diff
@@ -1092,3 +1177,67 @@ bump-repositories:
...
+
+      # We don't specify a version. To bump, just change what the latest version
+      # in the repository is.
+      RUN luet install -y system/provider-kairos
+    ELSE # Install from a branch
+      COPY github.com/kairos-io/provider-kairos:$PROVIDER_KAIROS_BRANCH+build-kairos-agent-provider/agent-provider-kairos /system/providers/agent-provider-kairos
+      RUN ln -s /system/providers/agent-provider-kairos /usr/bin/kairos
+    END
+
+INSTALL_NOHANG:
```
Itxaka, line 1204: should we just add this pacakge to the base images and ignore this altogether? Its just one package no?

Earthfile
```diff
@@ -1092,3 +1177,67 @@ bump-repositories:
...
+    COMMAND
+
+    ARG PROVIDER_KAIROS_BRANCH
+
+    COPY +luet/luet /usr/bin/luet
+
+    IF [[ "$PROVIDER_KAIROS_BRANCH" = "" ]] # Install with luet (released versions of the binary)
+      # If base image does not bundle a luet config use one
+      # TODO: Remove this, use luet config from base images so they are in sync
+      IF [ ! -e "/etc/luet/luet.yaml" ]
```
Itxaka, line 1192: I think this can be removed now.

The reason for this on the provider repo was that the base image from kairos could already provide the framework file and we could drop it from the provider repo, but now that we are building from here directly, we could just drop it altogether.

Earthfile
```diff
@@ -1092,3 +1177,67 @@ bump-repositories:
...
+    ELSE IF [ "$FLAVOR" = "ubuntu" ] || [ "$FLAVOR" = "ubuntu-20-lts" ] || [ "$FLAVOR" = "ubuntu-22-lts" ] || [ "$FLAVOR" = "debian" ]
+      RUN apt-get update && apt-get install -y nohang
+    END
+
+# Installs k3s (for "standard" images)
+INSTALL_K3S:
+    COMMAND
+
+    ARG K3S_VERSION
+    IF [[ "$K3S_VERSION" = "" ]]
```
Itxaka, line 1225: should we make this to be latest the default? Otherwise you get the full build until this point then it fails :/

Earthfile
```diff
@@ -1092,3 +1177,67 @@ bump-repositories:
...
+      ENV INSTALL_K3S_BIN_DIR="/usr/bin"
+      RUN curl -sfL https://get.k3s.io > installer.sh \
+          && INSTALL_K3S_SELINUX_WARN=true INSTALL_K3S_SKIP_START="true" INSTALL_K3S_SKIP_ENABLE="true" INSTALL_K3S_SKIP_SELINUX_RPM="true" bash installer.sh \
+          && INSTALL_K3S_SELINUX_WARN=true INSTALL_K3S_SKIP_START="true" INSTALL_K3S_SKIP_ENABLE="true" INSTALL_K3S_SKIP_SELINUX_RPM="true" bash installer.sh agent \
+          && rm -rf installer.sh
+    ELSE
+      ARG _LUET_K3S=$(k8s/k3s@${K3S_VERSION})
+    END
+
+    RUN luet install -y ${_LUET_K3S} utils/edgevpn utils/k9s utils/nerdctl container/kubectl utils/kube-vip && luet cleanup
```
Itxaka, line 1243: will this work if we go over the `[[ "$K3S_VERSION" = "latest" ]]` branch? It means that `${_LUET_K3S}` will be empty no? just woprried about luet misbehaving with empty spaces lol

Earthfile
```diff
@@ -98,19 +97,85 @@ go-deps-test:
...
+      RUN echo "export ISO_NAME=${OS_ID}-${VARIANT}-${DISTRO}-${TARGETARCH}-${MODEL}-${VERSION}.iso" >> ARTIFACT_NAMES_ENV
+    END
+
+    RUN echo "export CONTAINER_NAME=${PREFIX}-${FLAVOR}" >> ARTIFACT_NAMES_ENV
+    RUN echo "export IMG_CONTAINER_NAME=${PREFIX}-${FLAVOR}-img" >> ARTIFACT_NAMES_ENV
+  END
+
+  SAVE ARTIFACT ARTIFACT_NAMES_ENV AS LOCAL build/ARTIFACT_NAMES_ENV
+
+CONTAINER_IMAGE:
```
Itxaka, line 145: `CONTAINER_IMAGE_VERSION` maybe? `CONTAINER_IMAGE_NAMES` ?

To make it clear, because I saw it somewhere down there and I thought it may be building something or similar

.github/workflows/reusable-docker-arm-build.yaml
```diff
@@ -144,8 +145,11 @@ jobs:
           TAG: "latest"
           COSIGN_YES: true
         run: |
-          docker push "$IMAGE:$TAG"
-          image_ref=$(docker image inspect --format='{{index .RepoDigests 0}}' "$IMAGE:$TAG")
+          # Avoid pushing a new image for every commit (re-use latest)
+          export _IMG=$(cat build/IMAGE)
+          export _LATEST=$(echo $_IMG | cut -f1 -d:):latest
```
Itxaka, line 150: didnt we say that we were not using latest? We should push to master or a similar tag no?

Itxaka (dismissed):
All nits, not blockers

.github/workflows/reusable-docker-arm-build.yaml
jimmykarily: We decided to not fix this in this PR (which is already too big). We tried to change as little as possible.

Earthfile
jimmykarily: Should be possible, yes, will give this a try.

Earthfile
jimmykarily: It's easy to forget to pass this arg in a workflow. The result would be an artifact that will advertise itself (for example) as `1.25.11` but inside it would have the latest k3s. Given how Earthly caches things, when this fails, the next try is a lot faster. It's still annoying that it fails late but I think its worth it as a tradeoff, what do you think?

Earthfile
jimmykarily: We tried `latest` and it worked. Actually we intentionally refactored this code to work for the `latest` case because initially the `latest` branch didn't install `k9s` et al.

Earthfile
jimmykarily: Renamed to `CONTAINER_IMAGE_VERSION`

Earthfile
jimmykarily: From here as well I guess: https://github.com/kairos-io/kairos/pull/1689/files/3a483d716e3e9b464ff13f4ad2022f1f96d32e92#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R1229

Earthfile
jimmykarily: This never handled `almalinux`, `rockylinux`, `debian`... I should probably add the package to all of them?

Earthfile
```diff
@@ -1092,3 +1177,67 @@ bump-repositories:
...
+INSTALL_NOHANG:
+    COMMAND
+
+    ARG FLAVOR
+
+    # Install nohang
+    IF [ "$FLAVOR" = "opensuse-leap" ] || [ "$FLAVOR" = "opensuse-leap-arm-rpi" ]
+      RUN zypper ref && zypper in -y nohang
+    ELSE IF [ "$FLAVOR" = "alpine-ubuntu" ] || [ "$FLAVOR" = "alpine-opensuse-leap" ] || [ "$FLAVOR" = "alpine-arm-rpi" ]
+      RUN apk add grep
```
jimmykarily, line 1213: :confused: So since alpine doesn't have `nohang` we decided to install `grep` instead ? :D

jimmykarily: We only enable this on systemd: https://github.com/kairos-io/kairos/blob/master/overlay/files/system/oem/09_services.yaml#L21-L22

So the question is, how does this work on systemd flavors that don't have it?
(fedora, rockylinux, almalinux)

Itxaka: lol

Earthfile
Itxaka: yeah good to me, I would like a different option for this to fail earlier but this is good for now, lets merge this asap!

Itxaka (approved):
lets gooooo

images/Dockerfile.ubuntu-20-lts-arm-nvidia-jetson-agx-orin
```diff
@@ -28,6 +28,10 @@ ARG FRAMEWORK_VERSION=35.3
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y software-properties-common
 
+# nohang
+RUN add-apt-repository ppa:oibaf/test
+RUN apt-get install -y nohang
+
```
jimmykarily, line 34: This is not really needed since we don't build standard images for this flavor (https://github.com/kairos-io/kairos/pull/1689/files#diff-5cb44b260a8400267a5ff9938571e2022a5ef59730459941d173f5b5fffb389c).

PR #1702, opened 2023-08-09
:wrenchIcon: Fully remove machine-id and systemd-firstboot

Itxaka:
Remove machine-id fully on building adn mask teh systemd-firstboot service so it doesnt kick in

Fixes #


mauromorales (approved):
lgtm

Earthfile
```diff
@@ -378,20 +378,20 @@ base-image:
...
         ln -sf /etc/init.d/cos-setup-network /etc/runlevels/default/cos-setup-network  && \
         ln -sf /etc/init.d/cos-setup-reconcile /etc/runlevels/default/cos-setup-reconcile && \
         ln -sf /etc/init.d/kairos-agent /etc/runlevels/default/kairos-agent
     # Otherwise we assume systemd
     ELSE
-      # Empty machine-id so we dont accidentally run systemd-firstboot ¬_¬
-      RUN rm -rf /etc/machine-id && touch /etc/machine-id && chmod 444 /etc/machine-id
-      RUN ls -liah /etc/systemd/system
+      # mask systemd-firstboot dont accidentally run block booting
+      RUN systemctl mask systemd-firstboot
```
jimmykarily, line 394: According to [the docs](https://www.freedesktop.org/software/systemd/man/systemd-firstboot.html), we are not only masking the creation of machine-id but also:

```
The following settings may be configured:

    - The machine ID of the system
    - The system locale, more specifically the two locale variables LANG= and LC_MESSAGES
    - The system keyboard map
    - The system time zone
    - The system hostname
    - The kernel command line used when installing kernel images
    - The root user's password and shell
```

If that ok? Did we rely on this for any of the other settings?

Itxaka: we dont rely on anything there. systemd-firstboot is there for not initialized machine to configure the system on the first boot, but we are not a normal system, we already have all that configured in the rootfs :)

jimmykarily (approved):

PR #1703, opened 2023-08-09
:sparkles: Refactor Dockerfiles

mauromorales:
This PR makes use of multi-stage scripting to merge the existing ARM and AMD files.

As mentioned in #1656 and #1400, it makes it easier to maintain, since there's only one file where to put changes, so we don't forget to update different files. Plus it also makes it clear what are the differences across models, architectures, etc.


Itxaka (approved):
damn, pretty cool. And very nice with the comments, it makes it much easier to read!

PR #1710, opened 2023-08-10
Fix arm pipeline dependency

mauromorales:
Master ARM pipeline is currently failing with

```
Invalid workflow file: .github/workflows/image-arm.yaml#L90
The workflow is not valid. .github/workflows/image-arm.yaml (Line: 90, Col: 9): Job 'notify' depends on unknown job 'docker'.
```
see https://github.com/kairos-io/kairos/actions/runs/5819487177

this is because the name was changed during the refactoring


mauromorales:
I will cancel the runners since the fix will only run on master

PR #1748, opened 2023-08-21
Bundle luet versions in the framework

Itxaka:
Use the new luet 0.35 which provides a way of storing the installed artifacts in the system to bundle the versions file with the framework files so we always have a reference for what was installed in the framework and what versions.

Also reworks the CI pipeline for the bump repos to provide a diff of changes to **installed** packages, not to the full repo unfortunately, as part of the PR creation so its easily visible what changed that directly affects kairos

Also during iso building, extract and upload the versions to the artifacts so its easier to check what packages were installed in a given iso during CI.

Fixes #

### DIFF
--- a/.github/workflows/reusable-build-provider.yaml
+++ b/.github/workflows/reusable-build-provider.yaml
@@ -69,6 +69,7 @@ jobs:
           path: |
             *.iso
             *.sha256
+            versions.yaml
           if-no-files-found: error
       - name: Push to testing
         run: |
--- a/.github/workflows/bump_repos.yml
+++ b/.github/workflows/bump_repos.yml
@@ -13,8 +13,36 @@ jobs:
         with:
           repository: quay.io/kairos/packages
           packages: utils/earthly
+      - name: Old packages
+        run: |
+          earthly +base-image --VARIANT=standard --FLAVOR=opensuse-leap --K3S_VERSION=latest
+          mv build/versions.yaml build/versions.old.yaml
       - name: Bump cos 🔧
         run: earthly +bump-repositories
+      - name: New packages
+        run: |
+          earthly +base-image --VARIANT=standard --FLAVOR=opensuse-leap --K3S_VERSION=latest
+          mv build/versions.yaml build/versions.new.yaml
+      - name: Diff versions
+        run: |
+          yq -i -P '.|=sort_by(.name)|.[]|[{"name": .name, "category": .category, "version": .version}]' build/versions.old.yaml
+          yq -i -P '.|=sort_by(.name)|.[]|[{"name": .name, "category": .category, "version": .version}]' build/versions.new.yaml
+          echo "Bump of Kairos repositories" > pr-message
+          echo "--------------------------" >> pr-message
+          DIFF=$(diff -u build/versions.old.yaml build/versions.new.yaml)
+          if [[ $? == 1 ]]; then
+            echo "> [\!WARNING]" >> pr-message
+            echo "> There were changes to installed packages" >> pr-message
+            echo "\`\`\`diff" >> pr-message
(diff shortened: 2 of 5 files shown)


Itxaka:
PR messages looks like:






Bump of Kairos repositories
----
> [!WARNING]
> There were changes to installed packages
```diff
--- build/versions.old.yaml	2023-08-21 13:05:18.190300544 +0200
+++ build/versions.new.yaml	2023-08-21 13:05:18.208300883 +0200
@@ -1,6 +1,6 @@
 - name: grub-config
   category: static
-  version: "0.4"
+  version: "0.5"
 - name: grub2-efi
   category: system
   version: 2.06-150404
```


> [!IMPORTANT]
> Full package list from new repo
```yaml
- name: grub-config
  category: static
  version: "0.5"
- name: grub2-efi
  category: system
  version: 2.06-150404
- name: immucore
  category: dracut
  version: 0.1.2
- name: immucore
  category: system
  version: 0.1.2
- name: kairos-agent
  category: system
  version: 2.2.7
- name: kcrypt
  category: system
  version: 0.7.0
- name: kcrypt-challenger
  category: system
  version: 0.5.0
- name: network
  category: systemd-default
  version: "1.0"
- name: network-legacy
  category: dracut
  version: "0.9"
- name: suc-upgrade
  category: system
  version: 0.2.1
- name: sysext
  category: dracut
  version: "0.9"
- name: systemd
  category: init-svc
  version: 0.8.0
```

.github/workflows/reusable-build-provider.yaml
```diff
@@ -63,6 +63,15 @@ jobs:
           earthly +ci --SECURITY_SCANS=false --IMAGE=ttl.sh/kairos-${{ inputs.flavor }}-${{ github.sha }}-provider:24h --FLAVOR=${{ inputs.flavor }} --VARIANT=standard -K3S_VERSION=latest
           sudo mv build/* .
           sudo rm -rf build
+      - name: Extract packages versions
+        run: |
+          earthly +framework-versions --IMAGE=ttl.sh/kairos-${{ inputs.flavor }}-${{ github.sha }}-provider:24h --FLAVOR=${{ inputs.flavor }} --VARIANT=standard -K3S_VERSION=latest
```
jimmykarily, line 68: `framework` target (or `framework-versions`) has no need for `IMAGE` , `VARIANT` nor `K3S_VERSION` to be set. You can remove these. You only need to set `FLAVOR`.

Itxaka: BTW I dont even know if it works yet, and if it does I think it could be part of the +ci or +all targets, especially because on the provider we have to generate it later.

jimmykarily: It could be an artifact of `base-image` like `build/IMAGE` and `build/VERSION` so that we don't have to run the framework-versions target explicitly (we may not even need a new target then)

Itxaka: problem is, it needs to be runs after the framework stuff is installed as it requires the luet database to be intact but just after that we remove the database and all temp files, so it needs to be done twice for the provider I think, first ont eh base framework after just installing and then in the provider after installing k3s. And then on provider it needs to be merged together with the normal base versions xD

Itxaka: > It could be an artifact of `base-image` like `build/IMAGE` and `build/VERSION` so that we don't have to run the framework-versions target explicitly (we may not even need a new target then)

good one, I used that for the isos and the framework for the images so both have their versions included. Plus delayed the database removal so we can gather all the provider and k3s artifact versions !

mauromorales (approved):
fancy!

PR #1749, opened 2023-08-21
Fix test that wasn't testing anything

jimmykarily:
because it was returning success too early since the checks were too loose or even wrong.

E.g. we didn't even wait for the Pods to be ready (wrong test condition) and we only tested that there is a "v" in the version which was true even before the upgrade.

Fixes #

### DIFF
--- a/tests/assets/suc.yaml
+++ b/tests/assets/suc.yaml
@@ -9,13 +9,14 @@ metadata:
 spec:
   concurrency: 1
   #version:  latest
-  version: "opensuse-v1.23.5-44"
+  version: "v2.3.1-k3sv1.25.11-k3s1"
+
   nodeSelector:
     matchExpressions:
       - {key: kubernetes.io/hostname, operator: Exists}
   serviceAccountName: system-upgrade
   cordon: false
   upgrade:
-    image: quay.io/kairos/kairos
+    image: quay.io/kairos/kairos-opensuse-leap
     command:
     - "/usr/sbin/suc-upgrade"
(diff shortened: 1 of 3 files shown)


tests/provider_upgrade_k8s_test.go
```diff
@@ -145,6 +145,6 @@ var _ = Describe("k3s upgrade test", Label("provider", "provider-upgrade-k8s"),
 			out, _ = kubectl(vm, "get pods -A")
 			version, _ := vm.Sudo(getVersionCmd)
 			return version
-		}, 30*time.Minute, 10*time.Second).Should(ContainSubstring("v"), out)
+		}, 30*time.Minute, 10*time.Second).Should(ContainSubstring("v2.3.1-k3sv1.25.11+k3s1"), out)
```
mauromorales, line 148: maybe this should read the tests/assets/suc version

tests/assets/suc.yaml
```diff
@@ -9,13 +9,14 @@ metadata:
 spec:
   concurrency: 1
   #version:  latest
-  version: "opensuse-v1.23.5-44"
+  version: "v2.3.1-k3sv1.25.11-k3s1"
```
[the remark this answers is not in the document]
jimmykarily: Actually this test is doing a downgrade. Because what we build here is usually newer than any released image. Without changing this, it doesn't matter much to which version we downgrade too. If this breaks we just downgrade to a newer one (if that's the problem at all).

I don't like this much but I'm only trying to make the test actually test something. It needs further improvement to make more sense.

tests/provider_upgrade_k8s_test.go
jimmykarily: true, I'll give this a try.

jimmykarily: done

mauromorales (approved):
lgtm

PR #1753, opened 2023-08-22
Fix linting errors

jimmykarily:
https://www.shellcheck.net/wiki/SC2028
https://www.shellcheck.net/wiki/SC2129
https://www.shellcheck.net/wiki/SC2005

### DIFF
--- a/.github/repo-diff-msg.sh
+++ b/.github/repo-diff-msg.sh
@@ -12,7 +12,7 @@ if [[ $? == 1 ]]; then
     echo "\`\`\`diff"
     echo "${DIFF}"
     echo "\`\`\`"
-    printf '\\n'
+    echo
   } >> pr-message
 fi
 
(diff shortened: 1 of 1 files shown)


.github/repo-diff-msg.sh
```diff
@@ -6,16 +6,22 @@ echo "Bump of Kairos repositories" > pr-message
...
-  echo "${DIFF}" >> pr-message
-  echo "\`\`\`" >> pr-message
-  echo "\n" >> pr-message
+  {
+    echo "> [\!WARNING]"
+    echo "> There were changes to installed packages"
+    echo "\`\`\`diff"
+    echo "${DIFF}"
+    echo "\`\`\`"
+    printf '\\n'
```
jimmykarily, line 15: I'm not sure if we actually wanted a literal `\n` or a rendered new line. Given a simple `echo ""` would print a new line, I assumed we wanted a literal bashlash-n

mauromorales (dismissed):
lgtm

.github/repo-diff-msg.sh
mauromorales: I think we want an enter actually because the literal \n doesn't help here: https://github.com/kairos-io/kairos/pull/1744#pullrequestreview-1587408185

jimmykarily: True! I wasn't sure where the the text was actually printed, thanks! Will change to a simple `echo` then.

jimmykarily: done

PR #1768, opened 2023-08-25
Fix rpi4 serial

jimmykarily:
Fixes the serial output on rpi4. We also need to check if this works on rpi3 otherwise we have to apply this only on rpi4.

Also this fix is an explicit fix  for the `mini-uart's driver (8250)`. I'm not sure if there are other versions of that driver out there that need a similar option in the cmdline

### DIFF
--- a/images/Dockerfile.ubuntu
+++ b/images/Dockerfile.ubuntu
@@ -154,6 +154,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     linux-raspi \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 COPY rpi/bootargs.cfg /etc/cos/bootargs.cfg
+COPY rpi/config.txt /boot/config.txt
 
 FROM rpicommon AS ubuntu-20-lts-rpi
 RUN apt-get update && apt-get install -y --no-install-recommends \
--- a/images/rpi/config.txt
+++ b/images/rpi/config.txt
@@ -1,7 +1,6 @@
 # https://emteria.com/kb/connect-uart-rpi 
 # https://www.raspberrypi.com/documentation/computers/config_txt.html
 enable_uart=1
-console=serial0
 uart_2ndstage=1
 
 # Disabling bluetooth might change the serial device:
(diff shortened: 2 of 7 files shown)


mauromorales (dismissed):
🙌

images/rpi/config.txt
```diff
@@ -0,0 +1,10 @@
+# https://emteria.com/kb/connect-uart-rpi 
+# https://www.raspberrypi.com/documentation/computers/config_txt.html
+enable_uart=1
+console=serial0
```
jimmykarily, line 4: I gave this a try on rpi4 (by setting `console=serial0,115200` in bootargs) and it didn't work. Not sure if it's needed.

jimmykarily: Maybe it's a good idea to have a "rpi-config" bundle for the config.txt. This way the user's could just specify the options they want in the kairos config and the bundle will make sure they end up in the `/boot/config.txt`.

mauromorales: can't they replace it by putting a different file over it in a later stage? at least as a workaround?

jimmykarily: To be honest, I don't know how early this is read by rpi. Maybe before we even pivot to the actual system? Maybe it means this one doesn't have any effect? We need to test this more.

images/Dockerfile.ubuntu
```diff
@@ -154,6 +154,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     linux-raspi \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 COPY rpi/bootargs.cfg /etc/cos/bootargs.cfg
+COPY rpi/config.txt /boot/config.txt
```
jimmykarily, line 157: Let's add this on opensuse leap and TW and alpine too

images/rpi/config.txt
jimmykarily: I now remembered, when I used the option to disable the bluetooth, the serial device appeared as `ttyAMA0` which proves that it was read. Let's keep it around as is for now and worry later on how users can override it (https://github.com/kairos-io/kairos/issues/1392).

jimmykarily: I removed this line and it works fine.

images/Dockerfile.ubuntu
jimmykarily: done

Issue #1766, opened 2023-08-24
🤖 Release needs to be triggered multiple times because we hit an API limit


mauromorales:
Not that we moved the building of the standard images to this repository, we constantly hit the following error on release

```
⚠️ Unexpected error fetching GitHub release for tag refs/tags/v2.4.0-alpha3: HttpError: API rate limit exceeded for installation ID 29320181.
Error: API rate limit exceeded for installation ID 29320181.
```

Maybe we should serialize and build core images first and only until those are done, build standard ones, and potentially those also in batches by k3s versions

mauromorales:
Shoot if it's not an issue it's another :P 

```
Run softprops/action-gh-release@v1
♻️ Deleting previously uploaded asset IMAGE...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-grype.json...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-grype.sarif...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-initrd...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-ipxe-usb.img...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-ipxe.iso...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-kernel...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-sbom.spdx.json...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-sbom.syft.json...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-trivy.html...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-trivy.json...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4-trivy.sarif...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4.ipxe...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4.iso...
⬆️ Uploading kairos-core-rockylinux-amd64-generic-v2.4.0-alpha4.iso.sha256...
[...]

PR #1771, opened 2023-08-28
Reduce calls to release API

mauromorales:
relates to #1766 

From the generic, there are files that don't need to be uploaded, with this change we only upload iso and sha

From the other models, we group all json files into a single scan-reports.tar.gz so we make a few calls less per flavor/version


mauromorales:
cancelling runs since they are only for release

jimmykarily (approved):

PR #1773, opened 2023-08-28
Fix some release API issues

mauromorales:
relates to #1766

- reduces some of the calls on the amd release uploads
- fixes the issue with bad substitution from the previous PR
- removes files which should not be uploaded: VERSION, IMAGE and versions.yaml


mauromorales:
this only runs on release so cancelling jobs

jimmykarily (approved):

Issue #1708, opened 2023-08-10
:robot: Reset test is very flaky


Itxaka:
**Kairos version:**
master

**CPU architecture, OS, and Version:**
all

**Describe the bug**
Reset test is failing a lot lately, we should check to see if we can improve it somehow to be less flaky or identify if there is a corner case where it makes it fail due to some specific issue.

**To Reproduce**
check reset jobs on build PRs pipeline

**Expected behavior**
doesnt fail all the time

**Logs**

**Additional context**

jimmykarily:
Let's wait until master CI is green ( to avoid debugging irrelevant issues ), then let's create a PR that only runs this test (disabling the rest) and finally let's keep triggering this until it fails, because this was not reproducible locally and we no longer have a link to a failed job.

We want the make sure there isn't something wrong with our reset workflow, that's why we are insisting on this one.

mauromorales:
The test has been consistently green in the past 10 runs ... will try to run multiple times locally to validate

mauromorales:
still green on every master run and for many local runs ... we've been merging changes on peg, upgrading packages so hard to tell if one of this changes fixed it, we also have the faster runners which might have made the test more reliable. Closing for now but feel free to re-open

PR #1776, opened 2023-08-29
:sparkles: Cache testing image and simplify scripts

mauromorales:
- adds a golang-testing image which is cached for the golang version. No need to re-build on every test saving some time
- Reduce external call to generate datasource-iso, since the test call already does it internally and the artifact is not being used anywhere else 

relates to #1708


mauromorales:
no need to run arm so cancelling jobs

jimmykarily (approved):

PR #1780, opened 2023-08-30
Switch to self-hosted runner

jimmykarily:
because our paid runners stopped working. This is a temporary fix until we find the problem.


mauromorales (approved):
lgtm

mauromorales:
luet seems to be needed in the runner

jimmykarily:
> luet seems to be needed in the runner

I think it's docker that was missing on my runner. We install luet using docker (we extract the binary from the docker image). I now have docker, let's see.

mauromorales (dismissed):

PR #1781, opened 2023-08-30
Easy option to build non-hwe kernels

### LINKED ISSUE #1779: Native non-HWE images
Mandating that Ubuntu images all have the HWE kernel is problematic for supporting applications that are kernel specific.

Provide a non-HWE version of the images.

It is manual and tedious and seems to change with different versions of Kairos to create non-HWE images outside of the repo buildstream.  For example, now all Ubuntu versions are combined into a single Dockerfile which means retooling for those that were building custom images if we want to move forward.

**Additional context**
Non-HWE images should be provided as an option.  Not everyone wants the latest and greatest especially when using LTS in a stable environment

mauromorales:

Adds a flag `HWE` to build ubuntu without HWE kernles. It can be used as follows

```
earthly +all --FLAVOR=ubuntu-20-lts --HWE=-non-hwe
```

if you want the HWE kernel you don't need to pass it

It only works for LTS versions

Fixes #1779

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -300,6 +300,7 @@ base-image:
     ARG KAIROS_VERSION
     ARG BUILD_INITRD="true"
     ARG TARGETARCH
+    ARG HWE
 
     IF [ "$BASE_IMAGE" = "" ]
         # DISTRO is used to match the Linux distribution in the Dockerfile e.g. Dockerfile.ubuntu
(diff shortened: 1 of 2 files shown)


Earthfile
```diff
@@ -300,6 +300,7 @@ base-image:
     ARG KAIROS_VERSION
     ARG BUILD_INITRD="true"
     ARG TARGETARCH
+    ARG HWE
```
jimmykarily, line 303: What are the 2 possible values?

`-hwe` and `-non-hew` ? Since there is no use of this yet in the pipelines, it would be useful to add a comment with the possible values.

jimmykarily (approved):

PR #1796, opened 2023-09-07
:art: Bundle bootargs with framework images

### LINKED ISSUE #1770: You need to load the kernel first
Tried following instructions from: https://kairos.io/docs/reference/build-from-scratch/#build-bootable-asset
After a successful installation I cannot start the image due to:
```
Booting a command list

error: ../../grub-core/disk/loopback.c:98:device name already exists.
error: ../../grub-core/kern/fs.c:170:invalid file name `'.
error: ../../grub-core/loader/i386/efi/linux.c:168:you need to load the kernel first.
```

This also happens with image from: https://github.com/kairos-io/kairos/blob/master/examples/byoi/fedora-fips/Dockerfile

Itxaka:
Currently bootargs is not provided with the framework images so its missing a very important piece to boot which results into BYOI being broken.

This patch adds the bootargs.cfg to the framework image. Also the config.txt to the rpi images.

Also drops the kernel/initramfs vars from the bootargs as they are set to the default now in the package.

Also moves the luet clean under the package list generation as it requires the database of luet to list the packages.

Fixes #1770 

Requires: https://github.com/kairos-io/packages/pull/408

### DIFF
--- a/.github/workflows/reusable-upgrade-latest-test.yaml
+++ b/.github/workflows/reusable-upgrade-latest-test.yaml
@@ -51,13 +51,8 @@ jobs:
           # A flag to set the download target as latest release
           # The default value is 'false'
           latest: true
-          fileName: '*${{ inputs.flavor }}-v*.iso'
+          fileName: 'kairos-${{ inputs.flavor }}-v*.iso'
           out-file-path: ""
-      - name: Download ISO
-        id: iso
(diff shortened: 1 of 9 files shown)


jimmykarily (dismissed):

Itxaka:
missing package update with the default kernel/initrd variable set...

Issue #1797, opened 2023-09-07
k3s service is broken on systemd


jimmykarily:
systemd has a compatibility feature with sysv (https://www.turnkeylinux.org/blog/debugging-systemd-sysv-init-compat). Since we copy all services, both systemd and openrc, when the service is started, it fails with an error.

We need to remove the openrc bits from systemd systemds

jimmykarily:
> How is this not hitting our test suite? We have a check for k3s status no?

As it seems, no. We check various things but not that the service is running. We need to do add that check.

Itxaka:
ah maybe its because on testing we use latest and that installs the k3s service directly! So it only installs the proper service file, while installing from package would install our package with both files.

PR #1802, opened 2023-09-08
Install the right k3s package based on the init system

jimmykarily:
Fixes: #1797

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -416,7 +416,7 @@ base-image:
     IF [ "$VARIANT" = "standard" ]
         DO +PROVIDER_INSTALL -PROVIDER_KAIROS_BRANCH=${PROVIDER_KAIROS_BRANCH}
 
-        DO +INSTALL_K3S
+        DO +INSTALL_K3S --FLAVOR=$FLAVOR
 
         # Redo os-release with override settings to point to provider-kairos stuff
         # in earthly 0.7 we will be able to just override VARIANT here and just run the OSRELEASE once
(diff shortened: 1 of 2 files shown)


Itxaka:
Not sure this will work...

Not FLAVOR ARG in the COMMAND so it wont pick the --FLAVOR I think
Also, repos are not updated so the packages wont be in the older repo tag

Earthfile
```diff
@@ -416,7 +416,7 @@ base-image:
     IF [ "$VARIANT" = "standard" ]
         DO +PROVIDER_INSTALL -PROVIDER_KAIROS_BRANCH=${PROVIDER_KAIROS_BRANCH}
 
-        DO +INSTALL_K3S
+        DO +INSTALL_K3S --FLAVOR=$FLAVOR
```
Itxaka, line 419: This passes the arg but I dont think the command is getting it as its not declared there

jimmykarily: Yes, earthly is strange. This is a `DO` so it gets everything propagated automatically. Don't ask me more... :D

Itxaka (approved):

Earthfile
mudler: https://docs.earthly.dev/docs/earthfile#description-21

jimmykarily: ok found it. `FLAVOR` is a global (at the top of the file). That's why specifying it or not doesn't make any difference. Since we do that in other places, I will make it a "local" ARG for consistency.

jimmykarily:
To save some time, I merged it without waiting for CI. If it breaks we fix on master. Maybe we get enough time to cut a release.

Earthfile
Itxaka: Interesting, so it inherits the caller context and args but it also creates its own context. weird.

PR #1814, opened 2023-09-13
Use one of our luet packages to install k3s

jimmykarily:
Fixing this: https://github.com/kairos-io/kairos/issues/1797#issuecomment-1711315693


Itxaka (dismissed):

jimmykarily:
Same problem as in master: 

```
ls: cannot access '*.iso': No such file or directory
```

we moved the releases to kairos and it can't find one until we release at least once. Maybe we should just merge this and fix it post-release if it breaks.

PR #1854, opened 2023-09-25
Improve uki iso stuff

Itxaka:

 - Splits UKI tests into its own file as we may need to do heavy changes to it
 - Create /oem and /efi in the rootfs whehn building uki artifacts
 - Add missing oem labels in the cmdline
 - Comment out the generation of signing keys for uki artifacts
 - Bundle some pregenerated keys to sign efi artifacts for testing
 - ALWAYS sign the efi artifacts, no matter if we are using secure boot or not, it doesnt hurt (NOTE: This needs to be dropped on release, follow up needed)
 - Sign systemd-boot with our test keys for testing
 - Change boot method for uki. Now the fallback for uefi is set to systemd-boot which loads a config file for kairos. That way several efi variables get set properly. Also we gain having several versions in there and fallback.
 - Bump peg to support detaching CD on qemu
 - Bundle a custom efivars file for qemu that has secureboot enabled and our keys injected so we can test secureboot directly.
 - Add efivars.json with the stored efivars variables so we can regenerate the efivars.fd file
 - Add efivars.md to explain the efivars use and regeneration
 - Adds tests for uki:
   - boot mode (install or boot)
   - secureboot enabled
   - mounts missing on install mode
   - mounts exist on boot mode

Fixes #


Earthfile
```diff
@@ -495,8 +495,10 @@ uki-artifacts:
     FROM +base-image --BUILD_INITRD=false
     RUN /usr/bin/immucore version
     RUN ln -s /usr/bin/immucore /init
+    RUN mkdir -p /oem # be able to mount oem under here if found
```
Itxaka, line 498: this makes it easier to deal with mounts later

Earthfile
```diff
@@ -495,8 +495,10 @@ uki-artifacts:
...
+    RUN mkdir -p /efi # mount the esp under here if found
     RUN find . \( -path ./sys -prune -o -path ./run -prune -o -path ./dev -prune -o -path ./tmp -prune -o -path ./proc -prune \) -o -print | cpio -R root:root -H newc -o | gzip -2 > /tmp/initramfs.cpio.gz
-    RUN echo "console=tty1 console=ttyS0 net.ifnames=1 rd.immucore.debug rd.immucore.uki selinux=0" > /tmp/Cmdline
+    RUN echo "console=tty1 console=ttyS0 net.ifnames=1 rd.immucore.oemlabel=COS_OEM rd.immucore.oemtimeout=2 rd.immucore.debug rd.immucore.uki selinux=0" > /tmp/Cmdline
```
Itxaka, line 501: was missing the oem config

Earthfile
```diff
@@ -545,15 +547,15 @@ uki-signed:
     # Platform key
     RUN openssl req -new -x509 -subj "/CN=Kairos PK/" -days 3650 -nodes -newkey rsa:2048 -sha256 -keyout PK.key -out PK.crt
     # CER keys are for FW install
-    RUN openssl x509 -in PK.crt -out PK.cer -outform DER
+    RUN openssl x509 -in PK.crt -out PK.auth -outform DER
```
Itxaka, line 550: auth extension is the one used by systemd-boot to try to auto-enroll keys

Earthfile
```diff
@@ -646,16 +648,40 @@ iso-uki:
...
     FROM $OSBUILDER_IMAGE
     WORKDIR /build
-    COPY +uki/uki.efi /build/uki.efi
+    COPY +uki-signed/uki.efi .
+    COPY +uki-signed/PK.auth .
+    COPY +uki-signed/KEK.auth .
+    COPY +uki-signed/DB.auth .
+    COPY +uki-signed/MokManager.efi .
+    # Set the name for kairos manually as otherwise it picks it from the os-release automatically
+    RUN printf "title Kairos ${FLAVOR} ${VERSION}\nefi /EFI/kairos/kairos.efi" > kairos.conf
```
Itxaka, line 657: cretes the config for systemd-boot to show in the entries

Earthfile
```diff
@@ -646,16 +648,40 @@ iso-uki:
...
     RUN mmd -i /build/efi/efiboot.img ::EFI
     RUN mmd -i /build/efi/efiboot.img ::EFI/BOOT
+    RUN mmd -i /build/efi/efiboot.img ::EFI/kairos
+    RUN mmd -i /build/efi/efiboot.img ::EFI/tools
+    RUN mmd -i /build/efi/efiboot.img ::loader
+    RUN mmd -i /build/efi/efiboot.img ::loader/entries
+    RUN mmd -i /build/efi/efiboot.img ::loader/keys
+    RUN mmd -i /build/efi/efiboot.img ::loader/keys/kairos
+    # Copy keys, not sure which ones lol
+    RUN mcopy -i /build/efi/efiboot.img /build/PK.auth ::loader/keys/kairos/PK.auth
```
Itxaka, line 672: copy the keys into the iso, this is just for testing purpouses

framework-profile.yaml
```diff
@@ -173,9 +173,9 @@ repositories:
     priority: 2
     urls:
       - "quay.io/kairos/packages"
-    reference: 20230920112615-repository.yaml
+    reference: 20230925151606-repository.yaml
```
Itxaka, line 176: newer package with fixes in yip configs for rootfs mounting and accounts

Earthfile
```diff
@@ -520,100 +539,40 @@ uki:
...
+    #    --uname Uname \
+    #    --stub /usr/lib/systemd/boot/efi/linuxx64.efi.stub \
+    #    --secureboot-private-key DB.key \
+    #    --secureboot-certificate DB.crt \
+    #    --sign-kernel \
+    #    --pcr-private-key private.pem \
+    #    --pcr-public-key public.pem \
+    #    --measure \
+    #    --output $ISO_NAME.signed.efi
+    RUN sbsign --key DB.key --cert DB.crt --output systemd-bootx64.signed.efi /usr/lib/systemd/boot/efi/systemd-bootx64.efi
```
Itxaka, line 568: alwasy sign our efi artifacts, makes no sense to test them without signature

.github/workflows/uki.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Install earthly
+        uses: Luet-lab/luet-install-action@v1.1
+        with:
+          repository: quay.io/kairos/packages
+          packages: utils/earthly
+      - name: Build uki ISO 🔧
+        run: |
+          # Do fedora as its the smaller uki possible
+          # Use immmucore master as it has patches not released for uki
+          # Use kairos-agent main branch as it has patches not released for uki
```
mauromorales, line 42: i guess this will need to be updated once those patches are merged

.github/workflows/uki.yaml
```diff
@@ -0,0 +1,60 @@
...
+          KVM: true
+          MEMORY: 4000
+          CPUS: 2
+          FIRMWARE: /usr/share/OVMF/OVMF_CODE.fd
+        run: |
+          export ISO=$(ls $PWD/build/kairos-core-*opensuse-leap*.iso)
+          cp tests/go.* .
+          go run github.com/onsi/ginkgo/v2/ginkgo -v --label-filter "uki" --fail-fast -r ./tests/
+      - uses: actions/upload-artifact@v3
+        if: failure()
```
mauromorales, line 56: I guess this will help debug the issue ... how does one debug from here?

Earthfile
mauromorales: why does this only apply for UKI artifacts?

Earthfile
mauromorales: when releasing, I guess we don't want to have that immucore debug there ... I wonder if these could be passed in some other way to not have different targets depending on whether it's a release or a normal build

.github/workflows/uki.yaml
Itxaka: yep! Although with the heavy testing, Im guessing this can be left in master for both to keep testing it

.github/workflows/uki.yaml
Itxaka: gathers logs only if it fails.

Earthfile
Itxaka: there doesnt seem to be an oem dir on the rootfs when generating the uki artifact so if we want to mount something in there on boot, the path needs to exist before. No ieda why the rootfs doesnt have oem....

mauromorales: but should we create it at the Dockerfile then?

Earthfile
Itxaka: yes, there is a nice feature of systemd-boot in which you can have a separated efi artifact that contains the comand line only! And that one is also signed so supposedly it works with secureboot and such.

That would allow us to also have an install entry cmdline, a boot entry cmdline and a debug entry cmdline with minimal work.

Unfortunately I had no time to investigate this further but there is some work ongoing and it should be possible (but experimental) to do so: https://github.com/systemd/systemd/pull/27358

Earthfile
Itxaka: No idea, I need to investigate because the isos and images DO have it...so it has to be created somewhere...

mauromorales (approved):
lgtm

PR #1855, opened 2023-09-26
Do not remove cache dir

### LINKED ISSUE #1849: Core Ubuntu Images fail to install apt packages
When using the core ubuntu images, installing apt packages fails unless the system is updated and /var/cache/apt/archives/partial directoty is created manually.

For example the following containerfiles both fail to build:
```dockerfile
FROM quay.io/kairos/core-ubuntu-22-lts:v2.4.0
# install any package
RUN apt install -y fish
```
error: `#0 0.499 E: Unable to locate package fish` (this applies to any package fish is just an example)
```dockerfile
FROM quay.io/kairos/core-ubuntu-22-lts:v2.4.0
RUN apt update -y
# install any package
RUN apt install -y fish
```
error: `#0 2.103 E: Archives directory /var/cache/apt/archives/partial is missing. - Acquire (2: No such file or directory)`

but if the system is updated *and* the cache dir is created
```dockerfile
FROM quay.io/kairos/core-ubuntu-22-lts:v2.4.0
# fix bug
RUN apt update -y && mkdir -p /var/cache/apt/archives/partial
# install any package
RUN apt install -y fish
```
then the container builds just fine.

I've tried this on:  
`quay.io/kairos/core-ubuntu-20-lts:v2.4.0`  
`quay.io/kairos/core-ubuntu-20-lts:latest`  
`quay.io/kairos/core-ubuntu-22-lts:v2.4.0`  
`quay.io/kairos/core-ubuntu-22-lts:latest`  
and they all have the same results. I haven't tried other tags.

Requiring the system to be updated at build time seems fine if that's the desired workflow...this just wasn't required when I tried modifying the opensuse images so not sure if this is a bug or a feature. 

That being said IMO the `/var/cache/apt/archives/partial` shouldn't need to be manually created.

mauromorales:
Maybe this is not the right solution, but just want to get the ball rolling to talk about which one is

Fixes #1849

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -477,7 +477,6 @@ base-image:
     # luet cleanup
     RUN luet cleanup
     RUN rm -rf /var/luet
-    RUN rm -rf /var/cache
 
     SAVE IMAGE $_CIMG
     SAVE ARTIFACT IMAGE AS LOCAL build/IMAGE
--- a/images/Dockerfile.ubuntu
+++ b/images/Dockerfile.ubuntu
@@ -244,5 +244,4 @@ RUN systemctl enable ssh
 # Fixup sudo perms
 RUN chown root:root /usr/bin/sudo && chmod 4755 /usr/bin/sudo
 
-# Clear cache
-RUN rm -rf /var/cache/* && journalctl --vacuum-size=1K && rm /etc/machine-id && rm /var/lib/dbus/machine-id && rm /etc/hostname
+RUN journalctl --vacuum-size=1K && rm /etc/machine-id && rm /var/lib/dbus/machine-id && rm /etc/hostname


Itxaka:
after installing everything in our dockerfile my /var/cache is ridiculously small:

```
root@5bd452d0d714:/var/cache# du -chs *
0	PackageKit
0	adduser
0	apparmor
0	apt
3.9M	debconf
16K	ldconfig
0	private
0	snapd
4.0M	total
```
I think this is the correct approach, as we are already running apt-get clean which should remove the cache and stuff (maybe move it to the end of the file?)

The only thing in there left are the debconf templates adn that hasnt give us any issues. If we want to clean them up, we can easily just remove that /var/cache/debconf dir directly to gain 4Mb, but to be honest,  I would leave it there.

Itxaka:
IMO this is an apt-get issue becasue the Linux fundation clearly states that /var/cache can be gone at any moment and the application should be able to recover from that:

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s05.html

```
Files located under /var/cache may be expired in an application specific manner, by the system administrator, or both. The application must always be able to recover from manual deletion of these files (generally because of a disk space shortage). No other requirements are made on the data format of the cache directories.```

Itxaka (approved):

mauromorales:
yup agree, it should not break because of it. Ok I'll merge then, thanks!

PR #1891, opened 2023-10-05
Update SECURITY.md

vivekBoii:
Fixed typo error


mauromorales:
Thanks @vivekBoii, I think the `it` is still needed. We could maybe rephrase the `as` for `since`

PR #1900, opened 2023-10-09
[POC] Use a single network dracut package

Itxaka:
POC of using a single network package that loads and installs the needed module in dracut so we dont have to choose between network and network-legacy and can keep it down to just one package.

This checks the dracut module to check if it exists and if it does, it sets it as dependency of the kairos-network dracut module, so it includes the proper one on each.

This is also extensible, so we can include connman, NetworkManager and systemd-networkd at any moment.

This can be provided via luet packages and the other 2 network packages dropped altogether, so we can have just one package for all flavors, no need to manually choose and/or keep it.

Could even be included as part of the immucore dracut package so immucore deals with this directly and selects the proper network stuff, but that makes it dependent on immucore releases.

Tested with:
 - fedora
 - ubuntu
 - debian
 - opensuse-leap

All of them booted with `rd.neednet=1` and `rd.break=pre-pivot` to test inside initramfs. In all cases, the eth0 had an ip and external network connectivity.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -281,6 +281,9 @@ framework:
         END
     END
 
+    COPY kairos-network /framework/usr/lib/dracut/modules.d/29kairos-network
+    RUN mkdir -p /framework/etc/dracut.conf.d
+    COPY net.conf /framework/etc/dracut.conf.d/10-kairos-network.conf
     SAVE ARTIFACT --keep-own /framework/ framework
 
 build-framework-image:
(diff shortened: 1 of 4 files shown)


jimmykarily:
Nice! I guess this has to be move to the packages repo to be properly used right?

Itxaka:
> Nice! I guess this has to be move to the packages repo to be properly used right?

Yep!

Mainly here to test if it worked as expected!

mudler (approved):
nais!

Itxaka:
ok, seems that this had a good reception, will create a new package with it and move kairos to use it.

Issue #1897, opened 2023-10-09
Kairos dockerfiles should create full kairos images


jimmykarily:
Currently, we build the Dockerfiles, and then we [do many more things in Earthly](https://github.com/kairos-io/kairos/blob/be7961bc8086838c378839364a9f0e77861ecdbd/Earthfile#L340-L484).

All this logic can be moved to the Dockerfiles. This will allow the creation of other derivatives without any Earthly knowledge.

Items extracted from TODOs (https://github.com/kairos-io/kairos/pull/1912):

- [x] Merge Alma, Rocky and Fedora to one family Dockerfile
- [ ] Move rest of logic from Earthfile to Dockerfiles
- [x] Create a reusable workflow that simply calls docker build, passing the required build-args
- [x] Delete the earthly base-image target and force the use of the above workflow anywhere needed
- [ ] Make all remaining earthly targets to use a pre-cooked KAIROS_IMAGE rather than calling +base-image

jimmykarily:
Better merge this first: https://github.com/kairos-io/kairos/pull/1939

mauromorales:
I will better not merge since it makes it less explicit for some of the actions in which distro they should be run and in which not

PR #1912, opened 2023-10-11
1898 byoi

### LINKED ISSUE #1898: Dockefiles should separate the base image from the "build strategy" (aka flavor)

jimmykarily:
Fixes #1898


.github/workflows/image-arm.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
         git fetch --prune --unshallow
     - id: set-matrix
       run: |
-          content=`cat ./.github/flavors-arm.json`
+          content=`cat ./.github/flavors.json | jq -r 'map(select(.arch == "arm64" and .variant == "standard" and .model != "generic"))'`
```
jimmykarily, line 25: This is `get-core-matrix`. I think this should have been `.variant == "core"`.

jimmykarily, line 25:
```suggestion
          content=`cat ./.github/flavors.json | jq -r 'map(select(.arch == "arm64" and .variant == "core" and .model != "generic"))'`
```

.github/workflows/image.yaml
```diff
@@ -1,5 +1,8 @@
 name: Build and test images
 on:
+  pull_request:
+    paths:
+      - '**'
```
jimmykarily, line 5: Before we merge this:

```suggestion
```

.github/workflows/reusable-build-flavor.yaml
```diff
@@ -72,19 +90,41 @@ jobs:
...
+          # container image: quay.io/kairos/ubuntu:22.04-core-amd64-generic-v2.4.0
+          # iso: kairos-ubuntu-20.04-core-amd64-generic-v2.4.1-ipxe.iso
+          #
+          # TODO:
+          # - New branch:
+          #   - Merge alma, rocky, fedora to one family dockerfile
+          #   - Move rest of logic from Earthfile to Dockerfiles
+          #   - Create a reusable worflow that simply calls docker build passing the required build-args
+          #   - Delete the earthly base-image target and force the use of the above workflow anywhere needed
+          #   - Make all remaining earthly targets to use a pre-cooked KAIROS_IMAGE rather than calling +base-image
```
jimmykarily, line 105: Moved to the description here: https://github.com/kairos-io/kairos/issues/1897#issue-1932574939

.github/workflows/reusable-upgrade-latest-test.yaml
```diff
@@ -51,6 +54,7 @@ jobs:
           # A flag to set the download target as latest release
           # The default value is 'false'
           latest: true
+          # TODO: Until we publish with the new names, what should this be?
```
jimmykarily, line 57:
```suggestion
          # TODO: This should change after we release with the new artifact names otherwise it won't work
```

images/Dockerfile.ubuntu
```diff
@@ -244,3 +249,24 @@ RUN systemctl enable ssh
...
+#   system/suc-upgrade \
+#   system/grub2-efi \
+#   static/kairos-overlay-files \
+#   system/kcrypt \
+#   system/kcrypt-challenger \
+#   system/immucore \
+#   system/kairos-agent \
+#   init-svc/systemd \
+#   dracut/network-legacy-compat
+
```
jimmykarily, line 272:
```suggestion
```

.github/workflows/image-arm.yaml
jimmykarily: the reusable workflow that uses the matrix has the variant hardcoded to "core" anyway, that's why this worked. Will fix anyway so that we pick the right flavors.

.github/workflows/release.yaml
```diff
@@ -166,8 +164,21 @@ jobs:
       - name: Build  🔧
         env:
           FLAVOR: ${{ matrix.flavor }}
+          FLAVOR_RELEASE: ${{ matrix.flavorRelease }}
+          FAMILY: ${{ matrix.family }}
+          ARTIFACT: ${{ matrix.artifact }}
+          VARIANT: ${{ matrix.variant }}
+          MODEL: ${{ matrix.model }}
+          BASE_IMAGE: ${{ matrix.baseImage }}
```
mauromorales, line 172: I guess these can be removed (also the one above)

.github/workflows/release.yaml
```diff
@@ -166,8 +164,21 @@ jobs:
...
         run: |
-          earthly +all --VARIANT=core --FLAVOR=$FLAVOR
+          earthly +all  \
+            --VARIANT=${{ matrix.variant }} \
+            --FAMILY=${{ matrix.family }} \
+            --FLAVOR=${{ matrix.flavor }} \
+            --FLAVOR=${{ matrix.flavorRelease }} \
```
jimmykarily:
```suggestion
            --FLAVOR_RELEASE=${{ matrix.flavorRelease }} \
```

.github/workflows/release.yaml
jimmykarily: Correct! Fixed.

.github/workflows/release.yaml
jimmykarily: done

mauromorales (approved):
lgtm 👏

.github/workflows/release.yaml
```diff
@@ -94,7 +92,7 @@ jobs:
                 insecure = true
                 http = true
           EOF
-          earthly +build-framework-image --FLAVOR=${FLAVOR}
+          earthly +build-framework-image --FRAMEWORK_FLAVOR=${{ matrix.flavor }}
```
Itxaka, line 95: not valid anymore, framework images are now either generic or fips. We should just hardcode this to generic for now and we are just publishing generic images, not fips

.github/workflows/reusable-build-framework-flavor.yaml
```diff
@@ -48,6 +51,6 @@ jobs:
                 insecure = true
                 http = true
           EOF
-          earthly +build-framework-image --FLAVOR=${FLAVOR} --VERSION=master
+          earthly +build-framework-image --FRAMEWORK_FLAVOR=${FLAVOR} --VERSION=master
```
Itxaka, line 54: ditto

Earthfile
```diff
@@ -43,6 +37,15 @@ ARG K3S_VERSION
 
 all:
   ARG SECURITY_SCANS=true
+
+  ARG TARGETARCH
+  ARG --required FAMILY # The dockerfile to use
```
Itxaka, line 42: nice!

Earthfile
```diff
@@ -251,15 +251,16 @@ luet:
...
+
     WORKDIR /build
 
     COPY ./profile-build /build
     COPY +luet/luet /usr/bin/luet
 
     RUN go mod download
     COPY framework-profile.yaml /build
-    RUN go run main.go ${FLAVOR} framework-profile.yaml /framework
+    RUN go run main.go ${FRAMEWORK_FLAVOR} framework-profile.yaml /framework
```
Itxaka, line 263: this should be `generic`

Earthfile
```diff
@@ -467,23 +467,39 @@ base-image:
 
     RUN rm -rf /tmp/*
 
-    DO +CONTAINER_IMAGE_VERSION -VERSION=${OS_VERSION}
-    ARG _CIMG=$(cat IMAGE)
+    COPY ./naming.sh .
+    RUN ./naming.sh container_artifact_name
+    RUN ./naming.sh container_artifact_name > ./IMAGE
```
Itxaka, line 472: should we remove naming.sh from the base image?

Earthfile
```diff
@@ -467,23 +467,39 @@ base-image:
...
 
     # luet cleanup
     RUN luet cleanup
     RUN rm -rf /var/luet
 
+    ARG _CIMG=$(cat ./IMAGE)
```
Itxaka, line 478: are we using this IMAGE file anywhere else? Should we just make this `_CIMG=$(/naming.sh container_artifact_name)` ?

Earthfile
```diff
@@ -467,23 +467,39 @@ base-image:
...
+    ARG --required BASE_IMAGE
+    ARG --required MODEL
+    ARG --required VARIANT
+
     FROM +base-image
+
     SAVE ARTIFACT --keep-own /. rootfs
 
 uki-artifacts:
+    ARG --required FAMILY # The dockerfile to use
```
Itxaka, line 496: damn, cant we make this a global thing instead of having them in each target?

images/Dockerfile.ubuntu
```diff
@@ -143,73 +148,73 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-FROM base-ubuntu-20-lts AS hwe-20-lts
+FROM base-ubuntu-20.04 AS hwe-20.04
 RUN apt-get update && apt-get install -y --no-install-recommends \
     linux-image-generic-hwe-20.04 \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
-FROM hwe-22-lts AS amd64-ubuntu
-FROM hwe-22-lts AS amd64-ubuntu-22-lts
-FROM hwe-20-lts AS amd64-ubuntu-20-lts
+FROM hwe-22.04 AS amd64-ubuntu-23.04
```
Itxaka, line 161: 22.04 and then 23.04? Typo?

images/Dockerfile.ubuntu
```diff
@@ -143,73 +148,73 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 
-FROM amd64-non-hwe AS amd64-ubuntu-22-lts-non-hwe
-FROM amd64-non-hwe AS amd64-ubuntu-20-lts-non-hwe
+FROM amd64-non-hwe AS amd64-ubuntu-22.04-non-hwe
+FROM amd64-non-hwe AS amd64-ubuntu-20.04-non-hwe
 
-FROM base-ubuntu-22-lts AS arm64-ubuntu
-FROM base-ubuntu-22-lts AS arm64-ubuntu-22-lts
-FROM base-ubuntu-20-lts AS arm64-ubuntu-20-lts
+FROM base-ubuntu-22.04 AS arm64-ubuntu-23.04
```
Itxaka, line 173: again from 22 to 23?

images/Dockerfile.ubuntu
```diff
@@ -15,23 +17,26 @@ ARG MODEL=generic
 # the default value is empty, which means the HWE kernel WILL be installed
 # if you want to disable the HWE kernel, set HWE to "-non-hwe"
 ARG HWE=""
+ARG BASE_IMAGE=ubuntu:23.04
```
Itxaka, line 20: wait... this is making all systems come from this? Isnt like it was before much better? Also isnt it bnumping the ubuntu version at the same time?

Earthfile
```diff
@@ -604,10 +644,21 @@ iso:
 
 
 iso-uki:
+    FROM ubuntu
```
Itxaka, line 647: why ubuntu? A smaller image like alpine is better suited for this, especially if its just to run a script?

naming.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
```
Itxaka, line 1: shouldnt this go into osbuilder-tools instead of here?

Earthfile
```diff
@@ -106,42 +115,30 @@ go-deps-test:
...
+    # For naming.sh
+    ARG TARGETARCH # Earthly built-in (not passed)
+    ARG --required FAMILY
+    ARG --required FLAVOR
+    ARG --required FLAVOR_RELEASE
+    ARG --required VARIANT
+    ARG --required MODEL
+    ARG --required KAIROS_VERSION
+
+    COPY ./naming.sh .
```
Itxaka, line 136: should we have a target like the VERSION instead of copying stuff around? or a COMMAND?

Itxaka (changes requested):
some changes in regard to framework target and dockerfiles

.github/workflows/release.yaml
jimmykarily: but there is a matrix defined above (line 65) to "generic, fips". That should work fine no?

.github/workflows/reusable-build-framework-flavor.yaml
jimmykarily: This one is a "reusable" workflow. It could build the fips or the generic one. If we know we'll never build the fips framework image ever (or any other, other than the generic), we can remove all the "reusable" logic from everywhere and just have one workflow building the generic image.

images/Dockerfile.ubuntu
jimmykarily: yes, we should have no defaults. I don't remember why this default is there, maybe I screwed up in some rebase. I will remove.

images/Dockerfile.ubuntu
jimmykarily: This specific "step" `hwe-22.04` is similar both for 23.04 and 22.04 and this way we just avoid introducing an intermediate "common" step. I was also confused the first 3 times I saw this but then I got used to it :D.

jimmykarily: We put the version in the step names but keep in mind, it's only the BASE_IMAGE that defines what OS we are building on. The rest are steps to configure it and although most of the time, each version has it's own steps, sometimes they are similar. I guess we could avoid naming them with a specific version. Something like `hwe-22.04-and-23.04` would have been more appropriate but you see how this looks.

images/Dockerfile.ubuntu
jimmykarily: @mauromorales did I mess up on rebase?

naming.sh
jimmykarily: We call it in numerous places in the Earthfile and it's readily available as part of this repo. Do we have access to osbuilder-tools everywhere in all targets?

That said, we can move this script at any point.

.github/workflows/release.yaml
Itxaka: true, I though I removed this already!

images/Dockerfile.ubuntu
Itxaka: ok, I get it, still a bit confusing IMO

naming.sh
Itxaka: just wondering, as we are removing all this kind of stuff from kairos and osbuilder-tools is the one that provides a lot of stuff for this, I thought it would make sense, but indeed we can move it afterwards.

Itxaka (approved):
looks good to me!

Earthfile
jimmykarily: We could, yes. I'm hoping we'll have less earthly in the next iteration though. Should we see how it looks after this: https://github.com/kairos-io/kairos/issues/1897 and decide then?

Earthfile
```diff
@@ -278,81 +279,78 @@ framework:
 
 build-framework-image:
     FROM alpine
-    ARG FLAVOR
+
+    ARG --required FRAMEWORK_FLAVOR
```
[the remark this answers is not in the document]
jimmykarily: we'll start building the 2 framework images differently in another PR. We agreed to keep it as is in this PR.

Earthfile
jimmykarily: Good point. I'll do.

Earthfile
jimmykarily: We do in some workflows:

```
~/workspace/kairos/kairos (1898-byoi)*$ git grep "cat IMAGE"
.github/workflows/reusable-build-flavor.yaml:          IMAGE=$(cat IMAGE)
.github/workflows/reusable-build-flavor.yaml:          _IMG=$(cat IMAGE)
.github/workflows/reusable-build-provider.yaml:          _IMG=$(cat IMAGE)
```

Ideally we shouldn't include it in the images we produce but it was like that before this PR. We can try to fix it outside it.

Earthfile
jimmykarily: I'd rather not. Anything we make global becomes accessible to every target, even those that don't need it. Then you can't make it `--required` either (since it's not required by all targets, right?). It looks ugly right now but I'm hoping on the next iteration, the only thing most targets will need is the a "base image" arg which will be a pre-built image. That one will be build directly with docker to ensure that we can always build with just `docker`, no earthly required.

Earthfile
jimmykarily: alpine had no `bash`:

```
~/workspace/kairos/kairos (1898-byoi)*$ docker run -it --rm alpine /bin/bash
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/bin/bash": stat /bin/bash: no such file or directory: unknown.
```

We could implement the script in pure `sh` but we opted for an image that has bash.

Earthfile
jimmykarily: Done (right below)

PR #1924, opened 2023-10-19
:art: Build alpine initrd during image building

Itxaka:
Instead of using a pre-packaged one we can just build it during image build like we do on dracut systems

Fixes #


jimmykarily (approved):
Awesome! Thanks for the quick fix.

PR #1926, opened 2023-10-19
[POC] Use single framework image

Itxaka:
Uses a single framework image build under https://github.com/Itxaka/kairos-framework with all the artifacts on it.

The only difference is that the fips image is a separated one.

Works around the systemd starting sysinit services by removing those if we are on systemd.


Itxaka:
The openrc services issue can be easily fixed by moving those services to be added during initramfs by yip if we are on a openrc system, there is no need to have them in the image directly....

PR #1929, opened 2023-10-20
:art: Use a single framework image

### LINKED ISSUE #1896: There should be only one framework image with the logic for all distros
Currently we are pushing different framework images as tags of quay.io/kairos/framework:XXX .
We want to move the installation of the framework bits from Earthly to the dockerfiles themselves. It makes it simpler if there is only one framework image and not one per flavor and our users also won't have to worry when creating their own derivatives, what image they should use.

It seems that this is possible but maybe there is some logic that need to be added to the framework bits.

Expected outcome:

There is only one framework image that works for all flavors.

Part of: https://github.com/kairos-io/kairos/issues/116

Itxaka:
With changes to the packages we now generate the proper service files on the fly so we dont need to have those package separated.

We now have 2 framework images, a generic one used for ALL flavors and a fips one, only provided for external consumers as we dont publish any fips enabled artifacts.

We also bring both the systemd and openrc package together as they dont affect each other anymore so we can have a single framework.

This patch also removes the service enablement as its not needed anymore, it done via config files

Reworks the framework jobs to just build and publish the 2 different framework image (generic and fips)

Fixes #1896


jimmykarily (approved):

mudler:
Do we need to update the docs here https://kairos.io/docs/reference/build-from-scratch/ ?

PR #1939, opened 2023-10-23
Rely less on earthly and more on pure docker instructions

Itxaka:
So they can easily be moved into the dockerfiles

Fixes #


Earthfile
```diff
@@ -407,63 +402,34 @@ base-image:
...
-                RUN kernel=$(ls /boot/zImage-* | head -n1) && if [ -e "$kernel" ]; then ln -sf "${kernel#/boot/}" /boot/vmlinuz; fi
-            END
-        ELSE
-            # Debian has vmlinuz-VERSION
-            RUN kernel=$(ls /boot/vmlinuz-* | head -n1) && if [ -e "$kernel" ]; then ln -sf "${kernel#/boot/}" /boot/vmlinuz; fi
-            RUN kernel=$(ls /boot/Image-* | head -n1) && if [ -e "$kernel" ]; then ln -sf "${kernel#/boot/}" /boot/vmlinuz; fi
-        END
-    END
+    # Delete not needed initramfs files
+    RUN rm -rf /boot/initrd.img-* || true
```
Itxaka, line 418: always do and not fail if its not there, we dont care

Earthfile
```diff
@@ -407,63 +402,34 @@ base-image:
...
+    RUN rm -rf /boot/initramfs-* || true
+    RUN --no-cache kernel=$(ls /lib/modules | head -n1) && depmod -a "${kernel}"
+    # Recreate initrd and link it to /boot/initrd
+    RUN --no-cache if [ -f "/usr/bin/dracut" ]; then kernel=$(ls /lib/modules | head -n1) && dracut -f "/boot/initrd-${kernel}" "${kernel}" && ln -sf "initrd-${kernel}" /boot/initrd;fi
```
Itxaka, line 422: check in time for initrd builder

Earthfile
```diff
@@ -407,63 +402,34 @@ base-image:
...
+    RUN --no-cache if [ -f "/sbin/mkinitfs" ]; then kernel=$(ls /lib/modules | head -n1) && mkinitfs -o /boot/initrd $kernel; fi
+
+    # Create a symlink to the kernel to /boot/vmlinuz
```
Itxaka, line 425: just do all the linking here if the sources exists

mauromorales (approved):
nice! this should make the extraction to the dockerfile easier

Earthfile
```diff
@@ -407,63 +402,34 @@ base-image:
...
+    RUN --no-cache if [ -e "/boot/vmlinuz-lts" ]; then ln -sf /boot/vmlinuz-lts /boot/vmlinuz; fi
+    RUN --no-cache if [ -e "/boot/vmlinuz-rpi4" ]; then ln -sf /boot/vmlinuz-rpi4 /boot/vmlinuz; fi
+    RUN --no-cache if [ -e " /boot/Image" ]; then ln -sf  /boot/Image /boot/vmlinuz; fi
+    RUN --no-cache if [ -e "/boot/zImage" ]; then ln -sf /boot/zImage /boot/vmlinuz; fi
+    RUN --no-cache kernel=$(ls /boot/zImage-* 2> /dev/null | head -n1) && if [ -e "$kernel" ]; then ln -sf "${kernel#/boot/}" /boot/vmlinuz; fi
+    RUN --no-cache kernel=$(ls /boot/vmlinuz-* 2> /dev/null | head -n1) && if [ -e "$kernel" ]; then ln -sf "${kernel#/boot/}" /boot/vmlinuz; fi
+    RUN --no-cache kernel=$(ls /boot/Image-* 2> /dev/null | head -n1) && if [ -e "$kernel" ]; then ln -sf "${kernel#/boot/}" /boot/vmlinuz; fi
```
jimmykarily, line 432: Eventually we will have to move these to the dockerfiles and it would be good to remember which line is there for which flavor. Maybe keep that information around as comments? We can always look back in this commit but I'm afraid more changes in the future might make it hard to discover.

It should be too long until we move them though, so maybe it's fine as it is. I'm not sure.

jimmykarily (approved):

Earthfile
Itxaka: Is it? Or should we do ALL the lines and then fail if there is no link created? 

I mean, I worked them so they wont run unless it finds the file so they are mostly non-op and only one triggers.

This means we can add all of them to all the dockerfiles and they will work everywhere. Would be nicer if it was some script or whatever, but as pure docker goes, this should be valid for all the flavors in existance and should not fail anywhere.

jimmykarily: we can wrap these to a script you are right. But were does this script belong? Maybe it should be in the framework image. If we don't want it in the final images, we can just delete it in the dockerfiles after running it.

jimmykarily:
This is part of: https://github.com/kairos-io/kairos/issues/1897

E.g. if we move the rest of the steps to a script and we run it inside the Dockerfiles, 1897 is pretty much done.

jimmykarily:
Let keep this open until we investigate alternative ways (e.g. put each distro logic to its own Dockerfile, even if there is some duplication)

jimmykarily:
Closing in favor of: https://github.com/kairos-io/kairos/issues/1897

PR #1951, opened 2023-10-25
Fix debian repository

jimmykarily:
for some reason the CDN mirrors are not working

### DIFF
--- a/images/Dockerfile.debian
+++ b/images/Dockerfile.debian
@@ -11,7 +11,7 @@ ARG BASE_IMAGE=debian:testing
 # we use testing for amd64 because zfsutils-linux is not available in bookworm
 FROM $BASE_IMAGE AS amd64-upstream
 RUN <<EOF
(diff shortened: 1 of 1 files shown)


images/Dockerfile.debian
```diff
@@ -11,7 +11,7 @@ ARG BASE_IMAGE=debian:testing
 # we use testing for amd64 because zfsutils-linux is not available in bookworm
 FROM $BASE_IMAGE AS amd64-upstream
 RUN <<EOF
-echo "deb http://deb.debian.org/debian bookworm main contrib" >> /etc/apt/sources.list.d/kairos.list
+echo "deb https://ftp.debian.org/debian bookworm main contrib" >> /etc/apt/sources.list.d/kairos.list
```
mauromorales, line 14: I wonder what is better long term, since this will stay as part of the configuration of the debian images released

jimmykarily: It will be there but since the system can't be upgraded and packages can't be installed (read-only fs) it doesn't matter. It's only used once when building the image.

Itxaka (approved):

PR #1953, opened 2023-10-25
Introduce RHEL family

mauromorales:
- merges the files for rockylinux and almalinux into a rhel dockerfile
- decided not to add fedora into this since, it requires to add the epel packages, which might make fedora too close to rhel for our community users. But wonder what you think, is it a good/bad idea to try to also merge that one?

Clean up fedora dockerfile

- remove packages which are already installed by default "coreutils" and "curl"
- make it similar to the other dockerfiles for easier maintenance


Earthfile
```diff
@@ -435,7 +436,7 @@ base-image:
 
     # Set /boot/vmlinuz pointing to our kernel so kairos-agent can use it
     # https://github.com/kairos-io/kairos-agent/blob/0288fb111bc568a1bfca59cb09f39302220475b6/pkg/elemental/elemental.go#L548   q
-    IF [ "$FLAVOR" = "fedora" ] || [ "$FLAVOR" = "rockylinux" ] || [ "$FLAVOR" = "almalinux" ]
+    IF [ "$FLAVOR" = "fedora" ] || [ "$FAMILY" = "rhel" ]
```
Itxaka, line 439: I would just do a `RUN rm -rf /boot/initramfs-* || true` so it always runs instead of having switches here so we just try to remove it always and not fail.

Itxaka:
> decided not to add fedora into this since, it requires to add the epel packages, which might make fedora too close to rhel for our community users. But wonder what you think, is it a good/bad idea to try to also merge that one?

IMHO I would merge all of them. What packages come from the epel repos? Only the needed ones like zfs? I think merging them all under the same family would be ok as in the end they will have the minimal packages required for a base Kairos flavor, but they can expand it with whatever afterwards, so they can also drop the epel repos on their custom dockerfiles no?

Itxaka (approved):

mauromorales:
I tried doing a pure rhel but it's a lot more work :( lots of packages are not available, not sure if they are named differently or if I need to enable some repo

images/Dockerfile.rhel
```diff
@@ -1,10 +1,23 @@
...
 
 RUN echo "install_weak_deps=False" >> /etc/dnf/dnf.conf
 
+# Install packages according to distro
+# epel-packages https://docs.fedoraproject.org/en-US/epel/
+FROM base AS rockylinux
 RUN dnf install -y epel-release && dnf clean all
+RUN dnf install -y "https://zfsonlinux.org/epel/zfs-release-2-3$(rpm --eval "%{dist}").noarch.rpm" && dnf clean all
+
+FROM rockylinux AS almalinux
```
Itxaka, line 14: This breaks my head, can't we name it in a more intuitively way?

Itxaka (approved):

images/Dockerfile.rhel
mauromorales: good point, I added an `epel` stage and both alma and rocky will be based on that one

PR #1957, opened 2023-10-25
Merge opnesuse families into one

mauromorales:


Itxaka (commented):
do we even need the family anymore?

Let me expand on this. It seems that with this patch, the family and the flavor are the same in all but the nvidia board, and Im not even sure that needs a specific family when it has the model stuff no? And in fact it should be of the same family as the ubuntu, just it will have a different path I guess...

mauromorales:
@Itxaka IMO we still need the family, first because it lets us merge distros like fedora, alma and rocky into a rhel family. Second because it allows the community to try Kairosifying other distros we don't really support, for example if you use Linux Mint or elementary as your base image, you can use the Ubuntu family, if you have PostmarketOS as your base image you can use the alpine family.

mauromorales:
for the nvidia one, indeed it could be part of Ubuntu, we just felt that this is so unique and probably will not change often, so it was more at risk of braking if merged with the Ubuntu dockerfile. Specially because for the orin it only works with one version of ubuntu even.

jimmykarily (approved):

Itxaka:
> @Itxaka IMO we still need the family, first because it lets us merge distros like fedora, alma and rocky into a rhel family. Second because it allows the community to try Kairosifying other distros we don't really support, for example if you use Linux Mint or elementary as your base image, you can use the Ubuntu family, if you have PostmarketOS as your base image you can use the alpine family.

Thanks for the explanation!

Itxaka (approved):

PR #1965, opened 2023-10-27
Remove AlmaLinux from building

mauromorales:
Now that we have the families to build, we can kinda trust that maintaining the rocky flavor will automatically give us alma.

Would be great if we can build everything all the time but it's expensive and I don't see any users relying on Alma, plus they can still build on their own with our recipe

### DIFF
--- a/.github/flavors.json
+++ b/.github/flavors.json
@@ -340,26 +340,6 @@
         "worker": "self-hosted"
     },
     {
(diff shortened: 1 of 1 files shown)


mauromorales:
cancelling jobs since they are not testing alma

Itxaka (approved):

Issue #1816, opened 2023-09-14
Rpi3 artifacts


Ludea:
**Is your feature request related to a problem? Please describe.**

There is no binaries for RP3

**Describe the solution you'd like**

Build, and upload RPI3 assets on quay.
Add RPI3 section https://kairos.io/docs/reference/image_matrix/

**Describe alternatives you've considered**

Let user build Kairos Rpi3 img

**Additional context**

Separate Rpi3 img has added since 2.4 release

jimmykarily:
Possibly fixed here: https://github.com/kairos-io/kairos/pull/1820

jimmykarily:
Blocked by this: https://github.com/kairos-io/kairos/issues/1694

Issue #1527, opened 2023-06-19
:seedling: Differentiate images automatically tested by our CI, and images that are "experimental" or manually tested


mudler:
**Is your feature request related to a problem? Please describe.**

Currently there are a bunch of flavors/images that we don't test automatically.

This means that, we either test them manually, or rely on user-report to spot bugs.

This approach doesn't scale, and we have discussed the possibility to add automated test, however, this is not a solution for ARM platform images (as RPI or Jetson) neither for certain flavors (rockylinux, fedora, that requires specific cpu settings).

**Describe the solution you'd like**

We should clearly differentiate in our documentation and in our build artifacts (for example, by using a suffix?) images that are tested automatically by our QA pipelines and those which aren't.

We can, identify the images in two big categories: 
- Standard builds
- Community builds

**Describe alternatives you've considered**
Try to cover as much as possible with the CI. however, we are trying to find a general solution that works in most cases, and that doesn't scale with ARM specific images.

## Action items

- [ ] Identify flavors that are automatically tested
- [ ] Reflect in the docs (Image matrix, .... ) which are official and which not
- [ ] Change the build artifacts naming to reflect the change.
- [ ] Write a blog post about this change

mauromorales:
Added action item to document the testing matrix flavors reasoning

PR #1966, opened 2023-10-27
build RPI3 artifacts

Ludea:
Build Rpi3 artifacts

Fixes https://github.com/kairos-io/kairos/issues/1816


mauromorales:
@jimmykarily @mudler I'm not sure how to decide on this one. On the one hand I think it's a good idea to have as many artifacts as possible so users can easily test Kairos, on the other hand this expands the time our CI will be running and besides wait time, we also see issues on release API hits. Also if we plan to have so many ARM builds, it might be a good idea to also consider ARM workers.

IMO the ideal scenario would be to know which artifacts are most interesting for users to consume and just build those, but we don't have any numbers so I guess we need to base the decision differently. Also soon there will also be rpi5 so we will end up with 3 times the artifacts for arm that we originally had, but we don't really know which of the boards is actually mostly used at any point in time.

This might be related to https://github.com/kairos-io/kairos/issues/1527

.github/flavors.json
```diff
@@ -248,6 +348,16 @@
         "baseImage": "alpine:3.19",
         "arch": "arm64",
         "worker": "ubuntu-latest"
+    },
+        {
+        "family": "alpine",
+        "flavor": "alpine",
+        "flavorRelease": "3.18",
```
mauromorales, line 355:
```suggestion
        "flavorRelease": "3.19",
```

.github/flavors.json
```diff
@@ -248,6 +348,16 @@
...
+        "variant": "standard",
+        "model": "rpi3",
+        "baseImage": "alpine:3.18",
```
mauromorales, line 358:
```suggestion
        "baseImage": "alpine:3.19",
```

.github/flavors.json
```diff
@@ -258,6 +368,16 @@
         "baseImage": "alpine:3.19",
         "arch": "arm64",
         "worker": "ubuntu-latest"
+    },
+        {
+        "family": "alpine",
+        "flavor": "alpine",
+        "flavorRelease": "3.18",
```
mauromorales, line 375:
```suggestion
        "flavorRelease": "3.19",
```

.github/flavors.json
```diff
@@ -258,6 +368,16 @@
...
+        "variant": "core",
+        "model": "rpi3",
+        "baseImage": "alpine:3.18",
```
mauromorales, line 378:
```suggestion
        "baseImage": "alpine:3.19",
```

mudler:
> @jimmykarily @mudler I'm not sure how to decide on this one. On the one hand I think it's a good idea to have as many artifacts as possible so users can easily test Kairos, on the other hand this expands the time our CI will be running and besides wait time, we also see issues on release API hits. Also if we plan to have so many ARM builds, it might be a good idea to also consider ARM workers.
> 
> IMO the ideal scenario would be to know which artifacts are most interesting for users to consume and just build those, but we don't have any numbers so I guess we need to base the decision differently. Also soon there will also be rpi5 so we will end up with 3 times the artifacts for arm that we originally had, but we don't really know which of the boards is actually mostly used at any point in time.
> 
> This might be related to #1527

let's at least restrict the number of flavors we are going to support for RPI3 to the ones that are actually manually tested. Let's identify 1, or 2 of them that we can support for now. I'd tempted to say let's go alpine/openSUSE as alpine is small and openSUSE is well tested in our CIs (for amd64, but still). The other solution would be to use self-hosted runners and expand our CI's workers, but we might need to adapt the pipelines to make it work on self-hosted.

@Ludea would be fine with you?

Ludea:
Perfect for me, I would like to use alpine

mauromorales (approved):
fantastic, thanks @Ludea for the contribution and @mudler for chipping in.

.github/flavors.json
```diff
@@ -38,6 +48,16 @@
         "baseImage": "opensuse/leap:15.5",
         "arch": "arm64",
         "worker": "ubuntu-latest"
+    },
+        {
```
mauromorales, line 52:
```suggestion
    },
    {
```

.github/flavors.json
```diff
@@ -78,6 +98,16 @@
         "baseImage": "opensuse/tumbleweed",
         "arch": "arm64",
         "worker": "ubuntu-latest"
+    },
+        {
```
mauromorales, line 102:
```suggestion
    },
    {
```

.github/flavors.json
```diff
@@ -88,6 +118,16 @@
...
```
mauromorales, line 122:
```suggestion
    },
    {
```

.github/flavors.json
```diff
@@ -118,7 +158,7 @@
         "baseImage": "ubuntu:20.04",
         "arch": "arm64",
         "worker": "ubuntu-latest"
-    },
+    },
```
mauromorales, line 161:
```suggestion
    },
```

PR #1970, opened 2023-10-27
Fix arm generic job

Itxaka:

Fixes #

### DIFF
--- a/.github/workflows/reusable-image-and-iso-arm-generic.yaml
+++ b/.github/workflows/reusable-image-and-iso-arm-generic.yaml
@@ -6,6 +6,18 @@ on:
       flavor:
         required: true
         type: string
+      flavor_release:
+        required: true
+        type: string
+      family:
+        required: true
+        type: string
+      model:
+        required: true
+        type: string
+      base_image:
+        required: true
+        type: string
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-image-and-iso-arm-generic.yaml
```diff
@@ -35,15 +47,21 @@ jobs:
...
+              --FLAVOR=${{ inputs.flavor }} \
+              --FLAVOR_RELEASE=${{ inputs.flavor_release }} \
+              --BASE_IMAGE=${{ inputs.base_image }} \
+              --MODEL=${{ inputs.model }} \
+              --VARIANT=core \
+              --IMG_COMPRESSION=${{env.IMG_COMPRESSION}}
       - name: Push to quay
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+        #if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+        # disable artifact upload for now
```
jimmykarily, line 63: what for? Either we disabled it for good or we should write here when this is supposed to be enabled back. If we don't want "master" artifacts, let's just remove this from here and check that the release pipeline will produce something when we tag a new release.

jimmykarily (approved):

.github/workflows/reusable-image-and-iso-arm-generic.yaml
Itxaka: mainly because I have no idea under what repo should this go or how xDD

jimmykarily: ok we can fix it later but the flags we pass to the dockerfile/earthly are determining the artifact name. These are the same args we pass to the naming.sh script. `naming.sh` will tell you where it should be pushed if you instruct it to give you a "container_artifact_name".

Issue #1975, opened 2023-10-30
Move framework images to their own repo


jimmykarily:
Framework images should have their own lifecycle so that we can just copy the latest inside the dockerfiles.

Part of: #1897

PR #2024, opened 2023-11-24
Build and release multi arch framework images

mauromorales:
currently we only release amd64 framework images. Making them multi arch allows us to take advantage of the pre-built frameworks in arm. this will be particularly useful for the extraction of the earthly logic into dockerfiles related to #1897

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -264,7 +264,7 @@ luet:
 framework:
     FROM golang:alpine
 
-    ARG FRAMEWORK_FLAVOR
+    ARG --required SECURITY_PROFILE
 
     WORKDIR /build
 
@@ -273,7 +273,7 @@ framework:
 
     RUN go mod download
     COPY framework-profile.yaml /build
-    RUN go run main.go ${FRAMEWORK_FLAVOR} framework-profile.yaml /framework
+    RUN go run main.go ${SECURITY_PROFILE} framework-profile.yaml /framework
 
     RUN mkdir -p /framework/etc/kairos/
     RUN luet database --system-target /framework get-all-installed --output /framework/etc/kairos/versions.yaml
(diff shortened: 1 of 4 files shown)


Earthfile
```diff
@@ -290,29 +295,41 @@ framework:
 
     SAVE ARTIFACT --keep-own /framework/ framework
 
+multi-build-framework-image:
+    ARG --required SECUIRTY_PROFILE
+
+    BUILD --platform=linux/amd64 --platform=linux/arm64 +build-framework-image
```
jimmykarily, line 301: you can do that?! :D

Earthfile
```diff
@@ -290,29 +295,41 @@ framework:
...
+        ARG _SECUIRTY_PROFILE=generic
     END
 
+
     COPY +version/VERSION ./
     ARG VERSION=$(cat VERSION)
 
-    ARG _IMG="$IMAGE_REPOSITORY_ORG/framework:${VERSION}_${FRAMEWORK_FLAVOR}"
+    IF [ "$VERSION" ~= ".*dirty.*" ]
+        ARG FRAMEWORK_VERSION=master
```
jimmykarily, line 318: we will probably revisit this when we start this: https://github.com/kairos-io/kairos/issues/1975 but shouldn't we let the caller pass the version down? Matching `dirty` feels...well, dirty :D 

Feel free to ignore for now, this will change on the #1975 anyway.

Earthfile
mauromorales: yup, they do have some niiiice fetures :P

jimmykarily (approved):

Earthfile
mauromorales: yeah right now I thought there would be only 2 options, master or a release tag ... but indeed lets evaluate if there's a better way to do it

Itxaka:
late to this, but pretty cool @mauromorales !

Issue #1152, opened 2023-03-16
:seedling: spike: lifecycle for static measured boot


mudler:
try to have static measured boot with our UKI images manually by taking notes of the steps

mudler:
Blocked by #1150

jimmykarily:
Desired output of this spike:

- steps that reproduce the lifecycle even if it requires manual commands and such (will be later implemented through kairos)
- some notes|docs|images that describe the process

Itxaka:
Missing:
 - Upgrade path

Potential problems:
 - signatures changing...no boot
 - New TPM measurements have to be added to the luks volumes (so key0 would be the current system, key1 the new one) to be able to unlock them after upgrade. This is the one that has the most potential to failure. Not sure if we can enroll an extra key without the original password. So this may be a blocker.
The thing is, we first create the luks partition with a random key, add the TPM mesurements as an extra key to unlock it and then remove the password from the luks partition so its only unlockable via TPM measurements. But Im not sure the enroll key feature from systemd-supports using TPM as a key to add a new key. If it only accepts a password to enroll a new key we are screwed.,

Itxaka:
workflow that builds our uki with measurements: https://github.com/kairos-io/kairos/pull/2028
This builds an iso with systemd-boot (bootloader), and our uki file, a loader.conf (for the bootloader) and a kairos-VERSION.conf (for the bootloader as well)
This also has the posibility of building oci artifacts used for upgrade or potentially int he future, to generate isos with auroraboot
All files are signed with our test keys.
Files are also measured.
Uki isos are tested via our test workflow for correct use.
The workflow uses a firmware with SECUREBOOT enabled to test that we are signing them correctly.
The tests check for secureboot to be enabled.
This uses our TEST keys and TEST signatures for signing and measurements. THIS ARETIFACTS ARE NOT TO BE USED ANYWHERE BUT ON TESTING. Those keys are currently public so using those artifacts for anything other that development is a huge security issue.
Release workflow is to be discussed and implemented as we need to agree how to manage our keys.
Release workflow for end users to generate their isos/artifacts with private keys is to be discussed.



On agent, during uki install do via a subcommand `kairos-agent uki install` the code is here: https://github.com/kairos-io/kairos-agent/blob/main/pkg/uki/install.go
 - Creates only 3 partitions (EFI, OEM, PERSISTENT)
 - Currently hardcoded to mount the ISO via the ISO label to use it as source for the install.
[...]

PR #2028, opened 2023-11-27
[UKI] Measure

Itxaka:
Part of https://github.com/kairos-io/kairos/issues/1152


Earthfile
```diff
@@ -456,7 +456,7 @@ uki-tools-image:
 # This is for easy testing SecureBoot locally for development purposes
 # Installing this keys in other place than a VM for testing SecureBoot is irresponsible
 uki:
-    FROM ubuntu
+    FROM +uki-tools-image
```
mauromorales, line 459: I think this is the only mention of `uki-tools-image` target, should we get rid of the target?

build_uki.sh
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env bash
```
mauromorales, line 1: i miss where this file is being consumed

Itxaka: Currently a manual process to generate the images as I'm still working on it, order, retouches and also needs an extra path to generate the artifacts

mauromorales (approved):
lgtm

Earthfile
```diff
@@ -455,79 +436,149 @@ uki-tools-image:
...
+    SAVE ARTIFACT /output/efi efi
+
+# This is the final artifact, only the files on it
+uki-image:
+    ARG --required FAMILY
+    COPY +version/VERSION ./
+    ARG KAIROS_VERSION=$(cat VERSION)
+    FROM scratch
+    COPY +uki-image-artifacts/efi /
+    SAVE IMAGE --push $IMAGE_REPOSITORY_ORG/${FAMILY}:${KAIROS_VERSION}-uki
```
mauromorales, line 529: yup as discussed you should be able to compose the new name with `naming.sh`

```
    ARG OS_NAME=kairos-${VARIANT}-${FLAVOR}-${FLAVOR_RELEASE}
    RUN KAIROS_VERSION=$(cat ./VERSION) \
        OS_VERSION=$(cat ./VERSION) \
        OS_LABEL=$(cat ./VERSION) \
        OS_LABEL=$(naming.sh container_artifact_label) \
        OS_REPO=$(naming.sh container_artifact_repo) \
        ARTIFACT=$(naming.sh bootable_artifact_name) \
        envsubst >>/etc/os-release </usr/lib/os-release.tmpl
    RUN KAIROS_VERSION=$(cat ./VERSION) naming.sh container_artifact_name > /IMAGE

    ARG _CIMG=$(cat ./IMAGE)
```

mauromorales: we used to have a command for this, but since for a little while it was done in the dockerfile, I removed it, maybe I should bring it back, but we can do that after this PR

mauromorales (commented):
just the issue with the naming which you already brought up, other than that lgtm

Earthfile
jimmykarily: We will soon replace the naming.sh script with versioneer (though the standalone cli or kairos-agent). If you find `naming.sh` needing any changes, let's coordinate first.

tests/keys/private.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
jimmykarily, line 1: Are we signing with a publicly available private key? We are not pushing the uki artifact when we release right?

Itxaka: We currently are not pushing anything anywhere.

This is for testing mainly and done with public private keys and certs, all of them used in the uki test pipeline.

Final releases would need a specific workflow to sign and measure them with our private keys.

But that is like a whole other conversation on how to sign them, where, who has the key, how we cycle them, etc..

Not willing to open that discussion right now as it's a big one 🤣

jimmykarily: Sure, I'm just making sure I got this right.

Earthfile
Itxaka: yes indeed, we could have the uki as part of the name maybe in the versioneer? no idea where, if as the model or something (so generic and uki) ?

Itxaka:
tested and it seems to work on fedora >=38 and ubuntu >= 23.10
tumbleweed supposedly should work but I was not able to make it work somehow and had very weird issues.

Itxaka:
cancelled everything but the uki job

Earthfile
```diff
@@ -367,79 +348,158 @@ uki-tools-image:
 # NOTE: NEVER EVER EVER use this keys for signing anything that its going outside your computer
 # This is for easy testing SecureBoot locally for development purposes
 # Installing this keys in other place than a VM for testing SecureBoot is irresponsible
-uki:
```
jimmykarily, line 370: I'm also making some changes on this. We may get some conflicts: https://github.com/kairos-io/kairos/pull/2081/files#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R367

Itxaka:
This should be finished.

A big change here is that we are moving to use the "main" framework image. That ok for you @mauromorales @jimmykarily @mudler ? I could also cut a new version once @jimmykarily has merged the versioneer stuff, but currently the main framework version has a kairos-agent RC version, so it feels bad to cut a release with RC versions.

jimmykarily:
> This should be finished.
> 
> A big change here is that we are moving to use the "main" framework image. That ok for you @mauromorales @jimmykarily @mudler ? I could also cut a new version once @jimmykarily has merged the versioneer stuff, but currently the main framework version has a kairos-agent RC version, so it feels bad to cut a release with RC versions.

Oh I forgot we now need to build a new framework image in order to get a new kairos-agent tested. My changes on the kairos-agent should be done. I tagged another rc this morning but it's still building the arm package: https://github.com/kairos-io/packages/actions/runs/7260525990/job/19779962604

If you can to tag kairos-agent as stable and release a framework image with that, I'm fine with it. It will help me test my PR too.

Earthfile
```diff
@@ -9,7 +9,7 @@ ARG LUET_VERSION=0.35.0
 # renovate: datasource=docker depName=aquasec/trivy
 ARG TRIVY_VERSION=0.47.0
 # renovate: datasource=github-releases depName=kairos-io/kairos-framework
-ARG KAIROS_FRAMEWORK_VERSION="v2.4.4"
+ARG KAIROS_FRAMEWORK_VERSION="main"
```
jimmykarily, line 12: I don't think we override this in release pipelines. This means we will be releasing from "main" images. Is it needed only because of the kairos-agent rc? Let's cut a stable one then.

mauromorales:
> This should be finished.
> 
> 
> 
> A big change here is that we are moving to use the "main" framework image. That ok for you @mauromorales @jimmykarily @mudler ? I could also cut a new version once @jimmykarily has merged the versioneer stuff, but currently the main framework version has a kairos-agent RC version, so it feels bad to cut a release with RC versions.

For me it's ok to move to main but in that case I don't see the value of producing versions for the framework images. I think relying on main can be tricky because what do we release against? It means that either we always have a moving target or that master tests nightly while release build a version. What I mean is that if we consider these issues not to cause many problems then it's a go for me. But for a different component that moves faster I would say it's better not to.

mudler:
what maybe makes sense is having releases using tagged framework images, while main can just consume main (so can be used for development)

Itxaka:
bumped to use latest framework released version.

jimmykarily (approved):

PR #2061, opened 2023-12-06
:sparkels: Add labels to kairos container images

mauromorales:
The only way to distinguish our images right now is via the tag, however, this can be problematic for many reasons e.g. they don't allow all valid characters in semver versions. Also it's very common that one names the image something simple to work locally and then you might forget certain attributes. By adding the labels we would address these issues


Itxaka (approved):
Noice

images/Dockerfile.kairos
```diff
@@ -47,11 +47,25 @@ RUN rm -rf /etc/machine-id
...
+LABEL org.opencontainers.image.authors="Kairos Maintainers <team@kairos.io>"
+LABEL org.opencontainers.image.licenses="Apache-2.0"
+LABEL org.opencontainers.image.url="https://github.com/kairos-io/kairos"
+LABEL org.opencontainers.image.source="https://github.com/kairos-io/kairos.git"
+LABEL io.kairos.base-image=$BASE_IMAGE
+LABEL io.kairos.variant=$VARIANT
+LABEL io.kairos.family=$FAMILY
+LABEL io.kairos.flavor=$FLAVOR
+LABEL io.kairos.flavor_release=$FLAVOR_RELEASE
+LABEL io.kairos.version=$VERSION
```
jimmykarily, line 65: Maybe the model too?

jimmykarily (approved):

images/Dockerfile.kairos
mauromorales: good catch, thanks!

PR #2063, opened 2023-12-07
Fix release of framework image

mauromorales:
silly me wanted to have a version check originally by using just regex and of course didn't think of all the issues that would happen with release candidates or other semver variations. With this change, the default is master, if the consumer passes "git" then the git version is used (useful for releases). If the consumer passes anything else, then that value is used as framework image (useful for development or other special cases)


Itxaka (approved):
we need to move asap to the other repo for framework stuff xD

PR #2067, opened 2023-12-08
Re-enable provider-upgrade test

mauromorales:
And remove descentralized tests since they alwasy fail


.github/workflows/image.yaml
```diff
@@ -303,19 +303,20 @@ jobs:
...
-          #  flavorRelease: "leap-15.5"
-          #  label: "provider-upgrade"
           - flavor: "opensuse"
             flavorRelease: "leap-15.5"
-            label: "provider-decentralized-k8s"
+            label: "provider-upgrade"
+          # no point of running this on CI if it always fails
+          # - flavor: "opensuse"
+          #   flavorRelease: "leap-15.5"
+          #   label: "provider-decentralized-k8s"
```
mauromorales, line 312: Ok I get this one might be controversial but seriously, if they always fail what is the point of running them? The real solution is to fix them but that can be done on a PR I guess so I do think we should remove them from master, it just makes it harder to track when something actually broke on master

jimmykarily: I'm fine to comment them out and open a ticket to fix them (so we don't forget it). Preferably a high prio one. They've been broken for too long.

jimmykarily (approved):

PR #2074, opened 2023-12-11
Use framework images built on new repo

### LINKED ISSUE #1975: Move framework images to their own repo

mauromorales:
Fixes #1975


images/Dockerfile.kairos
```diff
@@ -1,6 +1,8 @@
 # WARNING: Do not build this image on its own, use the different Dockerfile.kairos-* instead
 FROM --platform="linux/${TARGETARCH}" quay.io/kairos/framework:${FRAMEWORK_VERSION}_${SECURITY_PROFILE} AS framework
 
+FROM quay.io/luet/base:0.35.0 AS luet
```
mauromorales, line 4: This is the only major change, since luet used to be part of the old framework images and now it will be only added until the kairos part of the factory. Any opinions on that?

images/Dockerfile.kairos-alpine
```diff
@@ -184,15 +188,6 @@ ARG MODEL
 ARG REGISTRY_AND_ORG="quay.io/kairos"
 ARG K3S_VERSION
 ARG TARGETARCH
-ARG OS_NAME=kairos-${VARIANT}-${FLAVOR}-${FLAVOR_RELEASE}
```
mauromorales, line 187: these changes were just because the files were not updated with the latest changes in the Dockerfile.kairos image. I will try to get back to the base images ticket to avoid having this issues

Itxaka (approved):
Nice. I wonder if we should provide luet in the framework image if it's a dependency. After all we add all Kairos bits and bops by this so let feels like it should be bundled

Earthfile
```diff
@@ -334,7 +257,7 @@ base-image:
     ARG KAIROS_VERSION=$(cat ./GIT_VERSION)
 
     IF [ "$FRAMEWORK_VERSION" = "" ]
-        ARG _FRAMEWORK_VERSION=master
+        ARG _FRAMEWORK_VERSION=main
     ELSE IF [ "$FRAMEWORK_VERSION" = "git" ]
         ARG _FRAMEWORK_VERSION=$VERSION
```
jimmykarily, line 262: This is not going to work anymore, is it?

images/Dockerfile.kairos
jimmykarily: You mean it won't be in the final images anymore right? Hm, I see we shell out to `luet` in the kairos-sdk (`git grep luet`). Maybe things will break if it's not there?

jimmykarily: Hm no, it's the final image afaict. It should be fine then.

jimmykarily (approved):

Earthfile
mauromorales: Shoot, but of course not, great catch!

mauromorales: @jimmykarily ok I made it so it uses the latest stable

images/Dockerfile.kairos
mauromorales: ok, moved it back to the framework image

PR #2073, opened 2023-12-11
:arrow_up: Update repositories and adapt `/etc/os-release` for the updated kairos-agent and provider-kairos binaries (#1999)

ci-robbot:
Bump of Kairos repositories
--------------------------
> [\!IMPORTANT]
> Full package list from new repo
```yaml
- name: edgevpn
  category: utils
  version: 0.24.6
- name: k3s-systemd
  category: k8s
  version: 1.28.2+1
- name: k9s
  category: utils
  version: 0.28.0
- name: kube-vip
  category: utils
  version: 0.6.3
- name: kubectl
  category: container
  version: 1.28.3
- name: nerdctl
  category: utils
  version: 1.7.0
- name: provider-kairos
  category: system
  version: 2.6.0
```

PR #2081, opened 2023-12-14
1999 - Update repositories and adapt /etc/os-release to versioneer

### LINKED ISSUE #1999: :bug: Adapt kairos upgrade list-releases command to handle new repositories
- [ ] we don't let the provider return values, we handle everything in the agent. [link1](https://github.com/kairos-io/kairos-agent/blob/main/internal/agent/upgrade.go#L37) [link2](https://github.com/kairos-io/provider-kairos/blob/main/internal/provider/upgrade.go#L17) - What about other providers? Shouldn't they be able to return new versions with their own logic? We can simply use the same implementation in kairos-agent and provider-kairos and let other providers handle that as they want.
- [ ] by default we filter from the existing attributes that the current image has, e.g. if it's amd64 we only show amd64 and so on for the rest of the attributes, which means we need a place to save model. Maybe the best is to keep it in os-release like we already do for most attributes
- [ ] maybe an option to show all tags without smart filtering
- [ ] ensure that the output can be used to copy paste and use in `kairos-agent upgrade --source` or some fancy selector

relates to https://github.com/kairos-io/kairos/issues/1998

jimmykarily:
Add missing `/etc/os-release` variables to work with versioneer

Fixes #1999 

Replaces https://github.com/kairos-io/kairos/pull/2073


Earthfile
```diff
@@ -368,22 +365,14 @@ uki-tools-image:
...
-    COPY +version/VERSION ./
-    RUN echo "version ${VERSION}"
-
-    ARG --required FAMILY # The dockerfile to use
-    ARG --required FLAVOR # The distribution E.g. "ubuntu"
-    ARG --required FLAVOR_RELEASE # The distribution release/version E.g. "20.04"
-    ARG --required VARIANT
-    ARG --required MODEL
-    ARG --required BASE_IMAGE
+    # Uncomment if ISO_NAME is needed (the code that uses it is commented out below)
```
Itxaka, line 368: not needed anymore, we used to provide the full anme for the artifacts but we do that later in other target :+1:

Earthfile
```diff
@@ -430,8 +419,8 @@ uki:
 # Copy uki artifacts into local build dir
 uki-local-artifacts:
     FROM +uki
-    COPY +version/VERSION ./
-    ARG VERSION=$(cat VERSION)
+    COPY +git-version/GIT_VERSION ./
```
Itxaka, line 422: what is GIT_VERSION ? does that match the old VERSION?

Earthfile
```diff
@@ -472,27 +444,20 @@ iso:
...
-    ARG TARGETARCH
-
-    ARG --required FAMILY # The dockerfile to use
-    ARG --required FLAVOR # The distribution E.g. "ubuntu"
-    ARG --required FLAVOR_RELEASE # The distribution release/version E.g. "20.04"
-    ARG --required VARIANT
-    ARG --required MODEL
-    ARG --required BASE_IMAGE
+    # +base-image will be called again by +uki but will be cached.
+    # We just use it here to take a shortcut to the artifact name
```
Itxaka, line 456: noice

Earthfile
jimmykarily: GIT_VERSION is just the version of kairos (with no k3s in it). It's just the result of `git describe --always --tags --dirty`.

jimmykarily:
I can reproduce this failure locally: https://github.com/kairos-io/kairos/actions/runs/7276077385/job/19825149889?pr=2081
but I don't know what wrong yet.

Earthfile
```diff
@@ -601,44 +526,16 @@ netboot:
...
 arm-image:
   ARG OSBUILDER_IMAGE
   ARG COMPRESS_IMG=true
   ARG IMG_COMPRESSION=xz
-  FROM $OSBUILDER_IMAGE
 
-  COPY +version/VERSION ./
-  ARG KAIROS_VERSION=$(cat VERSION)
-  RUN echo "version ${KAIROS_VERSION}"
+  FROM +base-image
```
mauromorales, line 534: this is the cause for the alpine not to build, because it is re-running the base image but on x86, so it doesn't find the rpi packages. I will try to produce an artifact and just consume that instead of doing this change

mauromorales (approved):

PR #2085, opened 2023-12-14
Comment while we decide what to do

mauromorales:
This will no longer work with versioneer so I commented the test. I'm wondering if the test should do something else though or if we remove it completely

### DIFF
--- a/tests/provider_upgrade_test.go
+++ b/tests/provider_upgrade_test.go
@@ -2,12 +2,12 @@
 package mos_test
 
 import (
-	"encoding/json"
-	"github.com/mudler/go-pluggable"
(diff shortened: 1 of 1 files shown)


jimmykarily:
It's already been tested here: https://github.com/kairos-io/kairos-sdk/blob/74a57eb67370af042872fb67bcd07aa0ee5d02eb/versioneer/tag_list_test.go#L356 not sure if we want to test is through kairos-agent too to make sure versioneer is consumed correctly.

mauromorales:
we still want the test for integration testing, but we should move into another test so we don't waste time spining up https://github.com/kairos-io/kairos/pull/2081

PR #2094, opened 2023-12-18
Use base (cached) images on master

### LINKED ISSUE #2034: Add base images to reduce build time on pipelines
The base images generated by the dockerfiles, do not change very often. We could save pipeline time by not re-building every single time but instead only build if the dockerfile changed.

- [x] push images when their dockerfile has changed
- [ ] use the cached images on master

mauromorales:

Fixes #2034


Itxaka (approved):
I don't get this, but if it works it works

PR #2096, opened 2023-12-19
[POC] Uki slim fedora

Itxaka:
This generates the smaller image possible to use with UKI.

Implementation is currently by breaking everything else in the pipelines LOL

This generates a final iso of 276Mb which boots and installs properly. I dont expect anything else needed for upgrade in the dependencies.
could be a bit lower if we copy the kernel under /lib/modules on /tpm and then remove it so its not bundled in the rootfs, as we are already inserting it in the uki file we dont need to ship it twice.

could probably remove dracut as it comes via deps but its not used.


Itxaka:
This was already done much nicely by @mauromorales by slimming all distros, nice!

PR #2122, opened 2024-01-09
UKI and default platform

mauromorales:


mauromorales:
cancelled jobs since the changes only run on releases

Itxaka (approved):

.github/workflows/release.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
           chmod +x luet
           sudo mv luet /usr/bin/luet
           # Construct an array like this from the found versions:
-          earthly --platform=linux/${{ inputs.arch }} +extract-framework-profile
+          earthly --platform=linux/amd64 +extract-framework-profile
```
Itxaka, line 52: this doesnt work as is ? dont we push for both arches?

mauromorales: I think it was a copy paste from the reusable job, but release is only for amd, while release-arm is for arm

PR #2140, opened 2024-01-11
:sparkles: Enable secureboot for all flavors (minus alpine)

Itxaka:
This PR brings new deps, bumps builders and agent to bring secureboot in all flavors but alpine (more ont hat below)

Currently to enable secureboot in a given X flavor we need to make sure to use the artifacts from that falvor specifically as they are all signed with the distro owner key, so the chain of validation works as expected.

shim (dual signed Microsoft and distro owner) -> grub (distro owner signed) -> kernel (distro owner signed)

So unless there comes a time where the shim is signed by Microsoft and ALL the distro owners (so you can chainload all grubs and kernel signed by all of them) we need to specifically use the distro artifacts so the signatures match.

This PR brings a new enki version(inside osbuilder) which builds the iso using the specifci distro artifacts (instead of generic ones as we used to do).

It also makes sure that all distros include the grub and shim signed artifacts in the base images.

Also brings a new framework that drops the generic grub artifacts (not used) and it also brings a new agent that uses those new artifacts during installation instead of the generic ones.

Unfortunately Alpine does not provide signed artifacts so SecureBoot is not supported under Alpine until they provide those. We still use our generic artifacts to enable uefi support for Alpine

Fixes #

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -9,7 +9,7 @@ ARG LUET_VERSION=0.35.0
 # renovate: datasource=docker depName=aquasec/trivy
 ARG TRIVY_VERSION=0.48.2
 # renovate: datasource=github-releases depName=kairos-io/kairos-framework
-ARG KAIROS_FRAMEWORK_VERSION="v2.6.0"
+ARG KAIROS_FRAMEWORK_VERSION="v2.6.3"
 ARG COSIGN_SKIP=".*quay.io/kairos/.*"
 # TODO: rename ISO_NAME to something like ARTIFACT_NAME because there are place where we use ISO_NAME to refer to the artifact name
 
@@ -21,7 +21,7 @@ END
 ARG COSIGN_EXPERIMENTAL=0
 ARG CGO_ENABLED=0
 # renovate: datasource=docker depName=quay.io/kairos/osbuilder-tools versioning=semver-coerced
-ARG OSBUILDER_VERSION=v0.10.2
+ARG OSBUILDER_VERSION=v0.11.1
 ARG OSBUILDER_IMAGE=quay.io/kairos/osbuilder-tools:$OSBUILDER_VERSION
 ARG GOLINT_VERSION=1.52.2
 # renovate: datasource=docker depName=golang
@@ -207,12 +207,20 @@ luet:
 kairos-dockerfile:
     ARG --required FAMILY
     COPY ./images .
-    RUN --no-cache cat <(echo "# This file is auto-generated with the command: earthly +kairos-dockerfile --FAMILY=${FAMILY}") \
-        <(sed -n '/# WARNING:/!p' Dockerfile.$FAMILY) \
-        <(echo) \
-        <(sed -n '/# WARNING:/!p' Dockerfile.kairos) \
-        > ./Dockerfile
-    SAVE ARTIFACT Dockerfile AS LOCAL images/Dockerfile.kairos-${FAMILY}
+    IF [ "$FAMILY" == "all" ]
+        ARG FAMILY_LIST="alpine debian opensuse rhel ubuntu"
+    ELSE
+        ARG FAMILY_LIST=$FAMILY
+    END
+    FOR F IN $FAMILY_LIST
+        RUN --no-cache cat <(echo "# This file is auto-generated with the command: earthly +kairos-dockerfile --FAMILY=${F}") \
+            <(sed -n '/# WARNING:/!p' Dockerfile.$F) \
+            <(echo) \
+            <(sed -n '/# WARNING:/!p' Dockerfile.kairos) \
+            > ./Dockerfile
+        SAVE ARTIFACT Dockerfile AS LOCAL images/Dockerfile.kairos-${F}
+    END
+
 
 extract-framework-profile:
     ARG FRAMEWORK_VERSION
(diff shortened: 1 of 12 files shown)


Itxaka:
leap 15.5:
 - boots with secureboot
 - installs with secureboot
 - no regexp module in the grub directly so cmdline is NOT correct. we need to ship the module during install and insmod it

Fedora 38:
 - boots with secureboot
 - installs with secureboot
 - regexp modules is bundled in grub

Itxaka:
Debian bookworm:
 - boots with secureboot
 - installs with secureboot
 - regexp bundled with grub


Ubuntu 23.10(probably the rest):
 - NOT boot with secureboot(incorrect naming, grub is stored as grubx64.efi but its looking for grub.efi)
image

Itxaka:
looks like on ubuntu is failing to find the actual shim and using the fallback so that is why it cant find the proper named thingie. So it may be a package missing

```
                +iso | INFO[2024-01-12T08:51:42Z] Creating EFI image...                        
                +iso | DEBU[2024-01-12T08:51:42Z] skip copying /tmp/enki-iso683099797/rootfs/usr/share/efi/x86_64/shim.efi: not found 
                +iso | DEBU[2024-01-12T08:51:42Z] skip copying /tmp/enki-iso683099797/rootfs/usr/lib/shim/shimx64.efi.signed: not found 
                +iso | DEBU[2024-01-12T08:51:42Z] skip copying /tmp/enki-iso683099797/rootfs/boot/efi/EFI/fedora/shim.efi: not found 
                +iso | DEBU[2024-01-12T08:51:42Z] skip copying /tmp/enki-iso683099797/rootfs/boot/efi/EFI/rocky/shim.efi: not found 
                +iso | DEBU[2024-01-12T08:51:42Z] skip copying /tmp/enki-iso683099797/rootfs/boot/efi/EFI/redhat/shim.efi: not found 
                +iso | DEBU[2024-01-12T08:51:42Z] skip copying /tmp/enki-iso683099797/rootfs/boot/efi/EFI/almalinux/shim.efi: not found 
                +iso | DEBU[2024-01-12T08:51:42Z] Using fallback shim file /efi/EFI/BOOT/bootx64.efi 
```

Itxaka:
yeah and the root problem is:

 - shimx64.signed is a link to the alternatives as you can have newer shims `lrwxrwxrwx  1 root root     36 Apr 16  2020 shimx64.efi.signed -> /etc/alternatives/shimx64.efi.signed`
 - in our rootfs alternatives is not installed so you get an empty link
 - testing in a normal image properly links the file


So the solution is to try and pick the `shimx64.efi.signed.latest` file instead or even better, they have a dualsigned one `shimx64.efi.dualsigned` signed by them AND microsoft, so thats even better.

Itxaka:
Using fix https://github.com/kairos-io/enki/pull/25 resutls in proper isos with secureboot for all distros (minus alpine)
Installs correctly for all distros
leap/opensuse still missing the regexp module bundled, will need to check further

Itxaka:
with that merged and pacakge bumped, the only missing fix is for agent:

 - On ubuntu install copy the `grub.cfg` under `/EFI/ubungu/grub.cfg`
 - On opensuse install, copy the regexp module for grub (check the proper paths where its copied)


The default grub config already does an `insmod regexp` so either is not being copied properly or its missing from the sources or its just on a wrong path.

Earthfile
```diff
@@ -207,12 +207,20 @@ luet:
 kairos-dockerfile:
     ARG --required FAMILY
     COPY ./images .
-    RUN --no-cache cat <(echo "# This file is auto-generated with the command: earthly +kairos-dockerfile --FAMILY=${FAMILY}") \
-        <(sed -n '/# WARNING:/!p' Dockerfile.$FAMILY) \
-        <(echo) \
-        <(sed -n '/# WARNING:/!p' Dockerfile.kairos) \
-        > ./Dockerfile
-    SAVE ARTIFACT Dockerfile AS LOCAL images/Dockerfile.kairos-${FAMILY}
+    IF [ "$FAMILY" == "all" ]
```
Itxaka, line 210: ehm, this is something that I added to be able to generate all flavors dockerfiles at the same time as I was getting upset doing it one by one.

Can be dropped

Itxaka:
https://github.com/kairos-io/kairos-agent/pull/203 fixes ubuntu boot.

Earthfile
jimmykarily: nice! I wanted to do the same but didn't get the chance.

Earthfile
```diff
@@ -286,6 +294,14 @@ base-image:
     ARG _CIMG=$(cat ./IMAGE)
 
     COPY +git-version/GIT_VERSION VERSION
+    ARG KAIROS_AGENT_DEV_BRANCH
+
+    IF [ "$KAIROS_AGENT_DEV_BRANCH" != "" ]
```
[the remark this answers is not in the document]
jimmykarily: good idea :+1:

mauromorales (approved):
nice!

Itxaka:
latest fixes should fix:
 - ubuntu grub copy https://github.com/kairos-io/kairos-agent/pull/203
 - opensuse regexp modules missing  https://github.com/kairos-io/packages/pull/619

Itxaka:
both fixes deployed via framework 2.6.3, needs manual QA

tests/tests_suite_test.go
```diff
@@ -304,6 +304,19 @@ func expectRebootedToActive(vm VM) {
...
 }
 
+func expectSecureBootEnabled(vm VM) {
+	// Check for secureboot before install, based on firmware env var
+	// if we set, then the test suite will load the secureboot firmware
+	secureboot := os.Getenv("FIRMWARE")
+
+	if secureboot == "true" {
+		By("checking that secureboot is enabled", func() {
+			out, _ := vm.Sudo("dmesge | grep -i secure")
```
jimmykarily, line 314: `dmesge` -> `dmesg` ? This shouldn't work...

Itxaka: Manually testing this, very weird indeed because the test actually passed??

Itxaka: Ah crap, maybe the if it's not correct so it doesn't go over this...

Itxaka: Yes the check is wrong, I'm stupid

Itxaka: fixed, good catch @jimmykarily !!

jimmykarily (approved):

PR #2161, opened 2024-01-22
Remove duplicated zfs installation on rhel

mauromorales:
Probably missed when refactoring to install different pacakges depending on whether the base image is using epel or fedora base


Itxaka (approved):
Good catch

Issue #870, opened 2023-02-10
Build UKI's with osbuilder


mudler:
To stay consistent with [the plan we have for "factory"](https://github.com/kairos-io/kairos/issues/1633#issuecomment-1827262555) we should implement this in enki and use enki in our pipelines to produce the UKI files.

- [x] #2170
- [x] #2174
- [x] #2171
- [x] #2175
- [x] #2176

jimmykarily:
WIP branch: https://github.com/kairos-io/enki/tree/870-build-uki

mudler:
we need also docs on how to generate keys, ISO and signed EFI files (similarly to Talos docs).

mauromorales:
@Itxaka regarding this item:

> check about /usr/lib/systemd/boot/efi/linuxx64.efi.stub and /usr/lib/systemd/boot/efi/systemd-bootx64.efi who should generate these files on other systems than fedora? in the meantime add them to the checkDeps

If I understand correctly, these files exist if the `systemd-boot` package is installed. So I guess we can trust they will always be in the right dir.

Will we ever need an arm version? probably not needed now but just wondering if we need to specifically check for the file or at least just validate that thers an efi and an efi.sub file under this location

Itxaka:
> @Itxaka regarding this item:
> 
> > check about /usr/lib/systemd/boot/efi/linuxx64.efi.stub and /usr/lib/systemd/boot/efi/systemd-bootx64.efi who should generate these files on other systems than fedora? in the meantime add them to the checkDeps
> 
> If I understand correctly, these files exist if the `systemd-boot` package is installed. So I guess we can trust they will always be in the right dir.
> 
> Will we ever need an arm version? probably not needed now but just wondering if we need to specifically check for the file or at least just validate that thers an efi and an efi.sub file under this location

Yep to the first, we can assume they will be there always I think.

For the second question.... I don't know. Probably yeah we better make sure we can build arm64 ukia fron the start. We can probably leverage the arch value and add an arch flag to cross build them

mauromorales:
thanks for the quick reply and my bad I just see you already made a comment in the original PR about this

mudler:
Closing this issue as it is complete

PR #2169, opened 2024-01-24
Add target to transform kairos image into UKI

### LINKED ISSUE #2170: Consume enki in Kairos Earthfile
Now that enki is able to produce the UKI images, consume it from the base Kairos repo 

Also remove existing UKI "stuff" targets) comment **in the meantime I've decided not yet to remove them because they are being used for development of other tickets. I will just rename them to say `dev` and we can after the entire UKI story is finished decide if we remove or not**

mauromorales:
This new target requires to pass an existing kairos image, which means it's a two step proces e.g.

```
earthly --push +base-image --....
earhtly +uki-enki --BASE_IMAGE=<pushed-image-from-previous-step>
```

I first tried to have it in a single step but I'm not sure if it's possible since enki cannot access the local registry from Earthly. Even when using `WITH DOCKER` this is problematic because our uki building image is on fedora. We have to use fedora because Ubuntu does not have the latest packages e.g. ukify. However, Earthly complains that it doesn't know the distro, so considers it's a Debian and then proceeds unsuccessfully to try to install packages via `apt`.

Please let me know if I'm missing something here. And if it's not possible, then the question for me is really, about why we are doing this on Earthly since we are trying to move away from it for the Factory.

Relates to #870
Fixes #2170

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -322,7 +322,8 @@ image-rootfs:
 uki-tools-image:
     FROM fedora:39
     # objcopy from binutils and systemd-stub from systemd
-    RUN dnf install -y binutils systemd-boot mtools efitools sbsigntools shim openssl systemd-ukify
+    RUN dnf install -y binutils systemd-boot mtools efitools sbsigntools shim openssl systemd-ukify dosfstools xorriso
+    SAVE IMAGE uki-tools
 
 # HOW TO: Generate the keys
 # Platform key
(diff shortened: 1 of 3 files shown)


Earthfile
```diff
@@ -476,9 +478,25 @@ uki-iso:
     RUN xorriso -as mkisofs -V 'UKI_ISO_INSTALL' -e efiboot.img -no-emul-boot -o $ISO_NAME.iso /tmp/efi
     SAVE ARTIFACT /build/$ISO_NAME.iso kairos.iso AS LOCAL build/$ISO_NAME.uki.iso
 
+enki-image:
+    FROM  quay.io/kairos/enki:v0.0.8
+    SAVE ARTIFACT /enki enki
+
 # Uki stuff End
+uki-enki:
+        ARG --required BASE_IMAGE # BASE_IMAGE is existing kairos image which needs to be converted to uki
```
Itxaka, line 487: not sure this is useful here. I mean, nice but will this build the current master kairos and use that as rootfs? Or this requires and already built and pushed image?

mauromorales: it requires a built and pushed image ... the problem is earthly does not play well with local images. I think the only option is to use the `WITH DOCKER` but even that I tested and it seems to only support ubuntu/debian or something because it breaks with fedora.

Earthfile
```diff
@@ -317,12 +317,33 @@ image-rootfs:
...
+enki-image:
+    FROM  quay.io/kairos/enki:v0.0.8
+    SAVE ARTIFACT /enki enki
+
+uki-iso:
+        ARG --required BASE_IMAGE # BASE_IMAGE is existing kairos image which needs to be converted to uki
+        FROM $BASE_IMAGE
+        ARG ISO_NAME=$(cat /etc/os-release | grep 'KAIROS_ARTIFACT' | sed 's/KAIROS_ARTIFACT=\"//' | sed 's/\"//')
+
+        FROM +uki-dev-tools-image
```
jimmykarily, line 329: The enki image ` quay.io/kairos/enki:v0.0.8` has all the dependencies already. You can use that as `FROM`, not just to copy the enki binary from it.

mauromorales: @jimmykarily That container has an entrypoint so it's not really picking up the RUN command properly. Do you know how to do this? I tried to use the `RUN --entrypoint` but no lock

Itxaka (approved):

PR #2172, opened 2024-01-25
[uki] Provide proper artifacts for auto-key-enrollment

Itxaka:
Provide:
 - esl archives: signature list of the certs
 - auth archives: SIGNED data to use in SetVariable

This produces the auth files that systemd-boot uses to auto enroll keys if the machine is in setup mode (no PK key in the firmware)

Good for our testing and to show how it needs to be done to produce them 

Enki part: https://github.com/kairos-io/enki/pull/33


mudler:
Do we also need to move this to enki right? @mauromorales @jimmykarily

mudler:
> Do we also need to move this to enki right? @mauromorales @jimmykarily 

Already done, nevermind :)

Itxaka:
Needs uki test to not provide an efivars.fd so we can test the autoenrollment by default on the test

PR #2201, opened 2024-01-31
Bump framework and osbuilder

Itxaka:

Fixes #

### DIFF
--- a/.github/workflows/reusable-upgrade-latest-test.yaml
+++ b/.github/workflows/reusable-upgrade-latest-test.yaml
@@ -18,7 +18,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: kvm
     env:
       MATCHER: ${{ inputs.release_matcher || inputs.flavor_release }}
     steps:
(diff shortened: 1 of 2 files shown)


Itxaka:
check is not working, @mauromorales was right :sob:

Itxaka:
issues seems to be a different thing, `kairos-agent state get kairos.init` is not working as expected and that was introduced a few days ago and got lumped with this change. It should as soon as framework is rebuilt with the updated package

Itxaka:
Most of the stuff seems to work now. Manual tested uki and also installs and triggers correctly. Will need to check the test.

upgrade job, no idea, need to manually test to see if there is any incompatibility between 2.5 and 3, but alpine one passes...

mauromorales (approved):

Itxaka:
so the actual error for the upgrade is:

```
      +run-qemu-test |   [FAILED] sudo: you do not exist in the passwd database
```

which is baffling... 

This is installing a v2.5.0 image, and it fails during that install...so I have no idea why that would fail somehow????

Itxaka:
current framework is main but I just cut 2.7.1, so if this pass with the current framework it will be ok to set the proper version and merge

PR #2234, opened 2024-02-08
Do not push sbom master files to repo

mauromorales:
on master we don't want to push all the sbom files since the tag is always the same and we don't go back to clean the old sbom files


mauromorales:
cancelling runs since they only run on master

Itxaka (approved):

PR #2242, opened 2024-02-12
Move versioneer call inside Dockerfile

### LINKED ISSUE #2168: :bug: machine-id of UKI images never changes
With the current implementation of UKI, the machine-id stays the same during LiveCD and after installation (for every installation). We might need to investigate this a bit further to know what is the best approach to provide a configurable one

mauromorales:
fixes #2168

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -309,9 +309,7 @@ base-image:
         fi
     END
 
-
-    RUN --no-cache kairos-agent version
-
+    ARG _CIMG=$(cat /IMAGE)
     SAVE IMAGE $_CIMG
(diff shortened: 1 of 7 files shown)


Earthfile
```diff
@@ -287,10 +287,6 @@ base-image:
       -f +kairos-dockerfile/Dockerfile \
       ./images
 
-
-    RUN kairos-agent versioneer container-artifact-name # To see the error if one occurs.
-    RUN kairos-agent versioneer container-artifact-name > /IMAGE
-
```
Itxaka, line 293: can you remove the kairos call in line 313 as well?

Itxaka (approved):

PR #2248, opened 2024-02-14
append k3s version to master builds

### LINKED ISSUE #2225: Nightly for standard needs to also include the k3s version in the name
Nightly is showing as follows `leap-15.5-standard-amd64-generic-master` because the version was manually set to master https://github.com/kairos-io/kairos/blob/master/.github/workflows/reusable-build-provider.yaml#L143

mauromorales:
Fixes #2225

### DIFF
--- a/.github/workflows/reusable-build-provider.yaml
+++ b/.github/workflows/reusable-build-provider.yaml
@@ -140,7 +140,8 @@ jobs:
         env:
           COSIGN_YES: true
         run: |
-          IMAGE=$(FLAVOR=${{ inputs.flavor }} FLAVOR_RELEASE="${{ inputs.flavor_release }}" MODEL=${{ inputs.model }} TARGETARCH=${{ inputs.arch }} VARIANT=${{ inputs.variant}} REGISTRY_AND_ORG="quay.io/kairos" RELEASE=master kairos-agent versioneer container-artifact-name)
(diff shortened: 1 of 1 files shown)


jimmykarily (approved):

.github/workflows/reusable-build-provider.yaml
```diff
@@ -140,7 +140,8 @@ jobs:
         env:
           COSIGN_YES: true
         run: |
-          IMAGE=$(FLAVOR=${{ inputs.flavor }} FLAVOR_RELEASE="${{ inputs.flavor_release }}" MODEL=${{ inputs.model }} TARGETARCH=${{ inputs.arch }} VARIANT=${{ inputs.variant}} REGISTRY_AND_ORG="quay.io/kairos" RELEASE=master kairos-agent versioneer container-artifact-name)
+          K3S_VERSION=$(sudo luet --config framework-profile.yaml search -o json k8s/k3s  | jq '.packages | map(.version) | unique | last' | tr -d '"')
```
jimmykarily, line 143: Hopefully, no new k3s version appears between this call and the other one in the previous step (build) :D

mauromorales: and I just relized this will be problematic because we will not overwrite older k3s versions .... maybe it's better not to fix, wdyt @jimmykarily ?

mauromorales: I'll go with it and we can clean manually worst case. I followed up with quay expiration dates but I don't think they will solve it :/

Issue #2195, opened 2024-01-30
:bug: nohang service fails on Ubuntu 23.10


mauromorales:
tried with the alpha but I found this issue already on 2.5.0

```
PRETTY_NAME="Ubuntu 23.10"
NAME="Ubuntu"
VERSION_ID="23.10"
VERSION="23.10 (Mantic Minotaur)"
VERSION_CODENAME=mantic
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=mantic
LOGO=ubuntu-logo
KAIROS_VERSION_ID="v3.0.0-alpha1-v1.29.0-k3s1"
KAIROS_ARTIFACT="kairos-ubuntu-23.10-standard-amd64-generic-v3.0.0-alpha1-k3sv1.29.0+k3s1"
KAIROS_FLAVOR_RELEASE="23.10"
KAIROS_MODEL="generic"
KAIROS_REGISTRY_AND_ORG="quay.io/kairos"
KAIROS_BUG_REPORT_URL="https://github.com/kairos-io/kairos/issues"
KAIROS_SOFTWARE_VERSION="v1.29.0+k3s1"
KAIROS_NAME="kairos-standard-ubuntu-23.10"
KAIROS_IMAGE_REPO="quay.io/kairos/ubuntu:23.10-standard-amd64-generic-v3.0.0-alpha1-k3sv1.29.0-k3s1"
KAIROS_RELEASE="v3.0.0-alpha1"
KAIROS_ID="kairos"
KAIROS_FLAVOR="ubuntu"
KAIROS_VARIANT="standard"
KAIROS_TARGETARCH="amd64"
KAIROS_HOME_URL="https://github.com/kairos-io/kairos"
KAIROS_SOFTWARE_VERSION_PREFIX="k3s"
KAIROS_VERSION="v3.0.0-alpha1-v1.29.0-k3s1"
KAIROS_ID_LIKE="kairos-standard-ubuntu-23.10"
KAIROS_PRETTY_NAME="kairos-standard-ubuntu-23.10 v3.0.0-alpha1-v1.29.0-k3s1"
[...]

PR #2284, opened 2024-02-22
Slim down Ubuntu

### LINKED ISSUE #2283: UKI: Reduce Ubuntu base image sizes
We are hitting a limit with EFI images and bumping into https://github.com/systemd/systemd/issues/25911, so we want to reduce the base image size

mauromorales:
In this PR:

- remove some packages
  - zfs: if we remove then there's no issue with nohang any more, but then we probably want to get rid of the test too? WDYT? https://github.com/kairos-io/kairos/issues/2195
- introduce the concept of having an image be built for UKI, this image will not install dracut and will not create an initramfs, it will also not include GRUB (see grub and systemd-boot targets)

Relates to #2283

### DIFF
--- a/images/Dockerfile.ubuntu
+++ b/images/Dockerfile.ubuntu
@@ -42,12 +42,12 @@ FROM ${FLAVOR}-${FLAVOR_RELEASE}-upstream AS common
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends \
+  && apt-get install -y --no-install-recommends --no-install-suggests \
     gnupg \
     software-properties-common \
   && add-apt-repository ppa:oibaf/test \
   && apt-get update \
-  && apt-get install -y --no-install-recommends \
+  && apt-get install -y --no-install-recommends --no-install-suggests \
     conntrack \
     console-data \
     cloud-guest-utils \
@@ -75,7 +75,8 @@ RUN apt-get update \
     # can I remove this? it brings perl
     # lldpd \
     lvm2 \
-    nano \
+    # there's pico
+    # nano \
     nbd-client \
     # vim is even bigger, so maybe just use pico
     # neovim \ 
@@ -103,7 +104,7 @@ FROM common AS systemd-boot
 
 FROM common AS grub
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends \
+  && apt-get install -y --no-install-recommends --no-install-suggests \
     grub2-common \
     # this brings grub as a dependency
     shim-signed \
@@ -114,7 +115,7 @@ FROM systemd-boot AS systemd-boot-arm64
 
 FROM grub AS grub-amd64
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends \
+  && apt-get install -y --no-install-recommends --no-install-suggests \
     grub2 \
     grub-efi-amd64-bin \
     grub-efi-amd64-signed \
@@ -124,7 +125,7 @@ RUN apt-get update \
 
 FROM grub AS grub-arm64
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends \
+  && apt-get install -y --no-install-recommends --no-install-suggests \
     grub-efi-arm64 \
     grub-efi-arm64-bin \
     grub-efi-arm64-signed \
@@ -136,7 +137,7 @@ RUN apt-get update \
 
 FROM ${BOOTLOADER}-${TARGETARCH} AS amd64
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends \
+  && apt-get install -y --no-install-recommends --no-install-suggests \
     dmsetup \
     efibootmgr \
     isc-dhcp-common \
@@ -159,7 +160,7 @@ RUN apt-get update \
 
 FROM ${BOOTLOADER}-${TARGETARCH} AS arm64
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends \
+  && apt-get install -y --no-install-recommends --no-install-suggests \
     keyutils \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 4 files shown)


mauromorales:
Closing because we will go with the other solution that allows to have bigger images

IMO we should still do this despite the new option because:

1. there is no reason to have stuff in the UKI image that is not used there at all
2. we should really evaluate if all the packages we have there are actually required, there were things added that I'm not sure anyone is using and to anyone consuming Kairos they should be adding their packages on top not directly upstream
3. also evaluate in cases like zfs which breaks nohang atm and probably also will in 24.10, and right now we just let it break but what is the point on shipping it like that
4. now that we have a hard constrain is a good time to push some maintenance on our side and on canvos side. Otherwise we don't have that much time for it, even automated PRs don't get easily merged
5. at some point someone will hit the limit, and then we will have to come back and do this work which we already digged into

but obviously not every package needs to be removed here and there's a timeline, so I don't want that to be problematic

images/Dockerfile.ubuntu
```diff
@@ -41,50 +42,51 @@ FROM ${FLAVOR}-${FLAVOR_RELEASE}-upstream AS common
...
-    neovim \
+    # vim is even bigger, so maybe just use pico
+    # neovim \ 
     nohang \
     open-iscsi \
     open-vm-tools \
     openssh-server \
     parted \
     rsync \
-    shim-signed \
```
Itxaka, line 85: I think we need the signed shim modules everywhere to support secureboot!

images/Dockerfile.ubuntu
```diff
@@ -41,50 +42,51 @@ FROM ${FLAVOR}-${FLAVOR_RELEASE}-upstream AS common
...
-    snapd \
-    snmpd \
+    # snapd \
```
Itxaka, line 87: Noice! I wonder if this will remove the /snap dir in the rootfs lol

images/Dockerfile.ubuntu
```diff
@@ -41,50 +42,51 @@ FROM ${FLAVOR}-${FLAVOR_RELEASE}-upstream AS common
...
+    # can I remove this? it brings perl
+    # snmpd \
     squashfs-tools \
```
Itxaka, line 90: Do we need this? We used to do so when we supported squashfs images as we needed the squash tools in the image but I don't think we longer need them?

Itxaka: Ah we may need it for dracut but we no longer ship images in squashfs but in the livecd and I think that has the modules already via luet package

images/Dockerfile.ubuntu
```diff
@@ -96,58 +98,87 @@ RUN apt-get update \
...
+FROM grub AS grub-amd64
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends --no-install-suggests \
+    # this will probably break 20.04
+    dracut-live \
+    grub2 \
+    grub-efi-amd64-bin \
+    grub-efi-amd64-signed \
+    grub-pc-bin \
+    os-prober \
```
Itxaka, line 125: We can drop this no? We generate the entries manually and we don't support any other os along Kairos...

Itxaka (commented):
I'm worried about the install suggests as that could bring a package that we need but we don't notice or know lol

mauromorales:
> I'm worried about the install suggests as that could bring a package that we need but we don't notice or know lol

ok, will revert

images/Dockerfile.ubuntu
mudler: snapd is something actually very useful and I'd keep. 

Is taking much space?

A user could easily add something with snapd, and avoid messing up with apt

images/Dockerfile.ubuntu
mudler: I don't think we need dracut with UKI

mudler:
I think we should have a separate image for ubuntu-trustedboot. We can remove then all grub, dracut, and all the steps that are generating even the initramfs

images/Dockerfile.ubuntu
mauromorales: ok, then I will move to the "grub" section so it's not installed for UKI images

images/Dockerfile.ubuntu
mudler: shouldn't be dumped from the LiveCD during the installation phase? 

Do we really need those in the OS runtime?

mauromorales: shim-signed, has as dependency GRUB, do we also need it for UKI? and if so is there a way to get only what we need and not GRUB?

mauromorales:
> I think we should have a separate image for ubuntu-trustedboot. We can remove then all grub, dracut, and all the steps that are generating even the initramfs

@mudler yes, this is what the `grub` target does

images/Dockerfile.ubuntu
Itxaka: what is the livecd but the os packed into a squashfs? We need to bundle those in the os otherwise no secureboot in non-uki installs

mudler:
> > I think we should have a separate image for ubuntu-trustedboot. We can remove then all grub, dracut, and all the steps that are generating even the initramfs
> 
> @mudler yes, this is what the `systemd-boot` vs `grub` targets do

sorry, didn't saw the workflow/additional dockerfile so I thought this was going into another direction. All good :+1:

images/Dockerfile.ubuntu
mauromorales: @mudler it increases the size by 100Mb but I see why you'd like to keep it

mauromorales: I can still boot the standard image with it so I'll keep it but it's space that the consumer will not have, we can decide later once Canvos can tell if the space is good enough

images/Dockerfile.ubuntu
mauromorales: yeah it's not being removed on non-uki, but do we need it for UKI?

images/Dockerfile.ubuntu
mauromorales: moved to the non-uki section

images/Dockerfile.ubuntu
Itxaka: No, for uki is not needed, all artifacts are provided by osbuilder

Itxaka (approved):
good for me as long as you disable the zfs job lol :D

mauromorales:
@Itxaka I ended up keeping zfs for non-UKI so we don't have to decide that now

PR #2290, opened 2024-02-27
Publish uki base image

mauromorales:
:warning: most important, do you agree with the following name for those base images?

e.g.:

-  ubuntu:22.04-core-amd64-generic-master-uki
-  ubuntu:22.04-core-amd64-generic-v3.0.0-uki

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -38,7 +38,7 @@ jobs:
         sudo apt update && sudo apt install -y jq
     - id: set-matrix
       run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -213,26 +213,106 @@ jobs:
           category: ${{ matrix.flavor }}
   build-core-uki:
     runs-on: ubuntu-latest
-    needs:
-      - build-core
-      - get-uki-matrix
```
Itxaka, line 218: this is still needed no? Otherwise the matrix below will fail?

mauromorales: :man_facepalming:

Itxaka (approved):

PR #2310, opened 2024-03-06
:robot: Do not double pull the BASE_IMAGE on uki iso

Itxaka:
As we now dont need the IMAGE var to extract the iso, there is no point in having to pull the full BASE_IMAGE just to extrac the name.

This should speed up the uki iso generation


mauromorales (approved):
good catch 🙌

PR #2314, opened 2024-03-07
Use Ubuntu 22 LTS for ZFS test

mauromorales:
This is because zfs is not being installed on 23+ since it clashes
with nohang and it's not a requirement to have


mauromorales:
cancelling jobs since it will only run on master

Itxaka (approved):

PR #2319, opened 2024-03-08
🤖 Expand uki tests

Itxaka:
Signed-off-by: Itxaka <itxaka@kairos.io>

Expands uki tests to check:
 - standard version instead of core
 - k3s works
 - can install calico and it runs
 - rootfs is RO
 - /proc is shared
 - rootfs is mounted as tmpfs

Fixes #


jimmykarily:
We can also consider switching to Ubuntu (https://github.com/kairos-io/kairos/issues/2363) in this PR, otherwise on a different one.

PR #2381, opened 2024-03-22
:robot: Bump osbuilder and drop keys

Itxaka:
Bumps osbuilder to latest, drops test keys and makes the earthly target create them on the fly when building the image as we only need them to sign the images and we dont plan to do anything else with them, we can simply fully drop them and generate them on demand for ci

Fixes #


Itxaka:
ahhh unless we set the loader.conf entry to be `secure-boot-enroll if-safe` which would autoenroll only in VMs, our uki tests are broken as we cannot select the entry programatically...

mauromorales:
ah this answers my question on the other PR hehehe

mauromorales (approved):
niceee

jimmykarily:
I hope git is clever enough to handle this but make sure that they are still removed after rebasing to this: https://github.com/kairos-io/kairos/pull/2331/files

PR #2415, opened 2024-04-02
Remove snap from Ubuntu based images

### LINKED ISSUE #2414: feat: remove snap from Ubuntu based images

mauromorales:
Originally requested by @santhoshdaivajna 

I indeed think snap should not be part of the base image but instead users who rely on it should have their own images and pass it through the kairos factory. However, it's already out there so I'm also wondering how we should inform users about it? Does it suffice to simply put a release note? @mudler 

Fixes #2414

### DIFF
--- a/images/Dockerfile.kairos-ubuntu
+++ b/images/Dockerfile.kairos-ubuntu
@@ -88,7 +88,6 @@ RUN  apt-get update \
     openssh-server \
     parted \
     rsync \
-    snapd \
     sudo \
     systemd \
     systemd-timesyncd \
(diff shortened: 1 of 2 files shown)


mudler:
> Originally requested by @santhoshdaivajna
> 
> I indeed think snap should not be part of the base image but instead users who rely on it should have their own images and pass it through the kairos factory. However, it's already out there so I'm also wondering how we should inform users about it? Does it suffice to simply put a release note? @mudler
> 
> Fixes #2414

release notes sounds good. we have also to make sure to update our docs if mentions snapd support directly

mauromorales:
I couldn't find any reference on the docs, so I think we're good

mudler (approved):

mauromorales:
tests consistently fail, digging deeper

Issue #2428, opened 2024-04-04
[patch-release] v3.0.x


mudler:
Tracker for the patch releases for the 3.0.x branch:

- [x] v3.0.1
- [x] v3.0.2 ( https://github.com/kairos-io/kairos/issues/2379 )
- [x] v3.0.3
- [x] #2427 ( https://github.com/kairos-io/kairos/issues/2384 )
- [x] v3.0.5 (https://github.com/kairos-io/kairos/pull/2433) ~ no need as restarting the job fixed the nohang issue ~
- [x] v3.0.7
- [x] v3.0.8
- [x] https://github.com/kairos-io/kairos/issues/2528
- [x] https://github.com/kairos-io/kairos/issues/2535
- [x] https://github.com/kairos-io/kairos/issues/2560
- [x] https://github.com/kairos-io/kairos/issues/2582
- [x] https://github.com/kairos-io/kairos/issues/2588
- [x] 3.0.14 https://github.com/kairos-io/kairos/releases/tag/v3.0.14
- [ ] #2603

ci-robbot:
Hello mudler, thank you for opening this issue. While I am a bot created to audit tickets for a project with specific rules, I am unable to process this issue as it is not formatted correctly for the project's requirements.

To meet the project's requirements, please ensure that the issue description has a clear subject and body that describes the issue at hand. If it's a bug, please provide steps to reproduce it. Lastly, please mention the versions of the artifacts being used.

Once you provide enough information, the issue will be reviewed, labeled, and assigned accordingly. Thank you for your understanding, and remember that I am an experiment created by @mudler and @jimmykarily.

If you have any questions or need further assistance, please don't hesitate to reach out.

jimmykarily:
Closing this. It's only jetson orin QA that needs to happen but it's not urgent so it might not happen for a while.

PR #2433, opened 2024-04-04
fix(nvidia): do not ship nohang in nvidia-arm builds

mudler:
Even if nohang is a welcome addition, first it's better to not use testing repo - second nohang is optional as it is not really a requirement for things to be functional

#2428


mauromorales (approved):

mauromorales:
I don't think any of the workflows run nvidia on PR, created a ticket https://github.com/kairos-io/kairos/issues/2434

PR #2463, opened 2024-04-11
Symlink any /boot/Image* to /boot/vmlinuz

### LINKED ISSUE #2461: Missing /boot/vmlinuz in jetson orin base image
Tried with kairos v2.4.3 and v3.0.4. Both results into same error.
image
And i noticed that with 2.3.2 image which was working before has this /boot/vmlinuz but neither 2.4.3 or 3.0.4 base image has it

mauromorales:
The jetson produces `/boot/Image` which wasn't caught by the previous mechanism

fixes #2461


mudler (approved):

jimmykarily (approved):

mauromorales:
tested manually, this tests dont' run on pr so merging

Issue #2467, opened 2024-04-12
Jetson orin not booting up


nianyush:
After resolving the issue in #2461, it still stuck in boot up process. Not able to login and collect logs
![image](https://github.com/kairos-io/kairos/assets/5505684/2a0de5a4-d539-43a3-b814-5fa77d7311b4)

mudler:
@nianyush can you reproduce this with the kairos images too?

jimmykarily:
relevant? https://forums.developer.nvidia.com/t/orin-nano-wont-boot-lt4launcher-attempting-recovery-boot/255302/4

jimmykarily:
> relevant? https://forums.developer.nvidia.com/t/orin-nano-wont-boot-lt4launcher-attempting-recovery-boot/255302/4

meh... not. It seems that ours is booting into palette.

mudler:
seems we moved the iscsi.conf from the framework image to: https://github.com/kairos-io/packages/blob/99dec691da2d34468e2fcf2e14d3b9d388848da4/packages/static/kairos-overlay-files/files/system/oem/12_nvidia.yaml#L8

![image (14)](https://github.com/kairos-io/kairos/assets/2420543/40f756e9-8ee2-47d9-802c-64ac23fbd8f0)

However, the dracut configuration is needed while building the initramfs (and has to be added into the framework image)

@nianyush for the orin you might need to add a patch in the meantime in your Dockerfiles:

```
echo 'omit_dracutmodules+=" iscsi "' > /etc/dracut.conf.d/iscsi.conf
```

**before** generating the initramfs with dracut

ci-robbot:
Hello! I'm a bot that audits the tickets opened in this Github project. It seems like your issue is not completely clear, as the steps to reproduce and the versions of the relevant artifacts being used are not mentioned. According to the project's rules, these details should be provided in any new tickets.

Please provide more information such as the steps to reproduce the problem, the versions of the kairos images and the Jetson platform being used, as well as any other relevant details. This will help the maintainers of the project to better understand and address your issue.

I'm sure you're eager to get your problem solved, but please remember that I'm just a bot and I can't provide any direct assistance with the technical issue itself. If you have any further questions, feel free to ask!

Note: I am a bot, an experiment of @mudler and @jimmykarily.

mauromorales:
Venkatesh confirmed that the jetson boots now

PR #2477, opened 2024-04-15
Push master ARM images to quay

mauromorales:
this is so we can offer a master build artifact for Nvidia or other arm images

relates to https://github.com/kairos-io/kairos/issues/2467


mauromorales:
Changes only run on master or releases, so cancelling runs

PR #2541, opened 2024-05-07
Remove double brackets from Earthly

### LINKED ISSUE #1759: Switch all double brackets in Earthly to single brackets
Under the hood, Earthly calls `sh` to execute control flow. Unfortuantely this is prone to errors because the underlying container can either use a posix compliant `sh`, or a different one e.g. `bash`. One example of the former is Ubuntu:

```
/bin/sh: 1: [[: not found
```

To make matters worse, Earthly doesn't halt because of this failure, so the process finishes successful but build something unexpected.

mauromorales:
This was a lot simpler than when the ticket was originally created thanks to us moving the logic to the Dockerfiles 🎉 

fixes #1759

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -606,7 +606,7 @@ arm-image:
   WORKDIR /build
   # These sizes are in MB
   ENV SIZE="15200"
-  IF [[ "$FLAVOR" = "ubuntu" ]]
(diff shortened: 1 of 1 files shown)


Itxaka:
if we got no brakes, we gonna crash! :P

mauromorales:
@Itxaka good point 😅

PR #2549, opened 2024-05-09
Do not auto reboot on uki tests

Itxaka:
Signed-off-by: Itxaka <itxaka@kairos.io>

Fixes #


.github/workflows/uki.yaml
```diff
@@ -11,125 +11,22 @@ concurrency:
...
-            --FAMILY=${{ matrix.family }} \
-            --BOOTLOADER=systemd-boot
-          sudo mv build/* .
-          sudo rm -rf build
-      - name: Install kairos-agent (for versioneer)
-        uses: Luet-lab/luet-install-action@v1.1
-        with:
-          repository: quay.io/kairos/packages
-          packages: system/kairos-agent
-      - name: Push to quay
```
mauromorales, line 111: I think we still need core because this is what the edge team has been using for their tests, no?

Itxaka: Core was just the same as the Ubuntu lts no? It's building 24.04? Or did I screwed it up?

Itxaka: Ohh this pushes it to quay on master!

Good catch, I thought we had that on the reusable job...we may need to have that there anyway?

mauromorales: No I think we don't have it on the reusable job but I guess we could use the reusable job to build and then still push just on the ones we want to push, since we don't want fedora atm for example

Itxaka: I added it to the reusable job, now it pushes AFTER tests pass and only if its from master and UBUNTU

I would change that to push always, so any image that we test is pushed....makes sense to me, if we run tests and it passes, we should make that image available no?

mauromorales (approved):

.github/workflows/uki.yaml
Itxaka: well, maybe not unless we "support" them somehow rigth?

Its nice to have the images pushed from master so we can test building isos from there and such. But on release, we are committing to it rigth? like we are saying is supported by making it an official release artifact rather than a in-progress dev version. I would be ok with that but our current focus is on ubuntu so not sure if it would make sense to make it officially supported with our bandwidth :(

Something to discuss on Monday I guess?

Itxaka:
Blocked until we can discuss on monday meeting what to do with fedora uki releases

mauromorales (approved):

Issue #2565, opened 2024-05-16
feat: remove firmware in Ubuntu uki images


mudler:
The firmware in the ubuntu image brings nearly ~500MB (see also the previous attempt in #2283 ).

We can drop the firmware in the core uki images such as those have to be consumed already in a process that includes remastering and customization - and document that we don't ship firmware and is left to the user to build images with the proper firmware.

```
@@ -229,8 +229,10 @@ FROM base-ubuntu-current AS kernel-ubuntu-24.04

 RUN apt-get update
 # If a kernel is already installed, don't try to install it again, this way the base image can
 # be non-hwe for older releases
+#RUN [ -z "$(ls -A /lib/modules/)" ] && apt-get install -y --no-install-recommends \
+#    linux-image-generic-hwe-24.04 || true
 RUN [ -z "$(ls -A /lib/modules/)" ] && apt-get install -y --no-install-recommends \
-    linux-image-generic-hwe-24.04 || true
+    amd64-microcode intel-microcode iucode-tool kmod linux-base linux-image-6.8.0-31-generic linux-modules-6.8.0-31-generic || true
 RUN apt-get clean && rm -rf /var/lib/apt/lists/*
```

ci-robbot:
Hello mudler, thank you for opening this issue (kairos-io/kairos#2565). In order to meet the project's requirements, please provide a detailed description of the issue and, if it's a bug, the steps to reproduce it. Additionally, it would be helpful if you could mention the versions of the relevant artifacts being used.
I apologize for any inconvenience this may cause, but as a bot, I have labeled this issue with 'enhancement' and 'triage' labels. Please update the issue with the necessary information to proceed with the review process. Remember, this is an experiment by @mudler and @jimmykarily. If you have any further questions, feel free to ask.

mauromorales:
@mudler what do you think about #2566 ? I also like the idea of no firmware

bencorrado:
For anyone who needs firmware and finds this you can install the firmware then remove the things you don't need.  I figured this out by booting a system with firmware removed and grepping dmesg for firmware then noting the missing directories and adding them to the code below.  This should go into a dockerfile before doing any enki packaging.

 
```
# Remove all files in /usr/lib/firmware except the specified ones in mediatek, i915, intel, amd, and their sub-directories
RUN find /usr/lib/firmware \
    ! -path "/usr/lib/firmware" \
    ! -path "/usr/lib/firmware/mediatek" \
[...]

PR #2566, opened 2024-05-16
Install limited amount of modules for UKI Ubuntu

mauromorales:
Not sure if this is the right approach but since it's quite simple to test this with the multi stage dockerfile then I wanted to give it a go

Not an original idea, recommended by Kevin

relates to https://github.com/kairos-io/kairos/issues/2565

### DIFF
--- a/images/Dockerfile.ubuntu
+++ b/images/Dockerfile.ubuntu
@@ -115,13 +115,13 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   kmod \
   linux-base \
   && apt-get remove -y unattended-upgrades && apt-get clean && rm -rf /var/lib/apt/lists/*
-# for UKI, install the linux-image pkg if no kernel is found
-RUN [ -z "$(ls -A /boot/vmlinuz*)" ] && apt-get update && apt-get install -y --no-install-recommends \
-  "$(apt-cache search linux-image | grep -oP 'linux-image-\d+.\d+.\d+-\d+-generic' | sort -V | tail -1)" \
-  && apt-get remove -y unattended-upgrades && apt-get clean && rm -rf /var/lib/apt/lists/*
 # for UKI, install the linux-modules pkg if no modules are found
 RUN [ -z "$(ls -A /lib/modules/)" ] && apt-get update && apt-get install -y --no-install-recommends \
   "$(apt-cache search linux-modules | grep -oP 'linux-modules-\d+.\d+.\d+-\d+-generic' | sort -V | tail -1)" \
+  && apt-get remove -y unattended-upgrades && apt-get clean && rm -rf /var/lib/apt/lists/*
+# for UKI, install the linux-image pkg if no kernel is found
+RUN [ -z "$(ls -A /boot/vmlinuz*)" ] && apt-get update && apt-get install -y --no-install-recommends \
+  "$(apt-cache search linux-image | grep -oP 'linux-image-\d+.\d+.\d+-\d+-generic' | sort -V | tail -1)" \
   && apt-get remove -y unattended-upgrades && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # This target is aimed at non-UKI images where size is not a concern
--- a/images/Dockerfile.ubuntu
+++ b/images/Dockerfile.ubuntu
@@ -109,6 +109,15 @@ RUN rm -rf /tmp/nohang.deb
 # This target is aimed at UKI images and avoids to install GRUB and any other unnecesary packages
 # to the image because there are limitations on the total size
 FROM common AS systemd-boot
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends \
+  $(dpkg --print-architecture)-microcode \
+  iucode-tool \
+  kmod \
+  linux-base \
+  $(apt-cache search linux-image | grep -oP 'linux-image-\d+.\d+.\d+-\d+-generic' | sort -V | tail -1) \
+  $(apt-cache search linux-image | grep -oP 'linux-modules-\d+.\d+.\d+-\d+-generic' | sort -V | tail -1) \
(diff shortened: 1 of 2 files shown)


Itxaka (commented):
That was fast! But this should be a separated image no? Or are we shipping this by default?

mauromorales:
@Itxaka that's where I'm not sure, I don't see why not ship it like this and if we see that it doesn't work in many places then we go back to the "big" image, but in general I think the space issue will be a concern for any user

Itxaka:
> @Itxaka that's where I'm not sure, I don't see why not ship it like this and if we see that it doesn't work in many places then we go back to the "big" image, but in general I think the space issue will be a concern for any user

I would say, we keep them separated until we can provide the modules/fw into the system with an extra step during install or image build or something no?

Otherwise we are exchanging something that works out of the box for 95% of systems with something that might or might not work out of the box for an unknown percentage of systems with no workaround other than....rebuild the image yourself directly and install those packages.


IMHO provided images by us should be pretty generic and support as much software as possible, even if we have to notice that some EFI implementations will not be able to boot those image, then point to this nano images as an alternative, less supported, more difficult ones.

But that's just my opinion. For sure if someone is gonna consume this, we need to get them in our workflow to test them in case we add a dep like a module that is missing. The rest is up for debate.

Itxaka:
We probably are also winning in the EFI firmware part with smaller images. After all fedora does not ship so much stuff and seems ok...

One for the other lol

mauromorales:
@Itxaka yeah good question, what is more generic, cuz if the image does not fit, then it doesn't matter that it has all the generics. But I don't have a strong opinion, we can go either or. I think this is something that maybe we can offer multiple options, with the least amount of firmware to all the firmware and let the user decide since they are the ones with the hardware

Itxaka:
> @Itxaka yeah good question, what is more generic, cuz if the image does not fit, then it doesn't matter that it has all the generics. But I don't have a strong opinion, we can go either or. I think this is something that maybe we can offer multiple options, with the least amount of firmware to all the firmware and let the user decide since they are the ones with the hardware

Yep indeed, I think it would be good to offer both options.

Full firmware, most of the stuff supported out of the box. Downsides, it may not boot in all hardware.

Lite, should boot in most of the systems out there. Downsides, it may not support all your hardware, might require post install steps to provide extra modules/firmware 


So we cover both ways and have alternatives for the uki stuff.

jimmykarily:
I'm still getting a huge ISO from this branch:

```
~/workspace/kairos/kairos (reduce-size-of-uki-image)*$ du -h build/kairos-ubuntu-24.04-core-amd64-generic-v3.0.4-71-ge24bc2ac.iso
1,2G	build/kairos-ubuntu-24.04-core-amd64-generic-v3.0.4-71-ge24bc2ac.iso
```

am I supposed to pass a flag to earthly or something?

mauromorales:
@jimmykarily yes you need to pass `--BOOTLOADER=systemd-boot`

jimmykarily:
```
$ ~/workspace/kairos/kairos (reduce-size-of-uki-image)*$ earthly +iso --VARIANT=core --FLAVOR=ubuntu --FLAVOR_RELEASE=24.04  --BASE_IMAGE=ubuntu:24.04  --MODEL=generic --FAMILY=ubuntu  --BOOTLOADER=systemd-boot

...

                +iso *failed* | INFO[2024-05-23T06:18:45Z] Starting rsync...                            
                +iso *failed* | DEBU[2024-05-23T06:18:45Z] Running cmd: 'rsync --progress --partial --human-readable --archive --xattrs --acls --exclude=/mnt --exclude=/proc --exclude=/sys --exclude=/dev --exclude=/tmp --exclude=/host --exclude=/run /grub2/ /tmp/enki-iso071332082/iso/' 
                +iso *failed* | INFO[2024-05-23T06:18:45Z] Finished syncing                             
                +iso *failed* | INFO[2024-05-23T06:18:45Z] Finished copying /grub2 into /tmp/enki-iso071332082/iso 
                +iso *failed* | ERRO[2024-05-23T06:18:45Z] No initrd file found                         
                +iso *failed* | ERRO[2024-05-23T06:18:45Z] Could not find kernel and/or initrd          
                +iso *failed* | ERRO[2024-05-23T06:18:45Z] Failed preparing ISO's root tree: No file found with prefixes: [initrd initramfs] 
                +iso *failed* | ERRO[2024-05-23T06:18:45Z] 1 error occurred:
                +iso *failed* | 	* No file found with prefixes: [initrd initramfs]
                +iso *failed* |  
                +iso *failed* | ERROR Earthfile line 544:4
                +iso *failed* |       The command
                +iso *failed* |           RUN /entrypoint.sh --name $ISO_NAME --debug build-iso --squash-no-compression --date=false dir:/build/image --output /build/
                +iso *failed* |       did not complete successfully. Exit code 1

Help: To debug your build, you can use the --interactive (-i) flag to drop into a shell of the failing RUN step: "earthly -i +iso --VARIANT=core --FLAVOR=ubuntu --FLAVOR_RELEASE=24.04 --BASE_IMAGE=ubuntu:24.04 --MODEL=generic --FAMILY=ubuntu --BOOTLOADER=systemd-boot"

...

**Comment by jimmykarily** [comment]
I'm stupid :facepalm: . This is not meant to build an iso (`+iso`) but only a base-image to be passed to enki.

**Review by jimmykarily — APPROVED** [review]

File: images/Dockerfile.ubuntu [hunk 1]
```diff
@@ -131,6 +140,8 @@ RUN apt-get update \
 FROM systemd-boot AS systemd-boot-amd64
 RUN apt-get update \
   && apt-get install -y --no-install-recommends \
+    amd64-microcode \
+    intel-microcode \
```
mauromorales, line 144:
```suggestion
```

PR #2625, opened 2024-06-14
:penguin: Add selinux packages to ubuntu 24.04

### LINKED ISSUE #2577: ubuntu 24.04 with k3s 1.28.9 crash at boot
**Kairos version:**

# kairos version
provider-kairos version: 2.6.4, compiled with: go1.21.7

arch: amd64
os: ubuntu 24.04

**Describe the bug**

After build and install ubuntu 24.04 with k3s 1.28.9, the os keep crashing when k3s service start. 
full error report is here https://github.com/k3s-io/k3s/issues/10101

**To Reproduce**

- build kairos with k3s 1.28.9 
- try to boot it
- system crash and reboot
**Expected behavior**

it shall boot and run

**Logs**

see https://github.com/k3s-io/k3s/issues/10101
**Additional context**

Itxaka:
Otherwise it wont work as expected

Fixes #2577


jimmykarily:
I wonder if this is needed on any other versions of Ubuntu. I don't think we tried all others.

jimmykarily (approved):

Itxaka:
> I wonder if this is needed on any other versions of Ubuntu. I don't think we tried all others.

There is something in the release notes of Ubuntu 24 talking on how now it's not possible to disable SE Linux during runtime but I could not get any more info on what does that exactly means.

Seems to come with the kernel so maybe that selinux=1 that we used to set is being ignored and selinux triggers and can't find any policies so it crashes hard.

Weird enough with kernel 6.10 this did not happen so maybe it's really a kernel bug that crashes heavily without the the policies and they fixed it on the kernel directly?

PR #2645, opened 2024-06-25
:wrench: : Update to use new osbuilder

Itxaka:
Brings latest osbuilder which has the pure golang ukifier


Itxaka:
Fedora seems to fail, which I dont find weird as the support for uki fedora is best-effort. That should be ok to merge anyway

Weirdly enough, alpine reset fails, while it passes locally? No idea why.

mauromorales (approved):

Itxaka:
Nice, all green!

Issue #2647, opened 2024-06-25
Address OpenSSF best practices issues


mauromorales:
https://scorecard.dev/viewer/?uri=github.com/kairos-io/kairos&sort_by=check-score&sort_direction=asc

Itxaka:
some of the issues were fixed by tweaking renovate to use digests

mauromorales:
Got it to green 🎉 and it should increase a bit more with some of the changes that are calculated by checking the last x commits.

PR #2649, opened 2024-06-25
Fix dangerous workflow

mauromorales:
relates to #2647


Itxaka (approved):
I dont get how this is considered dangerous? but whatever, looks good?

mauromorales:
@Itxaka yeah i was also confused about it, this doc explains it https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#example-of-a-script-injection-attack

PR #2676, opened 2024-06-27
[WIP] Define top permission as read-all

mauromorales:

Fixes #

### DIFF
--- a/.github/workflows/image-pr.yaml
+++ b/.github/workflows/image-pr.yaml
@@ -4,6 +4,7 @@ on:
     paths:
       - '**'
 
+permissions: read-all
 concurrency:
   group: ci-image-${{ github.head_ref || github.ref }}-${{ github.repository }}
   cancel-in-progress: true


.github/workflows/image-arm-pr.yaml
```diff
@@ -14,6 +15,21 @@ env:
 jobs:
   opensuse:
     uses: ./.github/workflows/reusable-docker-arm-build.yaml
+    permissions:
+      id-token: write  # OIDC support
+      contents: write
+      security-events: write
+      actions: read
```
Itxaka, line 22: Do we need to be explicit with the read ones? I would expect this entries to be read by default and only need to override the write ones?

Itxaka (approved):

.github/workflows/image-arm-pr.yaml
mauromorales: I first had it without but it didn't work ... maybe there's a way but i'm not sure about the syntax

Issue #2670, opened 2024-06-27
bump projects to go 1.22 and fix snyk cve


jimmykarily:
let's also produce fips versions of the binaries to be prepared for then that is certified.

## Acceptance Criteria
- We let the product team we are doing this so they know why and what they have to do to be able to consume 3.2.X
- All `kairos-io` golang repositories use go1.22 to build binaries (in pipelines)
- all `go.mod` files refer to go1.22 as well
- all packages (luet) are also built with go1.22 (even the fips ones)

mauromorales:
This PR should address some of the issues related to the main kairos repo and it doesn't affect fips since it's just testing https://github.com/kairos-io/kairos/pull/2678

Itxaka:
- https://github.com/kairos-io/kairos-agent/pull/405
 - https://github.com/kairos-io/immucore/pull/347
 - https://github.com/kairos-io/kairos-sdk/pull/168
 - https://github.com/kairos-io/provider-kairos/pull/594
 - https://github.com/kairos-io/kcrypt/pull/413
 - https://github.com/kairos-io/kcrypt-challenger/pull/70
 - https://github.com/kairos-io/enki/pull/135
 - https://github.com/kairos-io/AuroraBoot/pull/60
 - https://github.com/kairos-io/go-nodepair/pull/32
 - https://github.com/kairos-io/osbuilder/pull/172
 - https://github.com/kairos-io/cluster-api-provider-kairos/pull/29
 - https://github.com/kairos-io/entangle/pull/11

Itxaka:
Everything merged and bumped, all of them also had a tag cut after the go version was bumped so they can be consumed from the packages repo

Itxaka:
All packages updated in repos, consider this done

PR #2678, opened 2024-06-28
Add osv scanning for PRs

mauromorales:
relates to https://github.com/kairos-io/kairos/issues/2647 and #2670


Itxaka (approved):

Itxaka:
saw this on the ovs

`Skipping call analysis on Go code since Go is not installed.` 

Is this correct?

mauromorales:
> saw this on the ovs
> 
> `Skipping call analysis on Go code since Go is not installed.`
> 
> Is this correct?

Good question, my guess is that it doesn't see anything but the go.mod and go.sum files so not the tests dir with the go code there. But do we want to test the tests dir? Since it's not shipped to the user

mauromorales:
I think it's ok because by default it runs recursively

Issue #1792, opened 2023-09-05
📢 Kairos release v3.1.0


mudler:
## 🗺 What's left for release

- [x] #2117
- [x] https://github.com/kairos-io/kairos/issues/2200
- [x] https://github.com/kairos-io/kairos/issues/2565
- [x] https://github.com/kairos-io/kairos/issues/2429
- [x] https://github.com/kairos-io/kairos/issues/2138
- [x] https://github.com/kairos-io/kairos/issues/2217
- [x] https://github.com/kairos-io/kairos/issues/2430
- [x] https://github.com/kairos-io/kairos/issues/2048
- [x] https://github.com/kairos-io/kairos/issues/2547
- [x] https://github.com/kairos-io/kairos/issues/2586
- [x] https://github.com/kairos-io/kairos/issues/2608

:bug: 
- [x] https://github.com/kairos-io/kairos/issues/2526
- [x] https://github.com/kairos-io/kairos/issues/2212

Next release: https://github.com/kairos-io/kairos/issues/2052

## 🔦 Highlights

- This is the last release with Ubuntu 23.10 (marked for deprecation)

## ✅ Release Checklist

- [x] **Stage 0 - Finishing Touches**
    - [x] Check if Kairos-docs were updated and consider tagging them with the same version as Kairos
    - [x] Check if osbuilder is in the wanted version/latest
    - [x] Check if any kairos/packages was bumped and they were merged and repo updated (https://github.com/kairos-io/packages)
    - [x] Check latest repository update was merged, otherwise trigger its job (https://github.com/kairos-io/kairos/actions/workflows/bump_repos.yml)
    - [x] Make sure CI tests are passing.
    - [x] Consider cutting an `rc`, `alpha`, ... based on changes on the CI
- [x] **Stage 1 - Manual testing**
  - How: Using the assets from master, make sure that test scenarios not covered by automatic tests are passing, and that docs are still aligned
    - [x] Fedora flavor install, and manual upgrade works
    - [x] Any flavor interactive install
    - [x] Any flavor recovery reset
    - [x] ARM images (openSUSE, alpine) boots and manual upgrade works
    - [x] ARM images passive and recovery booting
    - [x] ARM images reset works
    - [x] ARM images /oem exists
- [x] **Stage 3 - Release**
  - [x] Tag the release on master.
[...]

PR #2705, opened 2024-07-09
Bump framework to v2.9.2

mauromorales:
https://github.com/kairos-io/kairos/issues/1792


mauromorales:
@Itxaka @mudler we already talked about this one before but I don't remember what was the resolution, are we ok with the broken Alpine encryption test? or should we fix this before 3.1?

mudler:
> @Itxaka @mudler we already talked about this one before but I don't remember what was the resolution, are we ok with the broken Alpine encryption test? or should we fix this before 3.1?

sounds good here - let's just not forget to mention it in the release notes, do we have already a card for it to track? that could go into a patch release too.

PR #2756, opened 2024-07-23
Remove ubuntu 23.10 from the pipelines

### LINKED ISSUE #2481: chore: Remove Ubuntu 23.10
Followup from #2138 

This ticket is about removing Ubuntu 23.10 from the pipelines so we don't produce new artifacts with it, but old artifacts will still be present in our repositories.

jimmykarily:
Fixes #2481

### DIFF
--- a/.github/flavors.json
+++ b/.github/flavors.json
@@ -29,7 +29,7 @@
         "arch": "arm64",
         "worker": "ubuntu-latest"
     },
-        {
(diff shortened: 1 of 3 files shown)


jimmykarily:
I should probably also remove it from the Dockerfiles. On it...

mauromorales (approved):

Issue #2773, opened 2024-07-29
uki installed system starts interactive-installer after first boot


jimmykarily:
Reproduce:

- Create a UKI iso with enki
- Boot it
- Complete the interactive installer to install the system
- Reboot

What happens:

Momentarily in tty1 the interactive installer is shown again. After a while it's replaced by a login prompt. The `kairos-agent state get boot` command says `active_boot`.

What was expected:

The interactive installer shouldn't be started at all

`/etc/os-release`:

```
kairos@localhost:~$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
KAIROS_FLAVOR="ubuntu"
KAIROS_RELEASE="v3.1.1-9-g496a7fad"
KAIROS_GITHUB_REPO="kairos-io/kairos"
KAIROS_SOFTWARE_VERSION_PREFIX="k3s"
KAIROS_ID_LIKE="kairos-core-ubuntu-22.04"
KAIROS_VERSION_ID="v3.1.1-9-g496a7fad"
KAIROS_PRETTY_NAME="kairos-core-ubuntu-22.04 v3.1.1-9-g496a7fad"
KAIROS_VARIANT="core"
KAIROS_VERSION="v3.1.1-9-g496a7fad"
KAIROS_ARTIFACT="kairos-ubuntu-22.04-core-amd64-generic-v3.1.1-9-g496a7fad"
KAIROS_FAMILY="ubuntu"
KAIROS_HOME_URL="https://github.com/kairos-io/kairos"
KAIROS_NAME="kairos-core-ubuntu-22.04"
KAIROS_MODEL="generic"
KAIROS_BUG_REPORT_URL="https://github.com/kairos-io/kairos/issues"
KAIROS_FLAVOR_RELEASE="22.04"
KAIROS_TARGETARCH="amd64"
[...]

PR #2784, opened 2024-08-02
:bug: Empty machine-id instead of removing it

Itxaka:
We used to rely on a hidden feature of sysinfo that created a filled machine-id during initramfs so systemd did not trigger the first boot service.

Now on the newer agent versions, that hidden functionality has gone away so if we remove it teh file, systemd will think that this is a new system and will do some work behind our back initializing and break things around.

So we need to create en empty machine-id file so systemd knows that we dont want to run the first boot services/target

Fixes https://github.com/kairos-io/kairos/issues/2773

### DIFF
--- a/images/Dockerfile.kairos
+++ b/images/Dockerfile.kairos
@@ -85,6 +85,12 @@ RUN luet database get-all-installed --output /etc/kairos/versions.yaml
 
 # TODO what about caches?
 RUN kernel=$(ls /lib/modules | head -n1) && depmod -a "${kernel}"
+
+# Set empty machine-id
+# This prevents systemd from thinking that the machine is on first boot
+# and recreating /etc/ dependencies in services and such
+# do this before initramfs so its in the initramfs
+RUN echo "" > /etc/machine-id || true
 # Regenerate initrd if necessary, proper config files with immucore and custom initrd should already be in there installed by framework
 # for systemd distros
 RUN if [ -f "/usr/bin/dracut" ]; then \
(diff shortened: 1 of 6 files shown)


jimmykarily:
I'd wish systemd had a way to disable the firstboot service without hacking with machine-id. But it is what it is I guess.

jimmykarily (approved):

Itxaka:
> I'd wish systemd had a way to disable the firstboot service without hacking with machine-id. But it is what it is I guess.

https://man.archlinux.org/man/machine-id.5.en#FIRST_BOOT_SEMANTICS

```
1.The kernel command argument systemd.condition_first_boot= may be used to override the autodetection logic, see [kernel-command-line(7)](https://man.archlinux.org/man/kernel-command-line.7.en).
2.Otherwise, if /etc/machine-id does not exist, this is a first boot. During early boot, systemd will write "uninitialized\n" to this file and overmount a temporary file which contains the actual machine ID. Later (after first-boot-complete.target has been reached), the real machine ID will be written to disk.
3.If /etc/machine-id contains the string "uninitialized", a boot is also considered the first boot. The same mechanism as above applies.
4.If /etc/machine-id exists and is empty, a boot is not considered the first boot. systemd will still bind-mount a file containing the actual machine-id over it and later try to commit it to disk (if /etc/ is writable).
5.If /etc/machine-id already contains a valid machine-id, this is not a first boot.
```

We could add the kernel cmdline as well to avoid this, but we will still have issues with the machine-id I think as we need to work around immutability and such

PR #2819, opened 2024-08-25
:book: Add alpine wifi cloud-config

jimmyjones2:

Adds a sample config for WiFi on alpine. Have noticed only Alpine & RPi openSUSE has wpa_supplicant - will update docs later in a separate PR and reference this config

I've never used connman before, so there could be a better way of doing this!


Itxaka (approved):

Itxaka:
and here again, thanks @jimmyjones2 !

PR #2846, opened 2024-09-06
Generate relative paths to files

jimmykarily:
I'm not sure why it worked in the past or why relative paths are ok but generating a sarif file locally and uploading it to https://sarifweb.azurewebsites.net/Validation
shows that the version generated by this commit is valid, while the old one was not:

image

Fixes #

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -4,7 +4,9 @@ FROM alpine:3.19
 # renovate: datasource=docker depName=quay.io/luet/base versioning=semver
 ARG LUET_VERSION=0.35.5
 # renovate: datasource=docker depName=aquasec/trivy versioning=semver
-ARG TRIVY_VERSION=0.53.0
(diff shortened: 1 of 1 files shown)


Itxaka (approved):

Earthfile
```diff
@@ -783,21 +785,25 @@ trivy-scan:
     SAVE ARTIFACT /build/results.json results.json AS LOCAL build/${ISO_NAME}-trivy.json
 
 grype:
-    FROM anchore/grype
+    ARG GRYPE_VERSION
```
Itxaka, line 788: Loool we were not using the pinned version?

mauromorales (approved):

Earthfile
jimmykarily: vivere pericolosamente :D

PR #2850, opened 2024-09-10
:robot: Make arm64 workers use docker mirror

### LINKED ISSUE #2838: Rate limit issues on pipeline
I'm not sure if dockerhub reduced their rate limits or if somehow we are just calling it more than before but it's been quite challenging to finish the release because of this:

```
logbus | Setting organization "" and project ""
              +all-arm | BASE_IMAGE=opensuse/tumbleweed FAMILY=opensuse FLAVOR=opensuse FLAVOR_RELEASE=tumbleweed K3S_VERSION=1.28.9 MODEL=rpi4 VARIANT=standard
              +all-arm | --> FROM +base
           alpine:3.19 | --> Load metadata alpine:3.19 linux/arm64
  Warning: you are not logged into registry-1.docker.io, you may experience rate-limitting when pulling images
              +all-arm | apply build +base: earthfile2llb for +base: Earthfile:2:0 apply FROM alpine:3.19: resolve image config for alpine:3.19: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/library/alpine/manifests/sha[25](https://github.com/kairos-io/kairos/actions/runs/10621879917/job/29463430008#step:12:27)6:95c16745f100f44cf9a0939fd3f357905f845f8b6fa7d0cde0e88c9764060185: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
              +all-arm | in		github.com/kairos-io/kairos:v3.1.2+base
  
  Help: DockerHub responded with a rate limit error. This is usually because you are not logged in.
  You can login using the command:
    docker login
```

Itxaka:

Fixes #2838

### DIFF
--- a/.github/workflows/reusable-docker-arm-build.yaml
+++ b/.github/workflows/reusable-docker-arm-build.yaml
@@ -105,6 +105,21 @@ jobs:
           registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
+      - name: Set local mirror
+        if: ${{ inputs.worker == 'ARM64' }}
+        run: |
+          # Configure earthly to use the docker mirror in CI
+          # https://docs.earthly.dev/ci-integration/pull-through-cache#configuring-earthly-to-use-the-cache
+          mkdir -p ~/.earthly/
+          cat << EOF > ~/.earthly/config.yml
+          global:
+            buildkit_additional_config: |
+              [registry."docker.io"]
+                mirrors = ["registry.docker-mirror.svc.cluster.local:5000"]
+              [registry."registry.docker-mirror.svc.cluster.local:5000"]
+                insecure = true
+                http = true
+          EOF
       - name: Set compression for PR
         if: ${{ github.event_name == 'pull_request' }}
         run: |
--- a/.github/workflows/reusable-docker-arm-build.yaml
+++ b/.github/workflows/reusable-docker-arm-build.yaml
@@ -25,7 +25,8 @@ on:
       base_image:
         required: true
         type: string
-
+env:
+  FORCE_COLOR: 1
 permissions: read-all
 jobs:
   build:


Itxaka:
canceled all non affected jobs. Will trigger this job several times: https://github.com/kairos-io/kairos/actions/runs/10788932103?pr=2850 so we can see if we hit a limit

.github/workflows/reusable-docker-arm-build.yaml
```diff
@@ -105,6 +106,21 @@ jobs:
           registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
+      - name: Set local mirror
+        if: ${{ inputs.worker == 'ARM64' }}
```
jimmykarily, line 110: Do we need a condition?

Itxaka: if we ever change workers to use like ubuntu-latest or something due to some issues with the current worker, this migth block us and we may forget it no? 

I just thougth that as in the past we had to quickly change the workers due to issues, we migth have to again in the future. I can drop it if you want?

Itxaka:
Seems to be hitting the mirror correctly :D

.github/workflows/reusable-docker-arm-build.yaml
jimmykarily: Cant' we have multiple labels on self hosted workers? E.g. add `self-hosted` to the arm64 ones so that we can select all self-hosted workers when needed. Because we could as well switch to non arm self-hosted runners in which case we would still want the mirror to work. Anyway, it's minor. We can improve it later.

jimmykarily (approved):

.github/workflows/reusable-docker-arm-build.yaml
Itxaka: I think they have a self-hosted label as well, so indeed we could and then this would not apply anymore I guess?

PR #2877, opened 2024-09-23
Test selecting disk by uuid+label

### LINKED ISSUE #1879: Failure to install to /dev/disk/by-id/<id>
Kairos fails to install to disk id (aka /dev/disk/by-id/<id>). I'm not sure if this is in the category of "not yet supported" or if this is a true bug.

Either way, when specifying to a disk id with the following or by setting such with the `--device` flag...
```bash
cat config.yaml
# #cloud-config
# hostname: nas-{{ trunc 4 .MachineID }}
# users:
# - name: clayton
#   groups: [ docker ]
#   shell: /usr/bin/fish
#   ssh_authorized_keys:
#     - github:clanktron
# 
# ssh_pwauth: false
# timezone: America/Los_Angeles
# 
# # kairos debug logging
# debug: true
# 
# install:
#   device: /dev/disk/by-id/ata-walram_60G_SCRW210906G0093
#   reboot: false
#   poweroff: true
#   auto: true
kairos-agent m config.yaml
```

The following errors occur: 
(apologies that these are screenshots and not raw text but it was going to be a hassle to grab the stdout/stderr of the install to the physical machine)
partition-error-1
partition-error-2

Looking at /dev/disk/by-id after the failure, it seems that the first partition is successfully created but that it has a different auto generated name than what kairos is attempting to find (see below).

successful-first-partition

I don't fiddle with udev rules a ton so I'm not sure how this partition auto naming scheme might vary from distro to distro if at all. For context this installer is running a lightly customized quay.io/kairos/core-ubuntu-22-lts:v2.4.0 image (just a few extra packages installed).

Itxaka:
Adds tests for the new functionality at https://github.com/kairos-io/kairos-agent/pull/552

Fixes #1879

Will get green once the PR is merged, new version released and new framework created and added as part of this PR

### DIFF
--- a/tests/install_target_test.go
+++ b/tests/install_target_test.go
@@ -1,0 +1,113 @@
+package mos_test
+
+import (
+	"fmt"
+	. "github.com/spectrocloud/peg/matcher"
+	"os"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("kairos install test different targets", Label("install-test-target"), func() {
+
+	var vm VM
+	BeforeEach(func() {
+
+		_, vm = startVM()
+		vm.EventuallyConnects(1200)
+		// Format the disk so it gets an uuid and label
+		_, err := vm.Sudo("mkfs.ext4 -L TESTDISK /dev/vda")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		Expect(vm.Destroy(nil)).ToNot(HaveOccurred())
+	})
+
+	Context("Selects the disk by uuid/label", func() {
+		It("Selects the correct disk if using uuid for target", func() {
+			expectSecureBootEnabled(vm)
+			// Get uuid of main disk
+			uuid, err := vm.Sudo("lsblk /dev/vda -o UUID -n")
+
+			cc := fmt.Sprintf(`#cloud-config
+				install:
+				  auto: true
+				  reboot: true
+				  device: /dev/disk/by-uuid/%s
+				
+				stages:
+				  initramfs:
+					- name: "Set user and password"
+					  users:
+						kairos:
+						  passwd: "kairos"
+					  hostname: kairos-{{ trunc 4 .Random }}
+				`, uuid)
+
+			By("Using the following config")
+			fmt.Fprintf(GinkgoWriter, cc)
+
+			t, err := os.CreateTemp("", "test")
+			ExpectWithOffset(1, err).ToNot(HaveOccurred())
+
+			defer os.RemoveAll(t.Name())
+			err = os.WriteFile(t.Name(), []byte(cc), os.ModePerm)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = vm.Scp(t.Name(), "/tmp/config.yaml", "0770")
+			Expect(err).ToNot(HaveOccurred())
+
+			var out string
+			// Test that install works
+			By("installing kairos", func() {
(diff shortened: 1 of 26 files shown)


Itxaka:
It passed when building with the new branch that brings this change, now that its removed, the test should fail

Itxaka:
now its pointing to master framework, so it should pass

tests/install_target_test.go
```diff
@@ -0,0 +1,151 @@
...
+				stateAssertVM(vm, "persistent.name", "/dev/vda")
+				stateAssertVM(vm, "state.mount_point", "/run/initramfs/cos-state")
+				stateAssertVM(vm, "oem.read_only", "false")
+				stateAssertVM(vm, "persistent.read_only", "false")
+				stateAssertVM(vm, "state.read_only", "true")
+				stateAssertVM(vm, "kairos.version", strings.ReplaceAll(strings.ReplaceAll(currentVersion, "\r", ""), "\n", ""))
+				stateContains(vm, "system.os.name", "alpine", "opensuse", "ubuntu", "debian")
+				stateContains(vm, "kairos.flavor", "alpine", "opensuse", "ubuntu", "debian")
+			})
+		})
```
jimmykarily, line 91: I don't see any check that the correct disk was actually used. We only check that the command didn't fail and that installation occurred. I guess, since we only have one disk, it was indeed selected but would be better if we had 2 disks and we made sure the correct one was selected, to avoid for example the flag being completely ignored and the default disk (bigger one?) gets selected instead.

But it can be tricky: https://github.com/kairos-io/kairos/blob/15f342f8cb3bcc8c00aba06dedcc9bc5362b7aea/tests/custom_partitioning_test.go#L68-L76

I don't have a strong opinion, I'll let you decide.

Itxaka: Yes indeed, that's a good point. I am letting it work as is because if you pass a wrong or nonexistent disk, the install fails but I'll have a look at peg to see if we can pass multiple disks to the VM, that would make it more robust

jimmykarily: https://github.com/spectrocloud/peg/commit/c12a7362c36c1c945ccc1c4e98e5ffee0808b7cb

jimmykarily: I introduced it for this exact reason. It's the boot order that is hard to do in qemu I think (see the link above).

Itxaka: but boot order should not matter no? It will try first all the disks and then livecd? Or it just tries the first disk only?

Itxaka: You are rigth, it doesnt try to boot from the second one somehow :sob:

Itxaka:
Now passing :D

tests/install_target_test.go
```diff
@@ -0,0 +1,168 @@
...
+		opts = append(opts, types.WithDriveSize("30000"))
+
+		m, err := machine.New(opts...)
+		Expect(err).ToNot(HaveOccurred())
+		vm = NewVM(m, stateDir)
+		_, err = vm.Start(context.Background())
+		Expect(err).ToNot(HaveOccurred())
+
+		vm.EventuallyConnects(1200)
+		// Format the second disk so it gets an uuid and label
```
jimmykarily, line 41: `/dev/vda` would be the "first" disk no?

tests/install_target_test.go
jimmykarily: Maybe we can add a default bootindex to all disks in qemu in peg, as in: first check disks (/dev/vda, /dev/vdb, etc) then cdrom.
This should allow us to try to target the second disk and see if it worked (since vda will be skipped as not being bootable).

jimmykarily (approved):

PR #2881, opened 2024-09-23
Actualizar Earthfile

Itxaka:


Itxaka:
This tests the new framework which has the new agent 2.14.x

Shows that there is a small issue with the new partitioner implementation and that its that we dont set the label of a fs. We set the partitionlabel but the kcrypt encryption needs the fs label.

PR #2885, opened 2024-09-24
:Robot: Cache triby DB before running the build

Itxaka:
If we hit a problem with the trivy download, we can make the step that downloads the database a retryable one so we dont stop the actual workflow

Fixes #

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -770,7 +770,7 @@ trivy:
     FROM aquasec/trivy:$TRIVY_VERSION
     RUN /usr/local/bin/trivy fs --download-db-only
     SAVE ARTIFACT /contrib contrib
-    SAVE ARTIFACT /root cache
+    SAVE ARTIFACT /root/.cache cache
     SAVE ARTIFACT /usr/local/bin/trivy /trivy
 
 trivy-scan:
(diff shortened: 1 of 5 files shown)


Earthfile
```diff
@@ -768,7 +768,9 @@ datasource-iso:
 trivy:
     ARG TRIVY_VERSION
     FROM aquasec/trivy:$TRIVY_VERSION
+    RUN /usr/local/bin/trivy fs --download-db-only
     SAVE ARTIFACT /contrib contrib
+    SAVE ARTIFACT /root cache
```
jimmykarily, line 773: maybe you meant a specific directory in `/root` here?

Itxaka: well, indeed its `/root/.cache` the one with the actua DB but I thought it would be helpful to capture it all just in case there is more stuff needed in there?

jimmykarily: but then you don't want to copy it at `/root/.cache` below, no? I mean, you would be copying `/root` in `/root/.cache` which would result in a directory structure like `/root/.cache/.cache`.

Itxaka: ummm...good point lol I did nothing!

PR #2889, opened 2024-09-25
:robot: Revert the trivy DB changes

Itxaka:
While locally it works as it caches the image witht he DB already on it, on CI it doesnt work and pulls the DB twice, so its the same as it was before.

### DIFF
--- a/.github/workflows/release-arm.yaml
+++ b/.github/workflows/release-arm.yaml
@@ -247,13 +247,6 @@ jobs:
           registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
-      - name: Cache trivy DB
-        uses: nick-fields/retry@v3
-        with:
-          max_attempts: 3
-          retry_on: error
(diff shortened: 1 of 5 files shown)


jimmykarily:
Strange though, why? Do we declare defeat?

Itxaka:
> Strange though, why? Do we declare defeat?

I do.

Itxaka:
mmmh. Maybe we can do this separatedly instead.

 - build image or whatever
 - then scan with trivy the docker image generated, should be the same as scanning the fs inside the image

so the scan can just have a retry AND maybe we can just use the default trivy action from github which Im sure its leass problematic

and less code in earthly

mauromorales:
Would it make sense to try something like https://github.com/marketplace/actions/cache-trivy ?

jimmykarily:
> Would it make sense to try something like https://github.com/marketplace/actions/cache-trivy ?

Moved to another ticket: https://github.com/kairos-io/kairos/issues/2904
Let's merge this revert and work on it on the new ticket.

jimmykarily (approved):

Issue #2716, opened 2024-07-11
Broken rngd.service on Tumbleweed rpi


mauromorales:
Tumbleweed on RPi has a broken service, but everything else seems to be working fine

```
localhost:/home/kairos # systemctl status --failed
x rngd.service
     Loaded: not-found (Reason: Unit rngd.service not found.)
     Active: failed (Result: exit-code) since Mon 2024-06-24 14:36:23 UTC; 2 weeks 2 days ago
   Duration: 272ms
   Main PID: 253 (code=exited, status=1/FAILURE)
        CPU: 65ms

Jun 24 14:36:23 localhost rngd[253]: Can't open any entropy source
Jun 24 14:36:23 localhost rngd[253]: Maybe RNG device modules are not loaded
Jun 24 14:36:23 localhost rngd[253]: Initializing available sources
Jun 24 14:36:23 localhost rngd[253]: [hwrng ]: Initialization Failed
Jun 24 14:36:23 localhost rngd[253]: [rndr  ]: No HW SUPPORT
Jun 24 14:36:23 localhost rngd[253]: [rndr  ]: Initialization Failed
Jun 24 14:36:23 localhost rngd[253]: [pkcs11]: Unable to load pkcs11 engine: (null)
Jun 24 14:36:23 localhost rngd[253]: [pkcs11]: Initialization Failed
Jun 24 14:36:23 localhost rngd[253]: Can't open any entropy source
Jun 24 14:36:23 localhost rngd[253]: Maybe RNG device modules are not loaded
```

**Kairos version:**

```
cat /etc/os-release
NAME="openSUSE Tumbleweed"
# VERSION="20240629"
ID="opensuse-tumbleweed"
ID_LIKE="opensuse suse"
VERSION_ID="20240629"
PRETTY_NAME="openSUSE Tumbleweed"
ANSI_COLOR="0;32"
# CPE 2.3 format, boo#1217921
CPE_NAME="cpe:2.3:o:opensuse:tumbleweed:20240629:*:*:*:*:*:*:*"
#CPE 2.2 format
#CPE_NAME="cpe:/o:opensuse:tumbleweed:20240629"
BUG_REPORT_URL="https://bugzilla.opensuse.org"
SUPPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org"
DOCUMENTATION_URL="https://en.opensuse.org/Portal:Tumbleweed"
LOGO="distributor-logo-Tumbleweed"
[...]

PR #2890, opened 2024-09-25
Don't run both rngd and haveged

jimmykarily:
because they clash. We keedp haveged only because we can't know if there is special rng hardware or not.

Fixes https://github.com/kairos-io/kairos/issues/2716


Itxaka (approved):
Nice catch!

Issue #2904, opened 2024-09-30
Try to cache trivy db


jimmykarily:
Let's give this a try: https://github.com/kairos-io/kairos/pull/2889#issuecomment-2376371602

TL;DR; Give this a try: https://github.com/marketplace/actions/cache-trivy

or find some other way to avoid the TOO MANY REQUESTS error when pulling trivy db

jimmykarily:
Maybe this: https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runspost
here: https://github.com/yogeshlonkar/trivy-cache-action/blob/5e3c2d406dd3a13d95a3eabc520e0ac19ab729b5/action.yml#L20
?

PR #2910, opened 2024-10-01
Cache trivy

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/2904

Fixes #

### DIFF
--- a/.github/workflows/image-pr.yaml
+++ b/.github/workflows/image-pr.yaml
@@ -12,9 +12,31 @@ env:
   FORCE_COLOR: 1
   EARTHLY_TOKEN: ${{ secrets.EARTHLY_TOKEN }}
 jobs:
+  # Populate the trivy cache once for all later jobs to use
+  trivy-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+        with:
+          fetch-depth: 0
+      - name: Install earthly
+        uses: Luet-lab/luet-install-action@cec77490c3f2416d7d07a47cfab04d448641d7ce # v1.1
+        with:
+          repository: quay.io/kairos/packages
+          packages: utils/earthly
+      - name: Restore trivy cache
+        uses: yogeshlonkar/trivy-cache-action@v0
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Populate trivy Cache
+        run: |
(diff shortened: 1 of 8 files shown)


jimmykarily:
I don't understand at which point the cache is saved. Looking at this pipeline: https://github.com/yogeshlonkar/trivy-cache-action/blob/5e3c2d406dd3a13d95a3eabc520e0ac19ab729b5/.github/workflows/workflow.yml#L48
I think we may need to run it before our step to restore it and after it to save it. But I'm trying to verify that it works like that in code.

.github/workflows/image-pr.yaml
```diff
@@ -12,9 +12,31 @@ env:
   FORCE_COLOR: 1
   EARTHLY_TOKEN: ${{ secrets.EARTHLY_TOKEN }}
 jobs:
+  # Populate the trivy cache once for all later jobs to use
+  trivy-cache:
```
jimmykarily, line 16: I need to add this "top level" job in the other pipelines too.

jimmykarily: done

jimmykarily:
> I don't understand at which point the cache is saved. Looking at this pipeline: https://github.com/yogeshlonkar/trivy-cache-action/blob/5e3c2d406dd3a13d95a3eabc520e0ac19ab729b5/.github/workflows/workflow.yml#L48 I think we may need to run it before our step to restore it and after it to save it. But I'm trying to verify that it works like that in code.

Found it.

So this solution works like this:

At the beginning of the pipeline and before any workflows request the trivy database:

- We run the `yogeshlonkar/trivy-cache-action@v0` action first to populate the `.trivy` directory from the GitHub cache. The cache key [includes the db hash and the branch name](https://github.com/yogeshlonkar/trivy-cache-action?tab=readme-ov-file#cache-scopes) so it's valid for the same branch for as long as the db is fresh
- We run the new earthly target `trivy-download-db` to populate the `.trivy` directory (either because no cache was found in the previous step or because there is a fresh db to be downloaded)
- The action [has a "post" step](https://github.com/yogeshlonkar/trivy-cache-action/blob/5e3c2d406dd3a13d95a3eabc520e0ac19ab729b5/action.yml#L20) which pushes the contents of the possibly updated `.trivy` directory back to GitHub cache.

After this is done, we are sure we have populated the GitHub cache with the latest db. From there on, any workflow that needs the trivy cache, will run the `trivy-cache-action` to populate the `.trivy` dir and pass it to the earthly targets using the new `TRIVY_CACHE_DIR` argument.

mauromorales (approved):

PR #2925, opened 2024-10-08
Run trivy-cache job on ARM64 worker

jimmykarily:
because we suspect this is the reason why later jobs that run on those workers don't find the cache.

Might be related to:
https://github.com/actions/cache/blob/main/tips-and-workarounds.md#cross-os-cache

Fixes #


jimmykarily:
seems to work

PR #2930, opened 2024-10-09
Add ubuntu 24.10 flavor

mauromorales:
24.10 has already been available for a bit, and it would be nice to already include it in our pipeline

This PR also cleans up a bit the order of the flavors to always go from newest to oldest


Itxaka (approved):
uki build?

PR #2940, opened 2024-10-14
:robot: Build Ubuntu 24.04 Standard Image UKI on Github

bencorrado:
Until the standard packages can be auto installed as sysext I think this gives a way to work with standard UKI images.  I can open a ticket on this for the future.  This will help me to build with the released file sets and test changes more quickly.

Discussed at https://cloud-native.slack.com/archives/C0707M8UEU8/p1727801588480279?thread_ts=1727801588.480279&cid=C0707M8UEU8

I don't have a great way to set all of this without the full CI/CD access, so if someone has a few minutes to review before accepting that would be great.

I changed the logging filename to include the variant, so they do not overwrite.  I could not find anything this affects, but I might not have full visibility.

### DIFF
--- a/.github/workflows/reusable-uki-test.yaml
+++ b/.github/workflows/reusable-uki-test.yaml
@@ -64,7 +64,7 @@ jobs:
             --FLAVOR_RELEASE=${{ inputs.flavor_release }} \
             --FAMILY=${{ inputs.family }} \
             --MODEL=generic \
-            --VARIANT=core \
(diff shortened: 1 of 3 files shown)


Itxaka:
I think @jimmykarily was also working on this?

Itxaka:
umm, I think this will never pass due Ben not having permissions to see the earthly token. Merging this directly, will fix it if we broke it somehow

PR #3009, opened 2024-11-13
:robot: Change trivy DB url

Itxaka:
To a one that seems to not have requests limits in place which means we can drop the workarounds we have for the failure to download

Fixes #


jimmykarily (approved):

Itxaka:
seems to work as expected, lets see for how long :D

PR #3067, opened 2024-12-10
Bump quic-go to fix security scanner

jimmykarily:
E.g.: https://github.com/kairos-io/kairos/actions/runs/12252916516/job/34180533523?pr=3066

Fixes #


Itxaka (approved):

jimmykarily:
Turns out it's not that simple. `edgevpn` breaks with this bump and fixing that would require bumping go-libp2p in edgevpn which it not a trivial fix.

We could get rid of the [edgvpn dependency in tests](https://github.com/kairos-io/kairos/blob/8d6442665b3170042833084a191c66f624a364c2/tests/tests_suite_test.go#L221) if we simply called edgevpn with docker to generate a token.

E.g.

```
// Generates a valid token for provider tests
func generateToken() string {
	cmd := exec.Command("docker", "run", "--rm", "quay.io/mudler/edgevpn:latest", "-g", "-b")
	out, err := cmd.CombinedOutput()
	Expect(err).ToNot(HaveOccurred())

	return string(out)
}
```

but we still get edgevpn pulled as a dependency by go-nodepair used [here](https://github.com/kairos-io/kairos/blob/8d6442665b3170042833084a191c66f624a364c2/tests/tests_suite_test.go#L255).

That one we could also get rid of by calling `kairos-agent register` in docker but I can't decide whether I like were this is going.

After all, these are all repositories that will need to be fixed at some point. Maybe we should focus on that fix instead.

Issue #2709, opened 2024-07-10
re-introduce provider-decentralized-k8s tests


mauromorales:
It's been a while since we ran the provider-decentralized-k8s. We remove them from CI because they were always breaking but they used to run fine on local. Maybe we can use the CNCF nodes and see if they get green

https://github.com/kairos-io/kairos/blob/master/.github/workflows/image.yaml#L375-L378

https://github.com/kairos-io/kairos/blob/master/.github/workflows/release-arm.yaml#L307

Itxaka:
https://github.com/kairos-io/kairos/pull/2730

problem comes from edgevpn

testing 0.23.1 works
testing 0.24.1 doesnt work

Error happens when connecting via dht:
```
[...]

PR #3078, opened 2024-12-12
Log files no longer have hardcoded names, use a glob for the test

jimmykarily:
This test is still run manually and fails later when it tries to upgrade to:

quay.io/kairos/kairos-opensuse:v1.0.0-rc2-k3sv1.21.14-k3s1

but I don't even know why it should be able to "upgrade" (actually downgrade) to such an old version successfully. Manually upgrading works.

This fix is meant to make it easier to run the test locally until we can run it again on CI (https://github.com/kairos-io/kairos/issues/2709).

Fixes #


jimmykarily:
Merging directly. Only changes one test which we don't even run on CI

PR #3110, opened 2025-01-09
Use the generic core matrix for the generic images

mauromorales:
They were not running, not sure why it shows the error, but it should be pointing to the generic matrix anyways

Screenshot 2025-01-09 at 10 07 13


jimmykarily (approved):

Itxaka:
Mauro fixing what some idiot broke, thanks @mauromorales !

PR #3111, opened 2025-01-09
Use the generic core matrix for the generic images

mauromorales:
https://github.com/kairos-io/kairos/actions/runs/12687897629/job/35363289989#step:8:8


mauromorales:
opps wrong branch

PR #3130, opened 2025-01-16
Test kairos-init

Itxaka:

Fixes #


jimmykarily (commented):
Nice job!

I don't see anything bad in it, but I haven't done any manual testing on the resulting images yet. I think we should test this a bit before merging.

images/Dockerfile
```diff
@@ -0,0 +1,26 @@
+ARG BASE_IMAGE=ubuntu:24.04
+ARG VARIANT=core
+
+FROM ttl.sh/kairos-init:main AS kairos-init
```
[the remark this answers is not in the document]
Itxaka: Indeed as many changes are being done to the init as tests uncover missing things I keep pushing new images to quickly test it. Also I it currently only pushes to TTL as I don't want to pollute quay too much with all the testing

images/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN /kairos-init -l debug -s init -m "${MODEL}" -v "${VARIANT}" -t "${TRUSTED_BOOT}"
+#RUN /kairos-init -l debug --validate -m "${MODEL}" -v "${VARIANT}" -t "${TRUSTED_BOOT}"
+RUN rm /kairos-init
+
+
+FROM base AS core
+
+FROM base AS standard
+COPY --from=luet /usr/bin/luet /usr/bin/luet
+RUN luet install -y k8s/k3s-$(grep -q alpine /etc/os-release && echo "openrc" || echo "systemd")
```
mauromorales, line 25: can we have here a variable to build also k0s? ideally something like `KUBERNETES_DISTRO` or such?

mauromorales: also, isn't it missing @${SOFTWARE_LUET_VERSION:-$SOFTWARE_VERSION} ? or am I missing something

.github/workflows/reusable-zfs-test.yaml
```diff
@@ -9,31 +9,61 @@ on:
...
+        with:
+          name: kairos-${{ inputs.flavor }}-${{ inputs.flavor_release }}-${{ inputs.variant}}.iso.zip
+      - name: Run tests
+        env:
+          USE_QEMU: true
+          KVM: true
+          MEMORY: 4000
+          CPUS: 2
+          DRIVE_SIZE: 30000
+          CONTAINER_IMAGE: ttl.sh/kairos-${{ inputs.flavor }}-${{ inputs.flavor_release }}-${{ inputs.variant }}-${{ github.sha }}:24h
```
mauromorales, line 57: do we need to add the ARCH here too? to avoid them steping on each others toes? or is arm never tested?

.github/workflows/reusable-upgrade-with-cli-test.yaml
```diff
@@ -80,7 +43,7 @@ jobs:
       - run: |
           earthly +run-qemu-test --PREBUILT_ISO=$(ls *${{ inputs.flavor }}*.iso) \
             --FLAVOR=${{ inputs.flavor }} \
-            --CONTAINER_IMAGE=ttl.sh/kairos-${{ inputs.flavor }}-${{ inputs.flavor_release }}-${{ github.sha }}:24h \
+            --CONTAINER_IMAGE=ttl.sh/kairos-${{ inputs.flavor }}-${{ inputs.flavor_release }}-${{ inputs.variant }}-${{ github.sha }}:24h \
```
mauromorales, line 46: same question about arm

.github/workflows/reusable-build-flavor.yaml
```diff
@@ -35,199 +35,45 @@ jobs:
...
-          # cosign sign $image_ref --attachment sbom
-          # in-toto attestation
-          # cosign attest --type spdx --predicate $spdx $image_ref
       - name: Push to testing
         run: |
-          _IMG=$(cat IMAGE)
-          TEST_IMAGE="ttl.sh/kairos-${{ inputs.flavor }}-${{ inputs.flavor_release }}-${{ github.sha }}:24h"
-          docker tag $_IMG $TEST_IMAGE
+          TEST_IMAGE="ttl.sh/kairos-${{ inputs.flavor }}-${{ inputs.flavor_release }}-${{ inputs.variant }}-${{ github.sha }}:24h"
+          docker tag ${{ inputs.flavor }}-init:${{ inputs.flavor_release }} $TEST_IMAGE
```
mauromorales, line 78: I think the same is used for the `uki` test, so I think it might be problematic if the two are running at the same time, no?

mauromorales (commented):
Screenshot 2025-01-30 at 21 04 25 😍

.github/workflows/reusable-build-flavor.yaml
Itxaka: Umm good point on the naming, indeed. I need to check this further but the idea would be to push the same image for both arches so they should share the name. Unfortunately I less you build it at the same time, they step over each other so not there yet. Plus this affects only PR and master I think, release still do its thing.

Also I think that we don't push the images or at least that's the idea, everything local instead of wasting time pushing and pulling unless it's necessary 

I will have a look at this stuff

images/Dockerfile
```diff
@@ -0,0 +1,26 @@
+ARG BASE_IMAGE=ubuntu:24.04
```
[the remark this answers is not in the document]
mauromorales: I'd just like to point out that `ubuntu:20.04` should be the default 🤣

images/Dockerfile
Itxaka: yes, it was discussed and currently we will NOT provide the way of building k3s/k0s bundled in the init, it has to be done by the user when building the image. This is easy as we have the frameowork so its just a simple luet install VERSION away.

I guess we can fix this down the line if we want to provide k3s/k0s support directly on the init part?

.github/workflows/reusable-zfs-test.yaml
Itxaka: Remove from here as its not necessary, but added arches around on things that need them, like upgrade tests

images/Dockerfile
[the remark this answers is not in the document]
Itxaka: will do before merge to avoid any issues. In any case, this is here becuase Docker requires it, but calling it without a base image should actually fail the process, but I could not make it work as is

.github/workflows/reusable-upgrade-with-cli-test.yaml
Itxaka: done

.github/workflows/reusable-build-flavor.yaml
Itxaka: done and done, now it has arches in them for the normal jobs in which we need to push them to be consumed. For uki, we dont push those init images so its not necessary

tests/install_target_test.go
```diff
@@ -44,7 +45,19 @@ var _ = Describe("kairos install test different targets", Label("install-test-ta
 	})
 
 	AfterEach(func() {
+		if CurrentSpecReport().Failed() {
+			serial, _ := os.ReadFile(filepath.Join(vm.StateDir, "serial.log"))
+			_ = os.MkdirAll("logs", os.ModePerm|os.ModeDir)
+			_ = os.WriteFile(filepath.Join("logs", "serial.log"), serial, os.ModePerm)
```
jimmykarily, line 51: do we need to `.gitignore` this?

jimmykarily (approved):

tests/install_target_test.go
Itxaka: we never did anywhere I think but it will be good for local testing?

PR #3142, opened 2025-01-24
Allow building of k0s images via earthly

mauromorales:
Bumps the version to one that handles k0s through the provider and allows passing of the software prefix and build version to create k0s images. By default it keeps the k3s functionality

### DIFF
--- a/images/Dockerfile.kairos
+++ b/images/Dockerfile.kairos
@@ -42,7 +42,7 @@ FROM base-kairos AS kairos-standard
 ARG SOFTWARE_VERSION
 LABEL io.kairos.k3s_version="${SOFTWARE_VERSION}"
 RUN luet install -y system/provider-kairos
-RUN luet install -y "k8s/k3s-$(which-init.sh)@${SOFTWARE_LUET_VERSION:-$SOFTWARE_VERSION}" utils/edgevpn utils/k9s utils/nerdctl container/kubectl utils/kube-vip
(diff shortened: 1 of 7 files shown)


images/Dockerfile.kairos-debian
```diff
@@ -114,10 +113,17 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     zerofree \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
+###############################################################
```
mauromorales, line 116: this change is not from this PR but from https://github.com/kairos-io/kairos/pull/3084/files

jimmykarily: rebase?

images/Dockerfile.kairos
```diff
@@ -42,7 +42,7 @@ FROM base-kairos AS kairos-standard
 ARG SOFTWARE_VERSION
 LABEL io.kairos.k3s_version="${SOFTWARE_VERSION}"
 RUN luet install -y system/provider-kairos
-RUN luet install -y "k8s/k3s-$(which-init.sh)@${SOFTWARE_LUET_VERSION:-$SOFTWARE_VERSION}" utils/edgevpn utils/k9s utils/nerdctl container/kubectl utils/kube-vip
+RUN luet install -y "k8s/${SOFTWARE_VERSION_PREFIX}-$(which-init.sh)@${SOFTWARE_LUET_VERSION:-$SOFTWARE_VERSION}" utils/edgevpn utils/k9s utils/nerdctl container/kubectl utils/kube-vip
```
jimmykarily, line 45: So it's no more just a "prefix". It actually defines the k8s distribution. We should consider renaming it in the future (preferably when we have gotten rid of Earhtly, to make change it in less places).

jimmykarily (approved):

images/Dockerfile.kairos
mauromorales: yup makes sense

PR #3154, opened 2025-01-29
Try to avoid complicated logic to select k3s version

jimmykarily:

Fixes #

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -322,6 +322,37 @@ jobs:
           IMAGE=$(cat release/IMAGE)
           docker push "$IMAGE"
           cosign sign $(docker image inspect --format='{{index .RepoDigests 0}}' "$IMAGE")
+
+      # https://github.com/aws-actions/configure-aws-credentials?tab=readme-ov-file#assumerole-with-static-iam-credentials-in-repository-secrets
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: eu-central-1
+          #role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          #role-external-id: ${{ secrets.AWS_ROLE_EXTERNAL_ID }}
+          role-duration-seconds: 1200
+          role-session-name: AWSCIPush
+      - name: Build and push AWS image
+        if: matrix.flavor == 'ubuntu' && matrix.flavorRelease == '24.04' && matrix.variant == 'standard' && matrix.model == 'generic' && matrix.arch == 'amd64'
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -322,6 +322,37 @@ jobs:
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: eu-central-1
+          #role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          #role-external-id: ${{ secrets.AWS_ROLE_EXTERNAL_ID }}
+          role-duration-seconds: 1200
+          role-session-name: AWSCIPush
+      - name: Build and push AWS image
+        if: matrix.flavor == 'ubuntu' && matrix.flavorRelease == '24.04' && matrix.variant == 'standard' && matrix.model == 'generic' && matrix.arch == 'amd64'
```
jimmykarily, line 338: This will match all 3 k3s versions. I don't see any way to upload only one without hardcoding it (or doing the things I deleted in this PR).

The other option is to upload the code image but if the user tries to "reset" to a standard one using userdata (which is a very possible thing), the partitions won't be big enough for the standard image. They can specify sizes for the partitions I think but the recovery one can't be resized cause it's what is booted that point.

Any ideas?

.github/workflows/release.yaml
```diff
@@ -340,94 +371,3 @@ jobs:
...
-              {
-                major: (.major | tonumber),
-                minor: (.minor | tonumber),
-                patch: (.patch | tonumber),
-                is_stable: (if .pre == null then 1 else 0 end),
-                pre: (.pre // "")
-              }
-            ) | max_by(.major, .minor, .patch, .is_stable, .pre) |
-              "\(.major).\(.minor).\(.patch)\(.pre | if . == "" then "" else "-"+. end)"' | \
-            sed -E 's/^/v/; s/$/-k3s1/')
```
jimmykarily, line 389: This "worked" kind of, but the artifacts we push are named:

```
 quay.io/kairos/ubuntu:24.10-standard-amd64-generic-v3.3.1-rc2-k3sv1.32.1-rc2-k3sk3s1
```
and this script was looking for:

```
quay.io/kairos/ubuntu:24.04-standard-amd64-generic-v3.3.1-rc2-k3sv1.32.1-rc2-k3s1
```

I think the first one is wrong which is a different issue to look at but in any  case, I thought it would be better if we didn't try to construct the image name again so this PR tries to push the AWS at a point where the image name is still available.

.github/workflows/release.yaml
jimmykarily: We discussed this in Slack and decided to just push a core image. We will then implement a flag in auroraboot to set the "minimum image size" which will override the calculated one when building the cloud image. This way we can set this to a hardcoded value that can fit all our image (standard or core). TODO: Add ticket link

.github/workflows/release.yaml
jimmykarily: https://github.com/kairos-io/kairos/pull/3156

Itxaka (approved):

PR #3160, opened 2025-01-30
feat(nvidia): add missing packages

mudler:

I don't think this fixes the issues I'm seeing, however I'm comparing the list of packages with the standard installation and these seems to be missing.

Fixes #


mudler:
closing as turns out was not necessary. let's keep the images small and usable as a good base for customization

PR #3170, opened 2025-02-06
Rework tests

Itxaka:

- Reduces complexity in our test pipelines by removing unneeded things
- Merges qemu tests into a single reusable job
- Merges flavor build into a single job (includes different arches, variants, models and outputs like RAW or ISO)
- Makes tests create their datasource ISOs instead of relaying in external things to pass them.
- Tries to get rid of the complex and blackbox matrix output of complex jobs. Having a simple declarative entries is nicer for now and makes it more clear whats gonna be build and why
- Runs jobs in its own workers depending on arch. If you are running an arm build it will be run in an arm worker.
- Tries to reduce the inputs to reusable workflows to a base_image only. We should be able to extract the needed data from the base_image only and the kairos-init can get all the info it needs from the running system so most of the data is not needed.
- Sets some defaults for inputs to reusable workflows so just calling them with base_image should be enough
- Tries to groups jobs under the same base image so its easier to check them in the workflow page
- Removes unneeded pushes of intermediate images. Now they are done locally and pushed at the end if needed
- Removes master pushes to official artifacts. I dont think its needed nor they give us anything special, its just wasted time and another point of failure.
- Release its now a reusable job that can be used by any type of arch,variant,model or base_image.

Future idea would be once, this is merged and streamlined as much as possible, we bring back a flavorsV2.json with a more compact, explicit way of setting the flavors and how we want them build, what arches, what models and so on. But for that, first we need to have a single pipeline to build everything so the inputs can be the same for all different stuff.

Final approach should be something like 1 reusable workflow for building the artifacts no matter what, 1 workflow for building master and PRs (event_type is the one that triggers some flavors or others) and 1 workflow for Release (mainly becuase I like to have Release on its own workflow for clarity as its the most important workflow and the most critical) but could easily have a single workflow for all 3 events (PR, master, release) and 1 reusable workflow for the Test.


.github/workflows/image-master.yaml
```diff
@@ -52,329 +30,110 @@ jobs:
...
-      - core
-    strategy:
-      fail-fast: true
       matrix:
-        include:
-          - flavor: opensuse
-            flavorRelease: leap-15.6
-            secureboot: false
-  install-target:
+        variant: ["core", "standard"]
```
mauromorales, line 42: I think this used to be only core before, just mentioning in case adding standard wasn't intentional

.github/workflows/reusable-build-flavor.yaml
```diff
@@ -3,38 +3,46 @@ name: Reusable workflow that builds a specific Kairos flavor
...
+        type: string
+      event_type:
         required: true
         type: string
 
 permissions: read-all
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ inputs.arch == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
```
mauromorales, line 31: what does this do? if arm64 it will output `ubuntu-24.04-arm` otherwise `ubuntu-24.04` ?

.github/workflows/release.yaml
```diff
@@ -1,355 +1,106 @@
...
-        if: startsWith(github.ref, 'refs/tags/')
-        with:
-          sarif_file: 'grype-results'
-          category: ${{ matrix.flavor }}-${{ matrix.flavor_release }}-${{ matrix.variant }}-${{ matrix.arch }}-${{ matrix.model }}-grype
+      matrix:
+        model: ["generic"]
+        variant: ["core"]
+        arch: ["amd64"]
+        base_image:
+          - "opensuse/leap:15.6"
```
mauromorales, line 23: why not use flavors.json? We might forget changes here if we are touching master, or viceversa

.github/workflows/reusable-build-flavor.yaml
Itxaka: yep! So each job runs in its proper runnner arch depending on its own arch.

Arm builds will run in the arm workers and amd ones in the normal ones :D

.github/workflows/release.yaml
Itxaka: because it sucks. There is a lot of duplication, its hidden behind a esoteric job that outputs magic and its complex.

I do want to bring it back but in a simpler way, i.e. with entries to mark if it runs on pr,master,release, models and such


So basically a merged one in which an entry like

```
{
  "base_image":  "ubuntu:24.04",
  "PR": true,
  "Master": true,
  "Release": true,
  "Arch": ["amd64, "arm64"],
  "variant": ["core", "standard"]
  "models": ["generic"]
}

And that would generate the matrix for all jobs, will build on PR,Master. Will Release artifacts for it. Will build core and standard. Will build both arches. Will build generic model only.


The idea is to try to merge everything as close as possible in order to centralize more things, but the move to flavorsV2.json will need to come later, once we find how shared things can be between the different jobs and such :D

File: .github/workflows/image-master.yaml (hunk 1 above)
Comment (Itxaka) [reply .github/workflows/image-master.yaml:42]: on master it run standard as well, no idea why. I would change this to be PR tests stuff (both core and standard), you cant merge with an outdated branch, and make master only build the artifacts of all flavors for testing the build part, but the full test suite I would do PR only. Once its merged we tend to forget about it.

File: .github/workflows/release.yaml (hunk 3 above)
Comment (mauromorales) [reply .github/workflows/release.yaml:23]: noice, makes sense ... we should not let that come too late though, otherwise I can see us releasing and having to do a patch release to add a missing flavor or so haha

File: .github/workflows/release.yaml [hunk 4]
```diff
@@ -1,355 +1,106 @@
...
-        env:
-          COSIGN_YES: true
-        run: |
-          SUFFIX="-uki"
-          IMAGE=$(cat IMAGE)
-          docker tag $(cat IMAGE) "$IMAGE$SUFFIX"
-          docker push "$IMAGE$SUFFIX"
-          image_ref=$(docker image inspect --format='{{index .RepoDigests 0}}' "$IMAGE$SUFFIX")
-          cosign sign $image_ref
+          kubernetes_versions=$(curl -s https://api.github.com/repos/k3s-io/k3s/releases | jq -r '[.[] | select(.prerelease == false and .draft == false and (.tag_name | test("rc") | not))] | .[:3] | .[].tag_name' | tr '\n' ' ')
```
mauromorales, line 72: I think this should not come from the k3s repo but from the luet releases

mauromorales (approved):
LGTM

.github/workflows/reusable-release.yaml
```diff
@@ -0,0 +1,158 @@
...
+          fetch-depth: 0
+      - name: Split base image
+        id: split
+        run: |
+          echo "flavor=$(echo ${{ inputs.base_image }} | cut -d ':' -f 1 | sed  's/\//-/g')" >> $GITHUB_OUTPUT
+          echo "flavor_release=$(echo ${{ inputs.base_image }} | cut -d ':' -f 2)" >> $GITHUB_OUTPUT
+      - name: Set Image Tag
+        run: |
+          echo "GIT_VERSION=$(git describe --always --tags --dirty)" >> $GITHUB_ENV
+          echo "IMAGE_TAG=quay.io/kairos/${{ steps.split.outputs.flavor }}:${{ steps.split.outputs.flavor_release }}-${{ inputs.variant }}-${{ inputs.arch }}-${{ inputs.model }}-${{ env.GIT_VERSION }}${{ inputs.kubernetes_version && '-'||''}}${{ inputs.kubernetes_version || ''}}${{ inputs.trusted_boot != 'false' && '-uki' || '' }}" >> $GITHUB_ENV
```
jimmykarily, line 50: Are we deprecating versioneer too? Because it's still used in `kairos-agent upgrade list-releases` and was also used when building the images, in order to prepare the values for `/etc/kairos-release`. If we still use it there, it would be good if the same tool was used everywhere to make sure we produce consistent names.

Another option would be to first build the image, then extract the artifact name from `/etc/kairos-release` in another step (from within the image) and finally push it with that name. But this creates a slightly more complicated workflow.

jimmykarily: oh I just saw, we already push in a second step so maybe the second option is not that more complicated.

.github/workflows/reusable-uki-test.yaml
```diff
@@ -121,38 +114,38 @@ jobs:
...
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && inputs.flavor == 'ubuntu' }}
         env:
           COSIGN_YES: true
         run: |
-          SUFFIX="-uki"
-          IMAGE=$(FLAVOR=${{ inputs.flavor }} FLAVOR_RELEASE="${{ inputs.flavor_release }}" MODEL=generic TARGETARCH=amd64 VARIANT=${{ inputs.variant }} REGISTRY_AND_ORG="quay.io/kairos" RELEASE=master kairos-agent versioneer container-artifact-name)
-          docker tag ${{ inputs.flavor }}-init:${{ inputs.flavor_release }} "$IMAGE$SUFFIX"
-          docker push "$IMAGE$SUFFIX"
-          image_ref=$(docker image inspect --format='{{index .RepoDigests 0}}' "$IMAGE$SUFFIX")
+          docker tag ${{ steps.split.outputs.flavor }}-init:${{ steps.split.outputs.flavor_release }} ${{ env.IMAGE_TAG }}
```
jimmykarily, line 142: Was it on purpose that this testing pipeline is pushing an image to `quay.io` ? All images to quay.io should come from the release pipeline I think (since they are, well, "released"). Unless we create some other org for temporary images like those we send to ttl.sh.

Earthfile
jimmykarily, line 1: :man_dancing:

Earthfile
```diff
@@ -1,1085 +0,0 @@
...
-  BUILD --platform=linux/arm64 +iso --MODEL=generic
-
-build-and-push-golang-testing:
-    ARG GO_VERSION
-    FROM golang:$GO_VERSION
-    # Enable backports repo for debian for swtpm
-    RUN . /etc/os-release && echo "deb http://deb.debian.org/debian $VERSION_CODENAME-backports main contrib non-free" > /etc/apt/sources.list.d/backports.list
-    RUN apt update
-    RUN apt install -y qemu-system-x86 qemu-utils git swtpm && apt clean
-    SAVE IMAGE --push $IMAGE_REPOSITORY_ORG/golang-testing:${GO_VERSION}
```
jimmykarily, line 120: Please remind me where was this used?

tests/bundles_test.go
```diff
@@ -3,27 +3,40 @@ package mos_test
...
+			serial, _ := os.ReadFile(filepath.Join(vm.StateDir, "serial.log"))
+			_ = os.MkdirAll("logs", os.ModePerm|os.ModeDir)
+			_ = os.WriteFile(filepath.Join("logs", "serial.log"), serial, os.ModePerm)
+			fmt.Println(string(serial))
+		}
+
+		err := vm.Destroy(nil)
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(os.Unsetenv("DATASOURCE")).ToNot(HaveOccurred())
```
jimmykarily, line 38: I think this is better: https://onsi.github.io/ginkgo/#cleaning-up-our-cleanup-code-defercleanup

tests/keys/db.auth
jimmykarily, line 1: Why did you delete these man? These are my production keys :p

jimmykarily (commented):
Amazing job @Itxaka !

tests/bundles_test.go
Itxaka: oh nice

Earthfile
Itxaka: it did create the same image that it itself consumed in a different target no? @mauromorales ?

.github/workflows/reusable-uki-test.yaml
Itxaka: `if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && inputs.flavor == 'ubuntu' }}`

This was not changed from the job, I think as a request from someone (Ben?) we started to push the ubuntu uki master images so they were more recent and he could consume them directly from there. I do agree that master images should NOT be pushed ever but....it is what it is.

.github/workflows/reusable-release.yaml
```diff
@@ -0,0 +1,158 @@
...
+          build-args: |
+            BASE_IMAGE=${{ inputs.base_image }}
+            MODEL=${{ inputs.model }}
+            VARIANT=${{ inputs.variant }}
+            KUBERNETES_VERSION=${{ inputs.kubernetes_version }}
+            TRUSTED_BOOT=${{ inputs.trusted_boot }}
+      - name: Grype table
+        uses: anchore/scan-action@v6.0.0
+        with:
+          image: ${{ env.IMAGE_TAG }}
```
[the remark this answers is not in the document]
Itxaka: well it should use the local image always, the idea here is to build the image and load it into the local daemon, do stuff on it locally (security scan, build iso from it, build raw, etc...) then once all that has passed (so no security issues, builds consumable artifacts properly) we build it a second time (for free due to everything being cached) and push it to the final repo. I.e. what we push has passed whatever tests and such properly.

.github/workflows/reusable-release.yaml
Itxaka: no, ideally we would use it although I did not see how to plug it in here properly so I went with the brute force approach of manually constructing it.

Ideally the versioneer should plug it in here easily, I just didnt have that in mind when dropping stuff xD

.github/workflows/image-master.yaml
```diff
@@ -52,329 +28,101 @@ jobs:
...
+    name: ${{ matrix.base_image }}
     secrets: inherit
-    uses: ./.github/workflows/reusable-install-test.yaml
+    uses: ./.github/workflows/reusable-qemu-test.yaml
     with:
-      flavor: ${{ matrix.flavor }}
-      flavor_release: ${{ matrix.flavorRelease }}
+      base_image: ${{ matrix.base_image }}
+      test: ${{ matrix.test }}
+      secureboot: ${{ matrix.secureboot || false}}
```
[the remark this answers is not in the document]
Itxaka: ummm, not in this case, as if your entry doesnt pass a secureboot, like in the include ones this fails completely as it passes an empty value, which is not a valid bool!

I found that the hard way xDDD

tests/bundles_test.go
Itxaka: I think Ill do this one in the follow up PRs as now this one is green and dont want to repush all the jobs

.github/workflows/reusable-uki-test.yaml
jimmykarily: ok let's revisit this after merging.

.github/workflows/reusable-release.yaml
jimmykarily: Even if we use the same tool (versioneer), calling it twice could lead to inconsistencies so the best option imo is to just get the name from `/etc/kairos-release` in the generated artifact. Let's do that after merging this PR.

.github/workflows/image-master.yaml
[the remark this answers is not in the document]
jimmykarily: oops! So an empty value acts like a boolean but not in every case :D , nice one.

jimmykarily (approved):

PR #3174, opened 2025-02-07
:robot: Fix master builds of arm64 artifacts

Itxaka:
Bump kairos-init to fix the nvidia build
On artifact push, use the same as the normal docker build but with push Use the proper Dockerfile
Pass the correct build args for the Kubernetes version Use the main dir as context

Fixes #

### DIFF
--- a/.github/workflows/reusable-docker-arm-build.yaml
+++ b/.github/workflows/reusable-docker-arm-build.yaml
@@ -97,7 +97,7 @@ jobs:
       - name: Build the docker image
         uses: docker/build-push-action@v6
         with:
-          context: images/
+          context: .
           file: images/Dockerfile
           platforms: linux/arm64
           load: true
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-docker-arm-build.yaml
```diff
@@ -142,20 +139,16 @@ jobs:
...
         with:
-          context: images/
-          file: images/Dockerfile.kairos-${{ inputs.family }}
+          context: .
+          file: images/Dockerfile
           platforms: linux/arm64
-          push: "true"
-          tags: ${{ env.IMAGE_REF }}
+          push: true
+          tags: ${{ env.IMAGE_REF }}, kairos:${{ env.GIT_VERSION }}
```
jimmykarily, line 146: the second tag here will be pushed multiple times no? I mean from every job that uses this same pipeline on the same commit, no matter what flavor it's building. Do we want that?

jimmykarily (approved):

.github/workflows/reusable-docker-arm-build.yaml
Itxaka: yep yep, I dropped already, good catch!

PR #3180, opened 2025-02-11
Add mirror to buildx in self hosted

Itxaka:

Fixes #


.github/workflows/release-arm.yaml
```diff
@@ -86,10 +86,21 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Set up Docker Buildx
-        id: buildx
+      - name: Set up Docker Buildx for public runners
+        if: runner.environment == 'github-hosted'
         uses: docker/setup-buildx-action@master
+      - name: Set up Docker Buildx for custom runners
+        if: runner.environment == 'self-hosted'
```
jimmykarily, line 93: is this is set according to the `runs-on` value, we sometimes use `kvm` as a value to point to our runners. Will it work then?

Itxaka: yep! It just differs between self-hosted (no matter the type, label, os, whatever) and github provided(same, no matter the arch, size, or whatever)

PR #3188, opened 2025-02-12
Move temp images to quay

Itxaka:
With a expires by, it will expire after 6 hours and its more stable than ttl

Fixes #

### DIFF
--- a/.github/workflows/reusable-build-flavor.yaml
+++ b/.github/workflows/reusable-build-flavor.yaml
@@ -66,6 +66,7 @@ jobs:
           platforms: linux/${{ inputs.arch }}
           load: true
           tags: ${{ env.IMAGE_NAME }}
+          # https://docs.redhat.com/en/documentation/red_hat_quay/3.4/html/use_red_hat_quay/working_with_tags#tag-expiration
           labels: |
             quay.expires-after=6h
           build-args: |
--- a/.github/workflows/reusable-build-flavor.yaml
+++ b/.github/workflows/reusable-build-flavor.yaml
@@ -49,9 +49,9 @@ jobs:
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@master
-      - name: Set Version
+      - name: Set Image tag
         run: |
-          echo "GIT_VERSION=$(git describe --always --tags --dirty)" >> $GITHUB_ENV
+          echo "IMAGE_NAME=quay.io/ci-temp-images:${{ steps.split.outputs.flavor }}-${{ steps.split.outputs.flavor_release }}-${{ inputs.variant }}-${{ inputs.arch }}-${{ github.sha }}" >> $GITHUB_ENV
       - name: Build container 🔧
         uses: docker/build-push-action@v6
         with:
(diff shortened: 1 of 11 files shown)


.github/workflows/reusable-build-flavor.yaml
```diff
@@ -49,17 +49,25 @@ jobs:
...
         uses: docker/build-push-action@v6
         with:
           context: .
           file: images/Dockerfile
           platforms: linux/${{ inputs.arch }}
           load: true
-          tags: ${{ steps.split.outputs.flavor }}-init:${{ steps.split.outputs.flavor_release }}
+          tags: ${{ env.IMAGE_NAME }}
+          labels: |
+            quay.expires-after=6h
```
jimmykarily, line 70:
```suggestion
            # https://docs.redhat.com/en/documentation/red_hat_quay/3.4/html/use_red_hat_quay/working_with_tags#tag-expiration
            quay.expires-after=6h
```

.github/workflows/reusable-uki-test.yaml
```diff
@@ -107,11 +118,12 @@ jobs:
             --output-dir /result --keys /keys --output-type container \
             dir:/unpacked
 
-          docker load -i build/*.tar
           image=$(docker load -i build/*.tar | grep "Loaded image" | awk -F: '{ st = index($0,":");print substr($0,st+1)}' | xargs)
           rand=$(head /dev/urandom | tr -dc a-z0-9 | head -c 7)
-          temp_image=ttl.sh/$(echo $image | sed "s/:latest/-${rand}:6h/")
-          docker tag $image $temp_image
+          temp_image=quay.io/kairos/ci-temp-images:$(echo $image | sed "s/:latest/-${rand}/")
```
jimmykarily, line 123: This might break if tags are more restrictive that image names in regards to allowed characters (I don't know, I'm just saying). We'll see.

tests/assets/bundles.yaml
```diff
@@ -22,4 +22,4 @@ install:
   bundles:
     - rootfs_path: /var/lib/extensions/kubo
       targets:
-        - container://ttl.sh/bundles-test:24h
\ No newline at end of file
+        - container://quay.io/kairos/ci-temp-images:bundles-test
```
jimmykarily, line 25: Any job might have built this image (since it always has the same name). So if multiple are run in parallel, they will be overwriting this same image. The image will be the same no matter where it's built but this might not be the best idea. It was like that already so no need to fix it here.

tests/bundles_test.go
```diff
@@ -101,7 +101,7 @@ var _ = Describe("kairos bundles test", Label("bundles"), func() {
 				out, _ := vm.Sudo("cat /oem/90_custom.yaml")
 				// https://pkg.go.dev/regexp/syntax
 				// ?s -> "let . match \n (default false)"
-				Expect(out).ToNot(MatchRegexp("(?s)ttl.sh.*ttl.sh"))
+				Expect(out).ToNot(MatchRegexp("(?s)quay.io/kairos/ci-temp-images.*quay.io/kairos/ci-temp-images"))
```
jimmykarily, line 104: I think we need an additional check that the entry is there once. This is easy to become false-positive (e.g. if we add `dimitris` to the regexp, it will always be green. It wasn't better before, so feel free to fix or postpone for other PR.

jimmykarily (approved):

tests/assets/bundles.yaml
Itxaka: yep indeed, when I checked this I found it weird. No idea how to fix this. It helps that the image and contents seem to be always the same and havent changed in years lol,  but what should we do? a fixed image? something that doesnt expire? I have no workaround for this :(

.github/workflows/reusable-build-flavor.yaml
Itxaka: done

tests/bundles_test.go
Itxaka: done

PR #3189, opened 2025-02-12
feat(Dockerfile): allow to override kairos-init via build arg

mudler:

Fixes #

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,6 +1,7 @@
 ARG BASE_IMAGE=ubuntu:20.04
+ARG KAIROS_INIT=v0.2.3
 
-FROM quay.io/kairos/kairos-init:v0.2.3 AS kairos-init
+FROM quay.io/kairos/kairos-init:${KAIROS_INIT} AS kairos-init
 
 FROM ${BASE_IMAGE} AS base
 ARG VARIANT=core


images/Dockerfile
```diff
@@ -1,6 +1,7 @@
 ARG BASE_IMAGE=ubuntu:20.04
+ARG KAIROS_INIT=v0.2.3
 
-FROM quay.io/kairos/kairos-init:v0.2.3 AS kairos-init
+FROM quay.io/kairos/kairos-init:${KAIROS_INIT} AS kairos-init
```
jimmykarily, line 4: Maybe it's better if we allow overriding the whole image? E.g. if we build one locally and push it to ttl.sh or something.

jimmykarily (approved):

images/Dockerfile
Itxaka: will this work for renovate? Do we really need to do this on our Dockerfile? Do we expect to run it locally with a different init? 

Just wondering why this would be wanted instead of a straightforward dockerfile. Doing local stuff is adding 3 lines so its not like its a very complex thing to create your own locally

Itxaka (changes requested):
renovate stanza probably needed so it bumps it properly automatically

images/Dockerfile
Itxaka: I rather indeed have the full address in kairos init var so renovate is able to bump it without having to start doing stupid regexes and custom managers again lol

jimmykarily (approved):

Issue #3157, opened 2025-01-29
feat: Add k0s artifacts to the pipeline


mauromorales:
Let's start by having 3 flavors to release, make sure to do some manual testing on each of them

- Ubuntu 24.04
- Opensuse Leap
- Rocky Linux

mauromorales:
@Itxaka nope, I haven't tested it yet, I'll try to give it a spin once everything is set on the systemd side. If it's working out of the box then I'll just add it here too, if it's not then I'll add a card to debug, fix and add

PR #3193, opened 2025-02-14
Add k0s artifacts

mauromorales:
fixes https://github.com/kairos-io/kairos/issues/3157

### DIFF
--- a/.github/workflows/image-pr.yaml
+++ b/.github/workflows/image-pr.yaml
@@ -41,10 +41,6 @@ jobs:
         base_image:
           - "ubuntu:24.04"
           - "alpine:3.19"
-        include:
-          - base_image: "ubuntu:24.04"
-            kubernetes_distro: "k0s"
-            variant: "standard"
   tests:
     name: ${{ matrix.base_image }}
     secrets: inherit
--- a/.github/workflows/image-pr.yaml
+++ b/.github/workflows/image-pr.yaml
@@ -33,7 +33,6 @@ jobs:
       model: "generic"
       arch: "amd64"
       event_type: ${{ github.event_name }}
-      kubernetes_distro: ${{ matrix.kubernetes_distro }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/image-pr.yaml
+++ b/.github/workflows/image-pr.yaml
@@ -33,6 +33,7 @@ jobs:
       model: "generic"
       arch: "amd64"
       event_type: ${{ github.event_name }}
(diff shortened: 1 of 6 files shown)


images/Dockerfile
```diff
@@ -6,11 +6,11 @@ FROM ${BASE_IMAGE} AS base
 ARG VARIANT=core
 ARG MODEL=generic
 ARG TRUSTED_BOOT=false
-ARG KUBERNETES_PROVIDER=k3s
+ARG KUBERNETES_DISTRO=k3s
```
mauromorales, line 9: I think DISTRO is a better name because providers we have the kairos provider which can handle multiple distros, there's also a k3s provider which happens to also have the k3s kubernetes distro but they are different things

jimmykarily (approved):
no better way to see if it works than to run it :D

.github/workflows/release.yaml
```diff
@@ -149,3 +171,40 @@ jobs:
...
+      statuses: read
+    strategy:
+      fail-fast: false
+      matrix:
+        kubernetes_version: ${{ fromJson(needs.get-k0s-versions.outputs.kubernetes_versions) }}
+        model: ["generic"]
+        variant: ["standard"]
+        arch: ["amd64"]
+        base_image:
+          - "opensuse/leap:15.6"
```
Itxaka, line 202: do we need those 3 systemd builds? why not opensuse and alpine for example?

Itxaka (approved):

.github/workflows/release.yaml
Itxaka: oh this is release, sorry, I though it was the PR one

.github/workflows/image-pr.yaml
```diff
@@ -33,6 +33,7 @@ jobs:
       model: "generic"
       arch: "amd64"
       event_type: ${{ github.event_name }}
+      kubernetes_distro: ${{ matrix.kubernetes_distro }}
```
Itxaka, line 36: is this needed?

Itxaka:
notice that we dont push to a fixed place anymore to waste time. Temp images can be found at https://quay.io/repository/kairos/ci-temp-images?tab=tags&tag=latest but they expire after 6 hours. Master builds are no longer pushed to a fixed place as well, no point into that when its that easy to build your own one :D

Itxaka:
> @jimmykarily i did run on the pr once just to be sure at least that one produces a working image ... but yeah master and release are always a gamble :P

mmmmh I take issue with that :sob:

PR #3199, opened 2025-02-17
Push master images to quay

mauromorales:
Just for easy access until we have the factory webui

### DIFF
--- a/.github/workflows/reusable-build-flavor.yaml
+++ b/.github/workflows/reusable-build-flavor.yaml
@@ -135,3 +135,16 @@ jobs:
         if: ${{ inputs.model == 'generic' }}
         run: |
           docker push ${{ env.IMAGE_NAME }}
+      - name: Tag and Push Nightly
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-build-flavor.yaml
```diff
@@ -135,3 +135,16 @@ jobs:
         if: ${{ inputs.model == 'generic' }}
         run: |
           docker push ${{ env.IMAGE_NAME }}
+      - name: Tag and Push Nightly
+        if: github.ref == 'refs/heads/master'
```
Itxaka, line 139: do we need to check if the action is a push or just nothing else would have that ref?

Itxaka (approved):
nice and easy :D

.github/workflows/reusable-build-flavor.yaml
mauromorales: good question, I think this is enough but not 100% ... maybe i test and see what happens

PR #3214, opened 2025-02-21
Bump and adapt to new kairos-init

Itxaka:
- fixes some issues with alpine
- brings missing logrotate
- installs expected utils in standard
- now requires a version to be passes to set the output version of the artifacts

Fixes #


images/Dockerfile
```diff
@@ -1,30 +1,18 @@
...
-RUN /kairos-init -f "${FRAMEWORK_VERSION}" -l debug -s init -m "${MODEL}" -v "${VARIANT}" -t "${TRUSTED_BOOT}" -k "${KUBERNETES_DISTRO}" --k8sversion "${KUBERNETES_VERSION}"
-RUN /kairos-init -f "${FRAMEWORK_VERSION}" -l debug --validate -m "${MODEL}" -v "${VARIANT}" -t "${TRUSTED_BOOT}" -k "${KUBERNETES_DISTRO}" --k8sversion "${KUBERNETES_VERSION}"
-RUN rm /kairos-init
-
-FROM base-kairos AS kairos-core
-
-FROM base-kairos AS kairos-standard
-COPY --from=framework /usr/bin/luet /usr/bin/luet
-COPY --from=framework /etc/luet/luet.yaml /etc/luet/luet.yaml
-RUN luet install -y utils/edgevpn utils/k9s utils/nerdctl container/kubectl utils/kube-vip
```
jimmykarily, line 28: is this happening on kairos-init now?

tests/upgrade_cli_test.go
```diff
@@ -71,12 +70,6 @@ var _ = Describe("k3s upgrade manual test", Label("upgrade-with-cli"), func() {
 				return err
 			}, 10*time.Minute, 10*time.Second).ShouldNot(HaveOccurred())
 
-			var v string
-			Eventually(func() string {
-				v, _ = vm.Sudo(getVersionCmd)
-				return v
-				// TODO: Add regex semver check here
-			}, 10*time.Minute, 10*time.Second).Should(ContainSubstring("v"), v)
```
jimmykarily, line 79: This check seemed intentional. What changed?

jimmykarily:
`fixes some issues with alpine` -> what issue was that?

tests/upgrade_cli_test.go
Itxaka: freeform semver for now.

Still this check is useless. Checking that the version just contains a v its absurd? Even if the upgrade didnt pass, this check would pass so I dont see how this is a useful check at all.

images/Dockerfile
Itxaka: https://github.com/kairos-io/kairos-init/commit/29ac33a8d8425378f679f05095fa5f7336a54261

mauromorales (approved):

tests/upgrade_cli_test.go
jimmykarily: I think we added that because we had a bug with a missing `v` . This covers it enough :D.

Itxaka:
> `fixes some issues with alpine` -> what issue was that?

It's actually an issue with future alpine versions, as they removed /var/lock and linked it to /run/lock which breaks the locking mechanism of yip on user creation

PR #3265, opened 2025-03-17
Disable fips flag for ubnutu example

Itxaka:
We dont support automatic fips for ubuntu so we should not pass that flag and show in the example how to build a ubutu fips image

Fixes #

### DIFF
--- a/examples/builds/ubuntu-fips/Dockerfile
+++ b/examples/builds/ubuntu-fips/Dockerfile
@@ -4,7 +4,7 @@ FROM ubuntu:20.04
 ARG VERSION=v0.0.1
 
 COPY --from=kairos-init /kairos-init /kairos-init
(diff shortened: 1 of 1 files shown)


jimmykarily:
That Dockerfile is under a directory called `ubuntu-fips`. If fips doesn't work then the name is wrong or this example shouldn't exist.

Itxaka:
> That Dockerfile is under a directory called `ubuntu-fips`. If fips doesn't work then the name is wrong or this example shouldn't exist.

The example generates a working fips image, but as it needs user modification (basically the ubuntu pro token) we put the example on how to do it, there is even a README in there, but unfortunately kairos-init cant do that for you.

But we still want to provide users with a example on how to build their own ubuntu fips :)

mauromorales (approved):

jimmykarily (approved):

PR #3287, opened 2025-03-21
Bump init to v0.4.2 and framework to v2.18.1

mauromorales:


Itxaka (approved):

Itxaka:
reset + encryption + uki failing (due to reset)


reset seems to be failing because on copylogs we try to mount the partions and if they are encrypted we cant find them.
on encryption tests, the same thing is observed. Looks like we should wrap the copylogs and bundleinstall on non uki with a unlock and then lock so tey can be mounted correctly

Some issues might also come from things trying to mount the disk directly by the label and getting the encrypted one instead of the unlocked as now both share the same label so it mistakenly returns the locked partition

CopyLogs was never meant to be 100% working and make the install fail as it was provided in a best effort basis, as user scripts can leave things around. we could try to make it work 100% of the time but its a bit complex.

At tleast the reset for the backport should be a simple disregard of the error coming from copylogs

Issue #2068, opened 2023-12-08
:robot: Make decentralized tests work on CI


mauromorales:
Decentralized tests were commented on https://github.com/kairos-io/kairos/pull/2067#discussion_r1420280932_ because they haven't been green since months. This ticket is about figuring out what makes them fail on CI and fixing it. Feel free to create sub tickets if necessary.

Itxaka:
decentralized tests are enabled in master now

PR #3288, opened 2025-03-21
[WIP] Add decentralized test in pr

mauromorales:
just want to test if this works with the new setup without having to do any debugging

relates to https://github.com/kairos-io/kairos/issues/2068

### DIFF
--- a/.github/workflows/image-pr.yaml
+++ b/.github/workflows/image-pr.yaml
@@ -40,6 +40,40 @@ jobs:
         base_image:
           - "ubuntu:24.04"
           - "alpine:3.21"
+  build-standard:
+    name: ${{ matrix.base_image }}
+    uses: ./.github/workflows/reusable-build-flavor.yaml
+    secrets: inherit
+    permissions:
+      contents: write
+      security-events: write
+      id-token: write
+      actions: read
+      attestations: read
+      checks: read
+      deployments: read
+      discussions: read
+      issues: read
+      packages: read
+      pages: read
+      pull-requests: read
+      repository-projects: read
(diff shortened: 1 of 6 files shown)


Itxaka:
FYI what I do for this is to have 1 commit called DROP in which I drop all the other tests so they dont have to run and can only run the single test Im interested on, so we dont use resources and can iterte faster on it


Also if this is the edgvpn test in which 2 nodes are trying to communicate and both are in a qemu user network (slirp, 10.0.2.X network) this stopped working due to some changes upstream in the dht libs I believe.

Lat working version of edgebpn was 0.25 or something like that!

Itxaka:
> @Itxaka oh cool, I need to try that out
> 
> As for the test, thing is, it works if I run it locally so it's the CI setup IMO, maybe they block some ports/networks hence it doesn't get to have everything needed to do the dht properly. So what I'm thinking to try is to run them without dht but being CI I'm not sure how to put these vms in the same network

I think peg just puts them into user networking: https://github.com/spectrocloud/peg/blob/master/pkg/machine/qemu.go#L86

There was a way of seting them into a bridge networking but it was a bit of a mess as you needed to install the qemu helpers, set up a bridge then use that bridge to attach the VMS to it, would be easier to test it manually locally lol.

Or try to understand why now it doesnt work when it worked with edgevpn 0.25. IIRC there were some extra fields added to the config and edgevpn was ignoring those, but we could not make that work at all when we tried :(

Check with @mudler he may remember it

mauromorales:
closing for now will try to pick it up later

PR #3301, opened 2025-03-26
Push cloud images when a new tag is created

jimmykarily:
because if we have released 3.4.0-beta1 and we then push 3.3.3 that release will never be pushed because the scheduled job only picks the latest tag from master.

Fixes #


Itxaka (approved):

Itxaka:
Wait but if we run 7 betas then it will try to upload those?

mauromorales (approved):

PR #3341, opened 2025-04-17
Re-enable decentralized test

jimmykarily:

Fixes #

### DIFF
--- a/tests/provider_decentralized_test.go
+++ b/tests/provider_decentralized_test.go
@@ -193,13 +193,12 @@ var _ = Describe("kairos decentralized k8s test", Label("provider", "provider-de
 					out, err = vm.Sudo(`curl -X POST http://localhost:8080/api/dns --header "Content-Type: application/json" -d '{ "Regex": "foo.bar", "Records": { "A": "2.2.2.2" } }'`)
 					Expect(err).ToNot(HaveOccurred(), out)
 
-					out, _ = vm.Sudo("dig +short foo.bar")
+					out, _ = vm.Sudo("ping -c 1 foo.bar")
 					return strings.TrimSpace(out)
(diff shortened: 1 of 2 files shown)


tests/provider_decentralized_test.go
```diff
@@ -193,21 +193,20 @@ var _ = Describe("kairos decentralized k8s test", Label("provider", "provider-de
 					out, err = vm.Sudo(`curl -X POST http://localhost:8080/api/dns --header "Content-Type: application/json" -d '{ "Regex": "foo.bar", "Records": { "A": "2.2.2.2" } }'`)
 					Expect(err).ToNot(HaveOccurred(), out)
 
-					out, _ = vm.Sudo("dig +short foo.bar")
```
jimmykarily, line 196: there is no `dig` available in the image

tests/provider_decentralized_test.go
```diff
@@ -193,21 +193,20 @@ var _ = Describe("kairos decentralized k8s test", Label("provider", "provider-de
...
+					return out
+				})
 			}
 		})
 
 		vmForEach("checking if it upgrades to a specific version", vms, func(vm VM) {
 			version, err := vm.Sudo(getVersionCmd)
 			Expect(err).ToNot(HaveOccurred(), version)
 
-			out, err := vm.Sudo("kairos-agent upgrade --source oci:quay.io/kairos/opensuse:leap-15.6-standard-amd64-generic-v3.2.3-k3sv1.31.2-k3s1")
```
jimmykarily, line 210: if the original OS is ubuntu, with `/etc/init.d` being mounted on persistent, it means the upgraded opensuse OS will try to start ubuntu specific services and it fails. This results in sshd not starting at all.

PR #3345, opened 2025-04-21
Fix the nvidia orin build on master

Itxaka:
Fixed the nvidia orin.

As init doesnt have the full set of packages for orin, we still need to build the base image as a separate thing, then use that as base for init to built the kairos image, which means that the base image does not confoft to what we are expecting as input, which breaks our distro/version separation.

We fix this by fixing the distro and tag manually in the nvidia build  case

### DIFF
--- a/.github/workflows/reusable-build-flavor.yaml
+++ b/.github/workflows/reusable-build-flavor.yaml
@@ -57,6 +57,14 @@ jobs:
             # Use the base image as the distro
             distro="$base_image"
           fi
+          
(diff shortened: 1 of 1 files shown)


ci-robbot (commented):
The changes look good and are focused on fixing the build for the NVIDIA Orin platform. However, I will wait for all CI checks to complete before approving the PR. Please ensure that all checks pass.

ci-robbot (approved):
The changes look good and only affect the master branch jobs. Merging this to see the ongoing changes.

PR #3348, opened 2025-04-21
Fix decentralized test

Itxaka:
- Run it under the github provided nodes as it seems that our workers network is not good enough
- Also rework the config so it enables k3s and sets a proper hostname rather than a random one, as that was causing on reboot the k3s nodes to have a different names and rebuild themselves, which is slower
- Drop the upgrade part as that its covered in a diferent test,here we are checking for decentralized communication and k3s setup

Fixes #


ci-robbot (commented):
I will provide a detailed review shortly.

PR #3430, opened 2025-05-26
feat(ci): add stale bot

mudler:
This PR adds stale bot

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,19 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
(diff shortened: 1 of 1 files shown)


mudler:
> it's a bummer to see this as I believe stalebot makes open source software worse. the activity on an Issue is not directly correlated with its value or importance, nor does stale bot make pleasant ideas which the project is not yet ready for but would be transformative in due time. I would make an Issue about it to reassess how well it's doing next May, but stalebot would close it by then.

We have discussed this in the planning and we decided to have the stale bot only for issues that are not confirmed by the team. 

While not ideal, it allows us to focus on the issues that are more relevant and, helps also to identify issues that are worth discussing again: the stale bot won't close issues immediately, but it will notify about stale issues before closing.

It will close them only if there is not even a comment on it after 5 days the notification was issued: https://github.com/kairos-io/kairos/blob/e28bb58c24c46dd7fc8f9f6c951aa41990edada8/.github/workflows/stale.yaml#L21

tyzbit:
Thank you for the context. I still think a better solution exists such as by using tags and filters instead of outright closing issues automatically, but this incarnation does take some steps to be less thoughtless and rigid.

PR #3452, opened 2025-05-31
Improve Alpine WiFi example

jimmyjones2:

Remove need for manually generating WiFI config on a working system and then copying into cloud-config. Now you can just add your WiFi network details and it will connect on first boot


Itxaka (approved):

Itxaka:
Thanks @jimmyjones2 !

PR #3523, opened 2025-07-10
Drop keylime example

Itxaka:
Moved it do docs instead


examples/keylime/README.md
```diff
@@ -3,16 +3,13 @@
 Most of the steps are already covered in the [Keylime documentation](https://keylime-docs.readthedocs.io/en/latest/). Here we will cover the steps that are specific to Kairos.
```
mudler, line 3:
```suggestion
This example how to use Keylime with Kairos in order to provide verified measurement for runtime binaries or files in a Kairos system. This, for instance involve having measurements for specific files in the persistent portion of the disk, or the configuration directories.

Most of the steps are already covered in the [Keylime documentation](https://keylime-docs.readthedocs.io/en/latest/). Here we will cover the steps that are specific to Kairos.

## Extend Kairos

First of all we need to create a Kairos derivative with the keylime agent, in order to do this we use the [Kairos factory process](https://kairos.io/docs/reference/kairos-factory/) where we will build our own OS as a container image and we will install the kairos-agent:
```

mudler (approved):
Really nice - to me it looks clear. I didn't tested it, but the flow and the example looks OK. Just a small nit at  the start of the example where I'd give some hints on the docs.

mudler:
@Itxaka I see we also have https://kairos.io/docs/examples/keylime/ - should we also update that part of the docs?

Itxaka:
> @Itxaka I see we also have https://kairos.io/docs/examples/keylime/ - should we also update that part of the docs or is it updated automatically?

That was just a pure copy of the previous readme here, so it needs either updating or maybe to simplify it with a remote include so it's always updated?

examples/keylime/README.md
```diff
@@ -1,18 +1,21 @@
...
 
+This example how to use Keylime with Kairos in order to provide verified measurement for runtime binaries or files in a Kairos system. This, for instance involve having measurements for specific files in the persistent portion of the disk, or the configuration directories.
+
 Most of the steps are already covered in the [Keylime documentation](https://keylime-docs.readthedocs.io/en/latest/). Here we will cover the steps that are specific to Kairos.
 
+## Extend Kairos
 
-We provide the keylime agent as a luet package for ease of installation.
-For it to be installeed you need to create your own derivative of the Kairos image and add the keylime-agent package to it.
+First of all we need to create a Kairos derivative with the keylime agent, in order to do this we use the [Kairos factory process](https://kairos.io/docs/reference/kairos-factory/) where we will build our own OS as a container image and we will install the kairos-agent:
```
Itxaka, line 9:
```suggestion
First of all we need to create a Kairos derivative with the keylime agent, in order to do this we use the [Kairos factory process](https://kairos.io/docs/reference/kairos-factory/) where we will build our own OS derivative to bundle keylime-agent:
```

examples/keylime/README.md
```diff
@@ -1,18 +1,21 @@
...
 
 
 ```Dockerfile
-FROM quay.io/kairos/ubuntu:24.04-core-amd64-generic-v3.2.1 AS base
-COPY luet.yaml /etc/luet/luet.yaml
-RUN luet install -y --relax utils/keylime-agent
+FROM quay.io/kairos/rockylinux:9-core-amd64-generic-v3.4.1 AS base
+RUN dnf -y update && dnf -y install --setopt=install_weak_deps=False keylime && dnf clean all
```
Itxaka, line 14:
```suggestion

# Build the keylime agent
FROM ubuntu:24.04 AS keylime-build
RUN apt-get update && apt-get install -y --no-install-recommends git gcc libclang-dev libssl-dev libtss2-dev libzmq3-dev pkg-config rustup make && rm -rf /var/lib/apt/lists/*
WORKDIR /keylime
RUN git clone --depth 1 --branch v0.2.7 https://github.com/keylime/rust-keylime.git /keylime
ENV DESTDIR=/keylime-output
RUN rustup default stable && make && make install

# Build the final image with the keylime agent copied over, only systemd supported for now
FROM quay.io/kairos/rockylinux:9-core-amd64-generic-v3.4.1 AS default
COPY --from=keylime-build /keylime-output/ /
```

Itxaka (commented):
dsf

Itxaka:
> > > @Itxaka I see we also have [kairos.io/docs/examples/keylime](https://kairos.io/docs/examples/keylime/) - should we also update that part of the docs or is it updated automatically?
> > 
> > 
> > That was just a pure copy of the previous readme here, so it needs either updating or maybe to simplify it with a remote include so it's always updated?
> 
> maybe we can just move the examples directly in the docs and we avoid having it in both parts? we could just link from here to the example section in the docs to avoid 404s

The idea on having the examples here was so on change (bump stuff for example) we could auto test this directly, but in this case as its a derivative, it makes no sense and we can instead have it on the docs, sounds good to me then.

PR #3587, opened 2025-08-12
Use kairos factory action for releases

mauromorales:
the only thing that I didn't test is the actual release artifacts, I say we can do a rc for the patch release test it and do any fixes for the final patch release


.github/workflows/release-arm.yaml
```diff
@@ -7,19 +7,26 @@ on:
...
     permissions:
       id-token: write  # OIDC support
       contents: write
-      actions: read
       security-events: write
+      actions: read
+      attestations: read
+      checks: read
+      deployments: read
+      discussions: read
```
jimmykarily, line 20: strange permissions, aren't these?

.github/workflows/reusable-uki-test.yaml
```diff
@@ -1,203 +0,0 @@
...
-          FIRMWARE: /usr/share/OVMF/OVMF_CODE.fd
-          EMULATE_TPM: true
-          UKI_TEST: true
-        run: |
-          export ISO=$(ls $PWD/build/kairos-*-uki.iso)
-          export UPGRADE_IMAGE=$(cat "${PWD}/build/upgrade_image_name")
-          export EXPECTED_NEW_VERSION="mynewversion"
-          export EXPECTED_SINGLE_ENTRY="testentry"
-          cp tests/go.* .
-          go run github.com/onsi/ginkgo/v2/ginkgo -v --label-filter "${{ inputs.test_label }}" --fail-fast -r ./tests/
```
jimmykarily, line 197: so we don't run these anymore?

mauromorales: @jimmykarily we do, here: https://github.com/kairos-io/kairos/blob/master/.github/workflows/uki.yaml#L158

mauromorales: it's already on master

.github/workflows/release-arm.yaml
mauromorales: LOL yeah the discussions is weird, maybe it was some cursor madness, let me try removing them

mauromorales: yup there's a read all at the top so they can all go away

Issue #3756, opened 2025-11-12
arm64 artifacts dropped on quay.io from 3.5.x release


Ludea:
**Kairos version:**

3.5.x

**CPU architecture, OS, and Version:**

arm64 hosts

**Describe the bug**

all arm64 hosts has been dropped into your cd pipelines, they are not uploaded to quay.io

**To Reproduce**

**Expected behavior**

Available arm64 artifacts on quay.io
**Logs**

**Additional context**

PR #3771, opened 2025-11-17
Set registry_domain so that we push the produced images

jimmykarily:
That's the condition used here:

https://github.com/kairos-io/kairos-factory-action/blob/a9439a91ef3c968c99a6d50209fa8e2db22afb85/.github/workflows/reusable-factory.yaml#L609C1-L609C41

Fixes https://github.com/kairos-io/kairos/issues/3756


Itxaka (approved):
Good catch

PR #3908, opened 2026-02-02
Use Cleanup

mauromorales:
moves the orin release to the core and not standard, just like we have done in every other release. Uses cleanup for nx.

### DIFF
--- a/.github/workflows/image-master-arm.yaml
+++ b/.github/workflows/image-master-arm.yaml
@@ -5,7 +5,6 @@ on:
     branches:
       - master
       - factory-master-arm
-      - use-cleanup
 
 permissions:
   contents: write


.github/workflows/image-master-arm.yaml
```diff
@@ -36,15 +37,19 @@ jobs:
...
       summary_artifacts: true
+      cleanup: ${{ matrix.cleanup }}
     strategy:
       fail-fast: false
       matrix:
         model: ["generic", "rpi4", "rpi3"]
         base_image:
           - "ubuntu:24.04"
           - "alpine:3.21"
+        cleanup: [false]
```
jimmykarily: what's the default?

PR #3912, opened 2026-02-04
feat: Make Hadron front and center of our testing and release

Itxaka:
- Standardized base image to "ghcr.io/kairos-io/hadron:v0.0.1" across all relevant workflows
- Updated ARM and PR workflows to include new models: "nvidia-jetson-agx-orin", "nvidia-jetson-orin-nx"
- Enhanced cleanup strategy for ARM builds

This PR makes sure we tests EVERYTHING under PRs, as we move to hadron only releases.

Also expands the test coverage to include the nvidia builds under arm.

It also streamlines everything by making it mostly a matrix instead of separated jobs, have a single test reusable workflow and drop the tests from the master builds, instead testing everything in the PR branches

### DIFF
--- a/.github/workflows/image-pr-arm.yaml
+++ b/.github/workflows/image-pr-arm.yaml
@@ -42,7 +42,7 @@ jobs:
       model: ${{ matrix.model }}
       arch: "arm64"
       version: "auto"
-      raw: true
+      raw: ${{ matrix.model == 'rpi3' || matrix.model == 'rpi4' }} # only produce raw images for rpi3 and rpi4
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -188,7 +139,8 @@ jobs:
       list_release_artifacts: true
       cosign: true
       release: true
-  build-standard-k0s:
+  standard-k0s:
```
mauromorales: k0s only gets built on release, maybe it's worth having one version in master

PR #3914, opened 2026-02-05
fix: build proper model for arm on release

Itxaka:
We were building the rpi4 only on the release arm, so things like nvidia were named as a job but the underlying job was always building the rpi4 model.

This fixes it by setting the model to the matrix model

Fixes #


Itxaka:
Force merging as this doesnt affect normal pipelines

PR #4049, opened 2026-04-21
Ignore grype cve until k0s/k3s have a fix

### LINKED ISSUE #4048: No k0s artifacts on latest releases
Because the pipeline failed on CVEs: https://github.com/kairos-io/kairos/actions/runs/23347743311/job/67917692341

### Discussed in https://github.com/kairos-io/kairos/discussions/4047

<div type='discussions-op-text'>

<sup>Originally posted by **SanekP** April 18, 2026</sup>
https://kairos.io/quickstart/
```
Can I use a different Kubernetes distribution with Kairos?

Yes, you can download the standard image with k0s. Both k3s and k0s are equally supported by the Kairos team.

Last updated on Feb 24, 2026 by Mauro Morales
```
I assume https://github.com/kairos-io/kairos/releases this is main channel where to check the releases. So v4.0.3 and v4.0.2 doesn't contain k0s images.</div>

jimmykarily:
also bump kairos-init to bump versions

Should fix: https://github.com/kairos-io/kairos/issues/4048

if we release but if we don't release for a while, the fix might make it in k0s/k3s and we can skip this.

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -22,3 +22,15 @@ ignore:
     package:
       name: stdlib
       location: /usr/bin/edgevpn
+
+  # CVE-2026-27143 is a stdlib vulnerability fixed in go1.25.9/1.26.2.
+  # k3s and k0s have not yet cut releases built with a fixed Go toolchain.
+  # Remove these ignores once k3s and k0s ship binaries built with go1.25.9+ or go1.26.2+.
+  - vulnerability: CVE-2026-27143
+    package:
+      name: stdlib
+      location: /usr/bin/k3s
+  - vulnerability: CVE-2026-27143
+    package:
+      name: stdlib
+      location: /usr/bin/k0s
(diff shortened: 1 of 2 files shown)


Itxaka (changes requested):
We agreed to not release stuff with know CVEs qas we quickly get reports about them and have to cut a patch release to avoid those, so I would not do this. If k0s/k3s has not released a patch, is on them not on us.

If the fix is actually not affecting us or its a low score then I guess its ok, but we should try to set the reason there so its easy to drop it afterwards and we dont keep a huge list of exceptions.

jimmykarily:
`If k0s/k3s has not released a patch, is on them not on us.` -> I agree, that's why when the time comes for us to release, we can't wait for them to cut a fixed version. But we are not releasing yet so I don't think we should merge this (yet). I opened it as an indicator of what's wrong currently.

mauromorales:
oops sorry somehow I missed this existed ...  closing in favor of https://github.com/kairos-io/kairos/pull/4052

Issue #3816, opened 2025-12-06
feat: Provide a CI job that compares sizes


Itxaka:
In a PR after building the images (trusted and non) we should pull the last main images and compare sizes.

True, containers does not provide an accurate size so we could maybe dump the images into a tar file and compare those instead or even after building the iso, build and iso from main and compare the isos as well

This would help us track the sizes over time and identify which sizes are increased or decreased when adding/removing/changing sotfware.

PR #4281, opened 2026-08-06
Feat/ci iso size diff

thev1ndu:

Adds a size-diff job to the PR image workflow that downloads the ISOs built for the PR and the ones from the last successful master build, then reports a per-artifact size delta to the job summary. Pairs the two sides by artifact name so version-suffixed ISO file names still match, and degrades to a PR-only report when no baseline is available.

The comparison logic lives in scripts/iso-size-diff.sh so it can be run and tested outside CI.

This PR covers the per-PR ephemeral signal only. The durable "track sizes over time" part of #3816 (porting the hadron size-history.sh pattern to ISOs) will land in a follow-up PR, which is what actually closes the issue.

Related to #3816


jimmykarily:
@thev1ndu thank you for the contribution.

Reading #3816 again, the ask is to "track the sizes over time and identify which sizes are increased or decreased when adding/removing/changing software". A step summary that only survives the length of one CI run does not really cover that half. Nobody is going to scroll back through old runs to reconstruct a trend.

We already solved the same shape of problem for container images in hadron. The split there is:

1. Per-PR ephemeral report against the :main baseline, close to what this PR does but for images:
https://github.com/kairos-io/hadron/blob/1b6ff27ecea3a9f376859a548934b8ebcbdf07e3/.github/scripts/image-size-report.sh
2. Per-merge durable history, appended to a CSV on a dedicated orphan size-history branch, with a Markdown table and an SVG chart regenerated from it on every merge:
https://github.com/kairos-io/hadron/blob/1b6ff27ecea3a9f376859a548934b8ebcbdf07e3/.github/scripts/size-history.sh

Short README that walks through how the two fit together: https://github.com/kairos-io/hadron/blob/1b6ff27ecea3a9f376859a548934b8ebcbdf07e3/.github/scripts/README.md

Workflow wiring for the merge job (main-only, continue-on-error: true on the push so a history hiccup never fails the build, release tags looked up from the API and annotated onto the merge row):
https://github.com/kairos-io/hadron/blob/1b6ff27ecea3a9f376859a548934b8ebcbdf07e3/.github/workflows/build-multiarch-images.yml#L1332-L1435

And the actual rendered output on hadron, so you can see what falls out of that pipeline: https://github.com/kairos-io/hadron/blob/size-history/SIZE_HISTORY.md

Two ways forward that would both work for me:

1. Land this PR as is for the per-PR signal, and open a follow-up that ports the hadron size-history.sh pattern to ISOs. That follow-up is what actually closes #3816.
2. Extend this PR to also record on merge to master, and close #3816 in one go.

Either is fine. I would just rather not mark #3816 done while the "over time" part is still missing.

thev1ndu:
Thanks @jimmykarily, that makes sense. Let's go with option 1.

I've dropped `Closes #3816` from the description (now `Related to #3816`) so merging this won't prematurely close the issue. This PR stays focused on the per-PR ephemeral report against the last master build.

I'll open a follow-up that ports the hadron `size-history.sh` pattern to ISOs durable CSV history on an orphan `size-history` branch with the regenerated Markdown table + SVG chart, and a main-only merge job wired with `continue-on-error: true` on the push. That follow-up will be the one that closes #3816.

jimmykarily (dismissed):

thev1ndu:
Hi @jimmykarily any update on this?

jimmykarily:
> Hi @jimmykarily any update on this?

@thev1ndu copilot's review contains valid comments. I'm waiting for those to be resolved. Also after the monorepo migration this branch has conflicts with master.

thev1ndu:
> > Hi @jimmykarily any update on this?
> 
> @thev1ndu copilot's review contains valid comments. I'm waiting for those to be resolved. Also after the monorepo migration this branch has conflicts with master.

Hi thanks I'll work on it

thev1ndu:
Pushed an update covering both things you flagged.

**Copilot comments**
- `file_size()` now reads the size from `stat` metadata (with a `wc -c` fallback for anything without `stat`) instead of streaming the whole ISO just to count bytes.
- When there is no baseline at all, the header says so and every row shows `n/a` for the delta rather than labelling everything `_new_`. `_new_` is now reserved for the real case: a baseline exists but does not contain that particular artifact.
- The baseline lookup uses `.databaseId // empty`, so a missing master run leaves `run_id` genuinely empty and we actually take the PR-only path instead of trying to download `null`.

**Conflict with master**
The monorepo migration retired `image-pr.yaml`, so the old size-diff job had nowhere to land. I moved it into the new `pr.yaml`: it now `needs: build-iso`, checks out the PR ref explicitly (since `pull_request_target` defaults to the base branch), and finds the baseline from the `master.yaml` run instead of the old `image-master.yaml`. The `kairos-hadron-*-amd64-generic.iso.zip` artifact naming survived the migration, so the download pattern still matches. GitHub shows the branch mergeable again.

Still going with option 1: this PR stays the per-PR signal, and the durable size-history port is what I will open next to close #3816.

thev1ndu:
@jimmykarily I have updated the PR

jimmykarily (approved):

PR #4282, opened 2026-08-06
:chart_with_upwards_trend: ci: track ISO sizes over time on merge to master

### LINKED ISSUE #3816: feat: Provide a CI job that compares sizes

thev1ndu:

Follow-up to #4281. That PR added the *ephemeral* per-PR ISO size delta; this one adds the *durable* "track sizes over time" half of #3816, ported from the hadron `size-history.sh` pattern.

On every merge to `master`, a new `size-history` job records the size of each built ISO into a CSV on an orphan `size-history` branch and regenerates a Markdown table and an SVG chart (one independently scaled panel per ISO) from it, so trends survive the log / step-summary purging the per-PR report is subject to.

- `scripts/iso-size-history.sh` holds the `record` / `render` / `all` logic so it runs and is testable outside CI.
- The CSV is long-format (one row per merge+artifact), and artifacts are paired by a **version-stripped canonical key**, so a changing set of ISOs and the per-build release suffix in artifact names both pair correctly over time.
- Release tags are looked up from the API and annotated onto their merge row (never a separate row).
- The publish step is `continue-on-error` so a history hiccup never fails the build; pushes rebase-and-retry so close merges don't clobber each other.

Opened as a draft for early review of the approach while #4281 lands.

Closes #3816

### DIFF
--- a/.github/workflows/image-master.yaml
+++ b/.github/workflows/image-master.yaml
@@ -57,3 +57,108 @@ jobs:
             # Report-only: k3s CVEs come from upstream and should not block
             # master builds / tests. Mirrors release.yaml (PR #4052).
             security_scan_mode: "report-only"
+  size-history:
+    # Durable ISO size time series: one data point per merge to master,
+    # appended to a CSV on the orphan `size-history` branch, with a Markdown
+    # table and an SVG chart regenerated from it. This is the "over time" half
+    # of the size tracking; the ephemeral per-PR delta lives in image-pr.yaml.
+    name: iso-size-history
+    if: ${{ github.ref_name == 'master' }}
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write   # push the size-history branch
+      actions: read      # download the built ISO artifacts
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Download master ISOs
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: "kairos-hadron-*-amd64-generic.iso.zip"
+          path: candidate
+      - name: Record ISO sizes and render history
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          REPO='${{ github.repository }}'
+          BRANCH='size-history'
+          DATA_DIR="${RUNNER_TEMP}/size-history"
+          REMOTE="https://x-access-token:${GH_TOKEN}@github.com/${REPO}.git"
+          # Load the existing history branch, or start a fresh orphan one.
+          if git clone --quiet --branch "$BRANCH" --single-branch "$REMOTE" "$DATA_DIR" 2>/dev/null; then
+            echo "Loaded existing '$BRANCH' branch"
+          else
+            echo "No '$BRANCH' branch yet; starting a fresh one"
+            mkdir -p "$DATA_DIR"
+            git -C "$DATA_DIR" init -q -b "$BRANCH"
+          fi
+          # Build a "<merge-sha> <release-tag>" map so the rendered history table
+          # can annotate a merge with its release version (a release shares the
+          # merge's sha, so it never becomes a second row). We dereference
+          # annotated tags to the commit they point at and keep only tags that
+          # correspond to a published release, falling back to all tags if the
+          # releases API is unavailable. A missing map just means no annotations
+          # -- never fail the history on it.
+          TAGS_FILE="${RUNNER_TEMP}/release-tags.txt"
+          : >"$TAGS_FILE"
+          RELEASES="$(gh api "repos/${REPO}/releases" --paginate \
+            --jq '.[].tag_name' 2>/dev/null || true)"
+          git ls-remote --tags "$REMOTE" 2>/dev/null | awk '
+            { sha=$1; ref=$2; d=0
+              if (ref ~ /\^\{\}$/) { d=1; sub(/\^\{\}$/, "", ref) }
+              sub(/^refs\/tags\//, "", ref)
+              if (d) commit[ref]=sha; else if (!(ref in light)) light[ref]=sha
+            }
+            END {
+              for (t in commit) print commit[t], t
+              for (t in light)  if (!(t in commit)) print light[t], t
+            }' >"${RUNNER_TEMP}/all-tags.txt" || true
+          if [ -n "$RELEASES" ]; then
+            awk 'NR==FNR{rel[$0]=1; next} ($2 in rel)' \
+              <(printf '%s\n' "$RELEASES") "${RUNNER_TEMP}/all-tags.txt" \
+              >"$TAGS_FILE" || true
+          else
+            cp "${RUNNER_TEMP}/all-tags.txt" "$TAGS_FILE" 2>/dev/null || true
+          fi
+          scripts/iso-size-history.sh all \
+            "${DATA_DIR}/size-history.csv" '${{ github.sha }}' candidate "$DATA_DIR" \
+            "$REPO" "$TAGS_FILE" \
+            | tee -a "$GITHUB_STEP_SUMMARY"
+      - name: Publish history branch
+        # Never fail the build on a history hiccup; the step summary above is
+        # still written even if the commit/push cannot complete.
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          REPO='${{ github.repository }}'
+          BRANCH='size-history'
+          DATA_DIR="${RUNNER_TEMP}/size-history"
+          REMOTE="https://x-access-token:${GH_TOKEN}@github.com/${REPO}.git"
+          cd "$DATA_DIR"
+          git add size-history.csv SIZE_HISTORY.md size-history.svg
+          if git diff --cached --quiet; then
+            echo "No size changes to commit"
+            exit 0
+          fi
+          git -c user.name='github-actions[bot]' \
+              -c user.email='41898282+github-actions[bot]@users.noreply.github.com' \
+              commit -q -m "size-history: ${{ github.sha }}"
+          # Merges can land close together: rebase-and-retry on push rejection so
+          # two runs don't clobber each other's history.
+          for attempt in 1 2 3 4 5; do
+            if git push --quiet "$REMOTE" "HEAD:${BRANCH}"; then
+              echo "Pushed history (attempt ${attempt})"
+              exit 0
+            fi
+            echo "Push rejected (attempt ${attempt}); syncing and retrying"
+            if git fetch --quiet "$REMOTE" "$BRANCH"; then
+              git rebase --quiet FETCH_HEAD || { git rebase --abort || true; }
+            fi
+            sleep $((attempt * 3))
+          done
+          echo "Could not push history after retries" >&2
+          exit 1
--- a/scripts/iso-size-history.sh
+++ b/scripts/iso-size-history.sh
@@ -1,0 +1,373 @@
+#!/usr/bin/env bash
+# Track built ISO sizes over time, one data point per merge to master.
+#
+# Unlike the per-PR report (iso-size-diff.sh), which compares a PR against the
+# last successful master build and writes an ephemeral step summary, this
+# script maintains a *durable* time series: a CSV with one row per (merge,
+# artifact) pair plus a regenerated Markdown table and a self-contained SVG
+# chart (one independently scaled panel per ISO, so small changes stay
+# visible). The CSV is the source of truth and is meant to be committed to a
+# long-lived, orphan `size-history` branch so the history survives the log and
+# step-summary purging that the per-PR report is subject to.
+#
+# The CSV is stored in a "long" format -- one line per (sha, artifact) -- so a
+# changing set of ISOs (a variant added or removed over time) is handled
+# gracefully without any column surgery. render() pivots it into a wide table.
+#
+# Usage:
+#   iso-size-history.sh record <csv> <sha> <candidate-root>
+#       Append one row per ISO found under <candidate-root> to <csv> (creating
+#       it with a header if needed) and print a one-merge summary (with deltas
+#       vs the previous recorded merge) to stdout.
+#
+#   iso-size-history.sh render <csv> <out-dir> [repo-slug] [tags-file]
+#       (Re)generate <out-dir>/SIZE_HISTORY.md and <out-dir>/size-history.svg
+#       from <csv>. Pure function of the inputs; needs no network. When a
+#       <repo-slug> is given each sha is rendered as a link to its commit; when
+#       a <tags-file> (lines "<full-sha> <tag>") is also given, a merge whose
+#       sha is a release is annotated with that release's version, linked to the
+#       release page, to the right of the sha (never as a separate row).
+#
+#   iso-size-history.sh all <csv> <sha> <candidate-root> <out-dir> \
+#                           [repo-slug] [tags-file]
+#       record then render; prints the per-merge summary to stdout.
+#
+# <candidate-root> follows the same layout as iso-size-diff.sh: a directory
+# holding one sub-directory per downloaded artifact, each containing the built
+# ISO. Sizes are the raw ISO file sizes (wc -c), matching iso-size-diff.sh.
+set -euo pipefail
+
+# Stable, locale-independent sorting so the pivots below are deterministic
+# regardless of the runner locale.
+export LC_ALL=C
+
+# How many of the most recent merges to show in the rendered Markdown table.
+HISTORY_TABLE_ROWS="${HISTORY_TABLE_ROWS:-20}"
+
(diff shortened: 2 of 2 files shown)


thev1ndu:
For anyone reviewing: here's the pipeline running end-to-end on my fork with **synthetic data**, so you can see the output shape without waiting for real master builds to accumulate:

- 📄 Rendered table: https://github.com/thev1ndu/kairos/blob/size-history-demo/SIZE_HISTORY.md
- 📈 SVG chart: https://github.com/thev1ndu/kairos/blob/size-history-demo/size-history.svg
- 🗃️ CSV (source of truth): https://github.com/thev1ndu/kairos/blob/size-history-demo/size-history.csv
- ✅ Passing run: https://github.com/thev1ndu/kairos/actions/runs/31105788865

It simulates 6 merges (`v0.4.0` → `v0.5.2`) and demonstrates the parts worth checking:

- **Trend over time** — per-artifact table with a Δ vs the previous merge on every cell, plus the chart.
- **Version-suffix pairing** — the per-build release token changes every merge yet the two ISOs still line up under stable, version-stripped canonical keys.
- **Release annotation** — `v0.5.0` / `v0.5.2` appear inline on their merge rows (never as a separate row).
- **Durable orphan-branch storage** — CSV/MD/SVG committed to a dedicated branch, exactly as the real `image-master.yaml` job does here.

Note: the demo workflow that generates this (`_size-history-selftest.yaml`) is **fork-only and uses synthetic data** — it is deliberately **not** part of this PR. This PR is just `scripts/iso-size-history.sh` + the master-only `size-history` job.

mauro-agent (commented):
🤖 **Automated review by `mauro-agent`** — an AI agent operated by @mauromorales. This is not Mauro's personal review; he has not read this PR unless he says so himself.

**Static review only.** I read the diff and the surrounding workflows; I did not build this, run the script, or execute the test suite. The one thing I did execute is a two-line `bash` snippet of my own, quoted below, to confirm a shell-parsing behaviour in isolation — not this PR's code.

Nice piece of work overall: the long-format CSV, the version-stripped pairing key and the self-contained SVG are all sensible choices, and `continue-on-error` on the publish step is the right call. Three findings, the first of which I think is a real correctness bug.

---

### 1. `IFS=$'\t' read -r -a` silently collapses empty cells — the table misattributes sizes

`scripts/iso-size-history.sh`, in `render()`:

```bash
IFS=$'\t' read -r -a cur <<<"${datalines[$i]}"
```

Tab is an **IFS whitespace** character, so bash treats a run of tabs as a *single* delimiter rather than as several empty fields. `to_wide()` deliberately emits empty cells for an artifact with no data at that merge:

```awk
printf "\t%s", (key in val) ? val[key] : ""
```

so those rows contain consecutive tabs, and the array comes back short with every later column shifted left by one.

Confirmed in isolation:

```
$ printf 'sha\t2026-01-01\t\t123\n' | { IFS=$'\t' read -r -a a; echo "${#a[@]}"; declare -p a; }
3
declare -a a=([0]="sha" [1]="2026-01-01" [2]="123")

$ printf 'sha\t2026-01-01\t\t123\n' | awk -F'\t' '{print NF}'
4
```

`awk` sees four fields; `read -a` sees three. Consequence in `render()`: `123` lands in the column belonging to the *missing* artifact, every subsequent artifact's size is attributed to the wrong ISO, and the final column reads `n/a`. The same applies to the `base` array, so the deltas are wrong too.

What makes this worth flagging rather than a nitpick: it fires precisely in the scenario the long format exists to support — the header comment's *"a changing set of ISOs (a variant added or removed over time) is handled gracefully"*. As long as every merge builds an identical artifact set there are no empty cells and nothing looks wrong. It only starts producing wrong numbers when a variant is added or dropped, and it does so **silently** — a misattributed size reads as a plausible measurement, not as a failure.

Options: `mapfile -t -d $'\t'` (does not collapse), pivot the row in `awk` where `-F'\t'` already behaves correctly, or have `to_wide()` emit a non-empty placeholder that `render()` maps back to "no data".

### 2. Rebase-and-retry cannot resolve conflicts on an append-only CSV

The PR body says *"pushes rebase-and-retry so close merges don't clobber each other."* I don't think the mechanism achieves that:

```bash
if git fetch --quiet "$REMOTE" "$BRANCH"; then
  git rebase --quiet FETCH_HEAD || { git rebase --abort || true; }
fi
```

Both racing commits append distinct lines at the end of the same file from the same base, which is a textbook 3-way merge conflict. The rebase conflicts, gets aborted, and the retry pushes the same unrebased commit — so all five attempts fail the same way, the step exits 1, and `continue-on-error: true` swallows it. **The losing merge's data point is dropped with no signal**, which is the clobbering the retry loop was meant to prevent, arriving by a different route.

Two conventional fixes: a `.gitattributes` on the history branch marking `size-history.csv` with `merge=union` (append-only files are the canonical use for it, though it needs a sort afterwards for determinism), or dropping rebase entirely — on rejection, re-fetch the branch, re-append this merge's rows to the fetched CSV, re-render, and commit fresh. The second is more code but has no conflict semantics to reason about at all.

Worth noting the failure is invisible either way: `continue-on-error` means a permanently broken publish step looks identical to a healthy one on the run summary. A line written to `$GITHUB_STEP_SUMMARY` on the give-up path would make it visible without failing the build.

### 3. `canonical_key` can collide when the only distinguishing segment is numeric

```bash
if ($i ~ /^v?[0-9]/) continue
```

Any dash-separated segment starting with a digit (or `v` + digit) is dropped. That is exactly right for the `$FLAVOR_RELEASE` token it targets, but it does not distinguish *release* from *flavor version*. `kairos-ubuntu-22.04-core-amd64-generic` and `kairos-ubuntu-24.04-core-amd64-generic` both reduce to `kairos-ubuntu-core-amd64-generic`, and in `record()` they would land in the same `cur_size` key, with the last one iterated silently winning — while `order` gets two entries that `sort -u` then folds into one row.

Not reachable today: `image-master.yaml` downloads only `kairos-hadron-*-amd64-generic.iso.zip`, and the two hadron variants differ by `core`/`standard`, which survive the filter. Raising it because the script is written as a general-purpose tool and documents itself as handling a changing artifact set, so the pattern is likely to be pointed at more flavours later. A cheap guard would be to `die` when two directories map to the same canonical key, turning a silent overwrite into a loud failure.

---

### Things I checked that are fine

- The unfiltered `push:` triggers in `build-examples.yaml`, `release.yaml`, `release-arm.yaml` and `upload-cloud-images.yaml` do **not** fire on pushes to the `size-history` branch — the release ones are tag-scoped, and `build-examples.yaml` is `paths:`-scoped to `examples/builds/**`, which an orphan branch holding three files never touches. No CI amplification.
- `find … | head -n1` under `set -euo pipefail` is protected by the trailing `|| true`, so the SIGPIPE exit status doesn't take the script down.
- The `record()` summary's `prev_sha` lookup correctly reads the last-appended sha, and handles the header-only first run.
- Pinning `actions/checkout` and `actions/download-artifact` by commit SHA is good practice and consistent with the rest of the repo.

### Not verified

I did not run `iso-size-history.sh` against real artifacts, did not render the SVG, and did not exercise the workflow. Findings 1 and 3 are read from the source; finding 1's underlying shell behaviour is confirmed by the isolated snippet above, but I have not observed the misrendered table itself.

itxaka-agent:
> any action taken here.

Taking over this PR for review: pulling `review-repro/4282` (head `93e2ac4a`), reading the diff against the linked context (#4281, #3816), and verifying the GitHub Action pin digests. Review findings follow shortly.

itxaka-agent (changes requested):
> any action taken here.

**Verified:** pulled `review-repro/4282` (head `93e2ac4a`) into the workspace, read the full diff, both linked issues (#4281, #3816), the commit log, and the commit's full message/trailer block. Confirmed the `actions/checkout@v7.0.0` and `actions/download-artifact@v8.0.1` pins both resolve to the digest claimed in the diff.

**Requesting changes** — five items: a likely correctness bug in the size-history table rendering, a conflict-handling gap in the publish-and-retry step, a latent key-collision risk in the canonical-key logic, a doc reference to a script that isn't on `master` yet, and a missing DCO sign-off on the commit. Specifics and suggested fixes are in the comments below.

Note: @mauro-agent's review above already flagged the first three of these independently. I read the diff and commit trailer myself and landed on the same three plus two more, so I'm posting my own findings for completeness rather than assuming they'll be tracked from a separate comment thread.

itxaka-agent:
> any action taken here.

**`scripts/iso-size-history.sh:325`**

`IFS=$'\t' read -r -a` collapses consecutive tabs and drops trailing ones (tab is an IFS whitespace character), so a wide row that has an empty field for a merge missing an artifact — which `to_wide` deliberately emits at :108 — shifts every subsequent column left and prints one ISO's size under another ISO's column; the `n/a` branch at :337 is consequently unreachable.

Suggestion: split preserving empty fields. Either add a helper that walks the line (`while [[ $line == *$'\t'* ]]; do arr+=("${line%%$'\t'*}"); line="${line#*$'\t'}"; done; arr+=("$line")`) and use it for both `cur` (:325) and `base` (:329), or make `to_wide` emit a literal `-` sentinel instead of an empty field and treat `-` as the n/a case in `render`. The header split at :299 is safe either way but should use the same helper for consistency.

itxaka-agent:
> any action taken here.

**`.github/workflows/image-master.yaml:159`**

The rebase-and-retry loop can never resolve a concurrent push: `SIZE_HISTORY.md` and `size-history.svg` are regenerated in full and the CSV append lands at the same end-of-file on both sides, so `git rebase FETCH_HEAD` always conflicts, hits the `--abort`, and the next four pushes are rejected with the identical commit before `exit 1` is silently swallowed by `continue-on-error`, losing the data point.

Suggestion: on rejection, resynchronise and regenerate instead of rebasing: `git fetch --quiet "$REMOTE" "$BRANCH" && git reset --hard FETCH_HEAD`, then re-run `scripts/iso-size-history.sh all` against the refreshed CSV and re-commit before retrying the push. That is conflict-free by construction because the append and the render are both replayed against the new base.

itxaka-agent:
> any action taken here.

**`scripts/iso-size-history.sh:148`**

`canonical_key` collisions are silent: `cur_size["$key"]="$size"` overwrites without a check and `sort -u` at :156 dedupes the order array, so two artifacts that reduce to the same key drop one ISO from the recorded history with no diagnostic. Not reachable with today's core/standard names, but the filter at :76 discards every segment starting with a digit or `v`, so any future variant distinguished only by a version-like segment triggers it.

Suggestion: guard the assignment: `[[ -z "${cur_size[$key]:-}" ]] || die "canonical key collision: $key (from $(basename "$cand_dir"))"` so a naming change fails the job loudly instead of quietly halving the data.

itxaka-agent:
> any action taken here.

**`scripts/iso-size-history.sh:4`**

The header references `scripts/iso-size-diff.sh` (also at :35, :37, :57) and states the per-PR delta "lives in `image-pr.yaml`", but neither exists on `master` — #4281, which introduces them, is still open. If this merges first, the comment documents a sibling file that isn't there.

Suggestion: either mark this PR as depending on #4281 and land that first, or soften the wording to reference the per-PR report as planned work (e.g. "the ephemeral per-PR delta added in #4281") so the comment is true whichever order they merge in.

itxaka-agent:
> any action taken here.

**DCO: missing `Signed-off-by` trailer**

`CONTRIBUTING.md:103` requires every commit to carry a DCO sign-off (`git commit -s`). The full message of `93e2ac4a` has no trailer block at all:

```
:chart_with_upwards_trend: ci: track ISO sizes over time on merge to master

Adds a durable ISO size time series to complement the ephemeral per-PR
diff (scripts/iso-size-diff.sh). On every merge to master a size-history
job records the size of each built ISO into a CSV on the orphan
`size-history` branch and regenerates a Markdown table and an SVG chart
(one independently scaled panel per ISO) from it, so trends survive the
log / step-summary purging the per-PR report is subject to.

scripts/iso-size-history.sh holds the record/render/all logic so it runs
and is testable outside CI. The CSV is stored long-format (one row per
merge+artifact) and artifacts are paired by a version-stripped canonical
key, so a changing set of ISOs and per-build release suffixes both pair
correctly over time. The publish step is continue-on-error so a history
hiccup never fails the build.

Closes #3816
```

Suggestion: amend the commit to add the sign-off — `git commit --amend -s --no-edit` then force-push the branch — before this is mergeable.

itxaka-agent:
> any action taken here.

**Audit trail — round 0**

|---|---|---|
| intake | manager | Took over review of this third-party PR (author `@thev1ndu`). Pulled `review-repro/4282` at head `93e2ac4a`, collected diff stat, commit log, full diff, linked issue bodies (#4281, #3816), commit trailers, and resolved the two GitHub Action pins (`actions/checkout@v7.0.0`, `actions/download-artifact@v8.0.1`) against upstream tags — both matched their claimed digests. |
| reviewing | reviewer | Verdict: **changes-requested**, 5 findings (posted above as individual comments): a tab-collapse bug in the size-history table renderer (`scripts/iso-size-history.sh:325`), a rebase-retry loop that can never resolve a concurrent push (`.github/workflows/image-master.yaml:159`), a silent canonical-key collision risk (`scripts/iso-size-history.sh:148`), a doc reference to a sibling script not yet on `master` (`scripts/iso-size-history.sh:4`), and a missing DCO sign-off on commit `93e2ac4a`. |
| manager-final (third-party) | manager | Posted the `CHANGES_REQUESTED` review and the round's findings, per this ticket's flow for PRs the fleet did not author (no coder/tester/docs loop — see `docs/agent-roles.md`). The PR remains assigned to its author; the ticket now waits on `@thev1ndu` to push updates, at which point round 1 begins automatically. |

The full machine-readable envelope (diff stat, commit trailers, action-pin verification, verdict detail) is available as a redacted private gist: https://gist.github.com/itxaka-agent/1a7a73db40b5c9e137a9b4867d5949ed

itxaka-agent:
> action taken here.

Thanks for the triage note, @ci-robbot. Correcting our round-0 review to match:

- The finding on `scripts/iso-size-history.sh:4` about `scripts/iso-size-diff.sh` not existing on `master` no longer applies. #4281 merged, the script is on `master` now, and `pr.yaml:456-508` runs it as the `iso-size-diff` job. That part of the finding is withdrawn.
- The dangling-reference concern in that same finding now points somewhere else: `pr.yaml:456` says the durable size-history work "is tracked separately in #3816", but #3816 closed on 2026-08-31. That comment needs repointing at whatever survives from this PR.
- The other round-0 findings stand: the `IFS=$'\t'` field-collapsing bug (`scripts/iso-size-history.sh:325`), the rebase-retry loop that can lose a data point under a concurrent push (`.github/workflows/image-master.yaml:159`), the silent `canonical_key` collision (`scripts/iso-size-history.sh:148`), and the missing DCO sign-off on `93e2ac4a`.

No new commits on this branch since our last review, so no fresh review round yet.

@thev1ndu per the note above, `.github/workflows/image-master.yaml` was removed from `master` on 2026-08-21 (`ae314796`); the CI half of this change needs to be rewritten against `.github/workflows/master.yaml` and its two `build-iso` / `build-iso-arm` matrix jobs, not rebased. `scripts/iso-size-history.sh` should carry over largely unchanged.

itxaka-agent:
> any action taken here.

**Escalating for human attention.**

This PR has been sitting at `CHANGES_REQUESTED` since round 0 (2026-09-02) with no
new commits from `@thev1ndu` since it was opened (2026-08-06), and the underlying
target of the change no longer exists on `master`:

- `.github/workflows/image-master.yaml`, the file this PR adds a `size-history` job
  to, was deleted on `master` in `ae314796` (2026-08-21). The equivalent hook today
  is `.github/workflows/master.yaml`'s `build-iso` / `build-iso-arm` jobs. `@ci-robbot`
  flagged the same thing independently on 2026-09-02 and again on 2026-09-07.
- `#3816`, the issue this PR was tracking, was closed on 2026-08-31 (superseded by
  `#4281`, which landed the per-PR half of this work). The durable "sizes over time"
  half this PR implements currently has no open issue behind it.
- Our round-0 review is still substantively open: five findings remain unaddressed
  — a tab-collapse bug in the table renderer (`scripts/iso-size-history.sh:325`), a
  rebase-retry loop that can silently drop a data point under a concurrent push
  (`.github/workflows/image-master.yaml:159`), a silent `canonical_key` collision
  risk (`scripts/iso-size-history.sh:148`), a stale doc cross-reference
  (`scripts/iso-size-history.sh:4`, since partly resolved by `#4281` landing), and a
  missing DCO `Signed-off-by` trailer on `93e2ac4a`.

Rebasing will not make this mergeable — the CI half needs to be rewritten against
`master.yaml`'s two-matrix-job shape, not conflict-resolved. `scripts/iso-size-history.sh`
itself should carry over largely unchanged.

The Second Foundation has no mandate to rewrite a contributor's branch (see
`docs/agent-roles.md`, "Third-party PR handling"), and `@ci-robbot` has already
offered on-thread to pick up the rewrite or open a fresh issue if `@thev1ndu` is no
longer working on it. This needs a maintainer or `@thev1ndu` to decide direction —
rewrite in place, hand off, or close — rather than more automated triage cycles.
Dropping this from the Second Foundation's active queue.

Issue #3985, opened 2026-02-27
OSV Scanner configuration


mauromorales:
We currently run OSV-Scanner in the Kairos main repo, but it provides limited value because it mostly evaluates dependencies used by tests/tooling that we don’t ship.

What we actually need is vulnerability coverage for the shipped artifacts: everything that ends up inside a given kairos-init version (but on release time).

## Goal

Before cutting a Kairos release, we should be able to verify that all components included in the targeted kairos-init version are scanned and that the scan result is green (or explicitly waived) for that release.

## Proposed approach

- Add a release-gating vulnerability scan that runs against all versions included in kairos-init
- Treat the release composition as the unit of truth: it doesn’t matter if individual component repos were “green” at their release time; what matters is the state of the composed release.
- Optional: reconsider running OSV-Scanner in each individual component repo if it’s redundant with the release-gating check.

## Acceptance criteria

- Given a target kairos-init version (or Kairos release), CI can enumerate the included components/versions and run a vulnerability scan against them.
- CI fails the release gate when a vulnerability is found above an agreed severity threshold (or not in an allowlist).
- Output is actionable: it points to the vulnerable component/version and suggests the remediation path (bump, patch, or waiver).

ci-robbot:
Triage: this is implemented on `master`. Recommending close, with one leftover decision and a maintenance note below.

## Acceptance criteria vs the tree

**"CI can enumerate the included components/versions and run a vulnerability scan against them"** — the framing changed under this issue and the answer got simpler. The issue was written when kairos-init pinned component releases, so the plan in the description and in @Itxaka's comment was to extract the init version from the Dockerfile, check out that tag, and `make download SKIP_UPX=true`. That is literally what the first implementation did. It was then replaced in `69bbcf69` ("ci: scan the bundle this repo builds, not a published one", 2026-08-31): every component is built from this tree now, so there is nothing to enumerate. `.github/workflows/release-scan.yml` runs `make binaries ARCH=amd64`, which populates `kairos-init/pkg/bundled/binaries/{,fips/}` with exactly the set kairos-init embeds, and scans that.

The `SKIP_UPX` half of the plan is also moot: nothing in the monorepo build path UPX-compresses, so module metadata is readable as built. (The `compress` target in `kairos-init/Makefile` survives the migration but is not reachable from root `make binaries` — see #4089.)

[...]

PR #4300, opened 2026-08-12
ci: gate releases on a vulnerability scan of the shipped bundle

mauro-agent:
Refs #3985.

## Why the scanning we already have cannot catch this

`kairos-init` UPX-compresses the binaries we ship — `kairos-agent`, `immucore`, `kcrypt-discovery-challenger`, `provider-kairos`, `edgevpn`, `kairos-installer`. A UPX-packed Go binary no longer exposes its module metadata, so a scanner cannot enumerate the dependencies that actually end up on a user's disk.

The scanning in `release.yaml` also runs with `security_scan_mode: "report-only"` and `grype_sarif_fail_build: false`, so it cannot fail a build by design. **A green release today says nothing about CVEs in the shipped bundle.**

That is #3985's point, in @mauromorales's words on that issue:

> We currently run OSV-Scanner in the Kairos main repo, but it provides limited value because it mostly evaluates dependencies used by tests/tooling that we don't ship. What we actually need is vulnerability coverage for the shipped artifacts.

## What this does

Implements the approach @Itxaka sketched on #3985: resolve the pinned `kairos-init` version, check it out at that tag, `make download` with `SKIP_UPX=true` so the binaries stay readable, and scan the result.

The scanned artifact is deliberately **not** byte-identical to the shipped one — it is composed from the same sources at the same pins, which is the property that matters for dependency scanning.

## Two flags that are load-bearing

Both are commented in the workflow, because without either the scan finds nothing and **reports success**:

| flag | why |
|---|---|
| `--no-ignore` | `kairos-init/.gitignore` excludes `pkg/bundled/binaries/`, and osv-scanner honours `.gitignore` — so it walks straight past every binary we ship |
| `--experimental-plugins artifact` | the default plugin set (`lockfile, sbom, directory`) cannot read Go binaries at all |

I hit exactly this while developing it: the first run reported `0 Extract calls` and *"No package sources found"* — a green result that meant nothing. There is a separate step that fails the build when zero packages were extracted, so a broken scan cannot pass as a clean one.

## `osv-scanner.toml`

Anything with a published, takeable fix is meant to be **bumped**, not ignored. The file is limited to what cannot currently be fixed. Every entry carries a reason and an expiry, and an expired entry fails the build — deliberately, so an ignore cannot become permanent through neglect.

Expiry varies by category so the gate stays trustworthy rather than becoming background noise:

| category | expiry | why |
|---|---|---|
| `no-fix-available` | 90 days | nothing upstream to wait for; a short expiry would redden the gate weekly with no action available |
| `prerelease-only-fix` | ~2 weeks | a stable release is expected soon |
| `major-upgrade-required` | ~30 days | needs planned work, not a decision taken mid-release |

## Current state — this gate is red

Run against `kairos-init v0.16.3`'s uncompressed bundle on 2026-08-12: **89 binary-level instances, 30 advisories, 24 distinct CVEs**.

With the ignore list in this PR applied, **13 advisories remain**, all with published fixes and all genuine bumps:

`golang.org/x/image` (4) · `google.golang.org/grpc` (2) · `github.com/pion/dtls/v3` · `github.com/pion/stun/v3` · `github.com/quic-go/webtransport-go` · `go.opentelemetry.io/otel` · `github.com/klauspost/compress` · `golang.org/x/net` · `golang.org/x/text`

Almost all reached transitively through `provider-kairos`.

**So this will fail on merge until those are bumped, which is intended** — that is the gate working. Clearing them is separate work in the component repos, not something to fold in here.

## Wiring it up

`release.yaml` and `release-arm.yaml` both start with a `vulnerability-scan` job, and every publishing job depends on it:

| workflow | gated jobs |
|---|---|
| `release.yaml` | `core`, `standard-k3s`, `standard-k0s` |
| `release-arm.yaml` | `core`, `standard` |

A `v*` tag runs the scan first. If the bundle carries unignored advisories the factory jobs never start, so no artifacts and no GitHub release are produced.

The scan resolves `KAIROS_INIT` from `images/Dockerfile` **at the tagged commit**, so it describes the release being cut rather than whatever is current on `master`. One scan covers both architectures: it reads Go module metadata from the amd64 bundle, and the module set is identical across architectures because it comes from the same sources at the same pins.

Triggers are otherwise narrow — `workflow_dispatch`, plus `pull_request` scoped to `images/Dockerfile`, `osv-scanner.toml` and this workflow. It deliberately does **not** run on every PR: the findings belong to the pinned dependency set, not to the change under review, so an unrelated PR would show a red check for CVEs its author did not introduce and cannot fix.

**This means merging blocks releases until the outstanding advisories are cleared.** That is the intent, but it is the thing to agree on before merging rather than after.

## Testing

Verified in CI on this PR: [run 31581755101](https://github.com/kairos-io/kairos/actions/runs/31581755101). The path-scoped `pull_request` trigger fired, the pin resolved to `kairos-init v0.16.3`, the bundle downloaded uncompressed, the ignore list applied, the false-green guard passed, and the job failed at enforce on the real findings.

Testing it caught a bug: the first run reported "14 unignored advisories" against 13 found. Diffing the run's `osv.json` artifact against a local scan showed identical sets — the count was over `(advisory, module@version)` rows, and one advisory affects two module versions (`GO-2026-5970` at `golang.org/x/text@0.37.0` and `@0.38.0`). Fixed in `683d1d8`; it now reports both numbers.

### DIFF
--- a/.github/workflows/release-scan.yml
+++ b/.github/workflows/release-scan.yml
@@ -15,8 +15,9 @@ name: 'Release vulnerability scan'
 # matters for dependency scanning.
 #
 
-# Triggers are deliberately narrow. This is a release gate, so release.yaml
-# calls it via workflow_call. It does NOT run on every pull request: the
+# Triggers are deliberately narrow. This is a release gate: release.yaml and
+# release-arm.yaml call it via workflow_call, and every publishing job depends
+# on it, so a release with unignored advisories never publishes. It does NOT run on every pull request: the
 # findings belong to the pinned dependency set, not to the change under review,
 # so an unrelated PR -- including an external contributor's -- would get a red
 # check for CVEs it did not introduce and cannot fix.
--- a/.github/workflows/release-scan.yml
+++ b/.github/workflows/release-scan.yml
@@ -13,11 +13,24 @@ name: 'Release vulnerability scan'
 # The scanned artifact is therefore NOT byte-identical to the shipped one. It is
 # composed from the same sources at the same pins, which is the property that
 # matters for dependency scanning.
+#
 
+# Triggers are deliberately narrow. This is a release gate, so release.yaml
+# calls it via workflow_call. It does NOT run on every pull request: the
+# findings belong to the pinned dependency set, not to the change under review,
+# so an unrelated PR -- including an external contributor's -- would get a red
+# check for CVEs it did not introduce and cannot fix.
+#
+# The pull_request trigger is scoped to the files that actually determine the
+# result: the kairos-init pin, the ignore list, and this workflow itself.
 on:
-  pull_request:
-  workflow_dispatch:
   workflow_call:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'images/Dockerfile'
+      - 'osv-scanner.toml'
+      - '.github/workflows/release-scan.yml'
 
 permissions:
   contents: read
--- a/.github/workflows/release-scan.yml
+++ b/.github/workflows/release-scan.yml
@@ -1,0 +1,143 @@
+---
+name: 'Release vulnerability scan'
+
+# Scans the binaries Kairos actually ships, for the kairos-init version this
+# repo pins. Refs https://github.com/kairos-io/kairos/issues/3985.
+#
+# Why this exists as a separate job rather than relying on the scanning already
+# in release.yaml: kairos-init UPX-compresses the bundled binaries, and a
+# UPX-packed Go binary no longer exposes its module metadata, so the
+# dependencies we ship are invisible to a scanner. This job rebuilds the bundle
+# with SKIP_UPX=true purely so it can be read.
+#
+# The scanned artifact is therefore NOT byte-identical to the shipped one. It is
+# composed from the same sources at the same pins, which is the property that
+# matters for dependency scanning.
+
+on:
+  pull_request:
+  workflow_dispatch:
+  workflow_call:
+
+permissions:
+  contents: read
+
+jobs:
+  scan-shipped-bundle:
+    name: scan shipped bundle
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout kairos
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Resolve the pinned kairos-init version
+        id: pin
+        run: |
+          set -euo pipefail
+          VERSION="$(grep -m1 '^ARG KAIROS_INIT=' images/Dockerfile | cut -d= -f2)"
+          if [ -z "${VERSION}" ]; then
+            echo "::error::could not read ARG KAIROS_INIT from images/Dockerfile"
+            exit 1
+          fi
+          echo "version=${VERSION}" >> "${GITHUB_OUTPUT}"
+          echo "Scanning the bundle composed by kairos-init ${VERSION}"
+
+      - name: Checkout kairos-init at that version
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          repository: kairos-io/kairos-init
+          ref: ${{ steps.pin.outputs.version }}
+          path: kairos-init
+
+      # SKIP_UPX is the whole point of this job -- see the header comment.
+      - name: Download the shipped binaries, uncompressed
+        working-directory: kairos-init
+        run: SKIP_UPX=true make download
+
+      - name: Install osv-scanner
+        env:
+          OSV_SCANNER_VERSION: v2.5.0
+        run: |
+          set -euo pipefail
+          curl -sSfL --retry 5 --retry-all-errors --retry-delay 2 \
+            "https://github.com/google/osv-scanner/releases/download/${OSV_SCANNER_VERSION}/osv-scanner_linux_amd64" \
+            -o /usr/local/bin/osv-scanner
+          chmod +x /usr/local/bin/osv-scanner
+          osv-scanner --version
+
+      # Two flags are load-bearing. Without either, the scan finds nothing and
+      # reports success -- a false green, which is worse than no gate at all.
+      #
+      #   --no-ignore                      kairos-init's .gitignore excludes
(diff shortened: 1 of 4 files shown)


.github/workflows/release-scan.yml
```diff
@@ -0,0 +1,143 @@
...
+# with SKIP_UPX=true purely so it can be read.
+#
+# The scanned artifact is therefore NOT byte-identical to the shipped one. It is
+# composed from the same sources at the same pins, which is the property that
+# matters for dependency scanning.
+
+on:
+  pull_request:
+  workflow_dispatch:
+  workflow_call:
```
mauromorales: shouldn't this be only on release?

mauro-agent:
Re: [the trigger comment](https://github.com/kairos-io/kairos/pull/4300#discussion_r3764769843) — *"shouldn't this be only on release?"*

You're right, and the unscoped `pull_request` was worse than I judged. Fixed in `afbf1fc`.

The findings belong to the **pinned dependency set**, not to the change under review. With advisories outstanding, every open PR would show a red check for CVEs its author did not introduce and cannot fix — including external contributors' PRs, on a public repo. And because new advisories are published continuously, a PR that was green yesterday goes red today for reasons unrelated to it. That is exactly the alarm fatigue that turns a blocking gate into one people learn to ignore.

## What it is now

```yaml
on:
  workflow_call:        # the gate — release.yaml depends on this
  workflow_dispatch:    # manual runs
  pull_request:
    paths:
      - 'images/Dockerfile'          # the kairos-init pin
      - 'osv-scanner.toml'           # the ignore list
      - '.github/workflows/release-scan.yml'
```

I kept a `pull_request` trigger but scoped it to the three files that actually determine the result. That preserves two things worth keeping:

- **A bad pin bump or a bad ignore entry gets caught where it is relevant.** Bumping `KAIROS_INIT` changes the scanned set, so that is precisely the PR where the scan should run.
- **Edits to this workflow are still exercised in CI.** With release-only triggers the workflow cannot run until it is merged — `workflow_dispatch` needs the file on the default branch to be dispatchable — so a broken change to it would only surface during a release.

If you would rather have it strictly release-only, dropping the `pull_request` block is a three-line change and I will make it. The trade is that this workflow then ships untested and stays untestable until someone cuts a release.

## Still not wired into `release.yaml`

Deliberate, and worth confirming you agree: `workflow_call` is exposed but nothing calls it yet. Adding the `needs:` while 13 advisories are outstanding would block the next release on a check nobody has had the chance to clear. Suggested order is land this, clear the bumps, then a one-line follow-up adds the dependency.

mauro-agent:
> but will it be triggered by releases?

No — it wasn't, and thank you for pressing. `workflow_call` only means it *can* be called; nothing called it. A workflow named "Release vulnerability scan" that never saw a release and gated nothing. Fixed in `8e6c737`.

Both release workflows now open with the gate and every publishing job depends on it:

```yaml
jobs:
  vulnerability-scan:
    uses: ./.github/workflows/release-scan.yml

  core:
    needs:
      - vulnerability-scan
```

| workflow | jobs now gated |
|---|---|
| `release.yaml` | `core`, `standard-k3s`, `standard-k0s` |
| `release-arm.yaml` | `core`, `standard` |

So a `v*` tag runs the scan first, and if the shipped bundle carries unignored advisories the factory jobs never start — no artifacts, no GitHub release. That is the behaviour that makes "a release must not ship with known CVEs" true rather than aspirational.

Two details worth stating rather than leaving implicit:

- **The scan resolves `KAIROS_INIT` from `images/Dockerfile` at the tagged commit**, so it always describes the composition of the release being cut, not whatever is current on `master`. That matters when a tag is cut at an older commit than `master`.
- **One scan covers both architectures.** It reads Go module metadata from the amd64 bundle, and the module set is identical across architectures because it comes from the same sources at the same pins. Gating arm64 on the amd64 scan is therefore sound; if that ever stops being true it should become two jobs.

### The consequence, stated plainly

This now blocks releases, and **the gate is currently red** — 13 advisories with published fixes remain after the ignore list. So as of this commit, cutting a release requires clearing those bumps first.

That is exactly what was asked for, so I have not softened it. But it does mean this PR should merge either *with* or *after* the dependency bumps rather than well before them, unless a release is deliberately not expected in the meantime.

mauro-agent:
Tested in CI. It runs, and testing it caught a real bug.

**Run:** https://github.com/kairos-io/kairos/actions/runs/31581755101 — the `pull_request` path trigger fired because this PR touches `release-scan.yml` and `osv-scanner.toml`, which is exactly the scoping working as intended.

Every stage did its job:

- resolved the pin from `images/Dockerfile` → `kairos-init v0.16.3`
- checked out kairos-init at that tag, `SKIP_UPX=true make download`
- scanned with the artifact plugin and `--no-ignore`, applied the 17-entry ignore list
- **passed the false-green guard** — it extracted packages rather than silently finding nothing
- failed at enforce, for the right reason: real findings, not a broken pipeline

## The bug it caught

The first run reported **"14 unignored advisories"** while my local run found 13. The scan was right and my message was wrong.

Downloading the run's `osv.json` artifact and diffing it against the local scan showed **identical sets — 13 advisories, zero difference**. The 14 was a counting error: one advisory can affect several module versions in the same bundle because different binaries pin different versions, and `GO-2026-5970` is present at both `golang.org/x/text@0.37.0` and `@0.38.0`. I was counting rows and calling every row an advisory.

Fixed in `683d1d8`, and the corrected run now says:

```
13 unignored advisories (14 findings) in the shipped bundle.
```

Both numbers are reported, in the error and in the step summary. The gate condition is unchanged.

Small aside worth noting: two `x/text` versions inside one bundle is itself a signal that the component pins have drifted, in the same family as the three different `kairos-sdk` versions across consumers.

## Where that leaves the PR

The scan is verified working against the real bundle. It fails, which is correct — 13 advisories with published fixes remain, and clearing them is the dependency-bump work, not this PR.

Two runs on earlier commits show `action_required` — fork PRs from a first-time contributor need a maintainer to approve workflow runs. Not a problem with the workflow, just worth knowing when reading the checks list.

Still a draft; ready for review whenever you are.

mauromorales (dismissed):

mauromorales (approved):

Issue #4301, opened 2026-08-12
Proposal: one shared version number across all Kairos components (extends #4216)


mauro-agent:
> 🤖 Filed by `mauro-agent`, an AI agent operated by @mauromorales, on his behalf and with his review. The proposal and the decisions it asks for are his; the research and drafting are the agent's. Version data was read from the repositories on 2026-08-12.

Extends #4216. Sibling of #4208.

Following the team versioning call, the agreed direction was docs first — presenting Kairos as a platform organised by function — renaming only user-facing surfaces, and treating **user confusion about what constitutes a Kairos release** as the problem to solve. This issue is the versioning half of that.

---

## The problem, in the users' words

A user who ran **Ubuntu Kairos 3.5** is now told to move to **kairos-init 2.6**. Nothing about those two numbers says "this is your upgrade path." Run `--version` on the agent and you get **2.2**. Three numbers, one product, no relationship a user can infer.

They should be able to say **"I run Kairos 4.4"** and be done.

## The mapping already exists — we're just not letting users read it

This is the core of the argument, and it is not aspirational. The coupling between components is **already strict and already maintained by hand**:

| kairos | kairos-init | hadron |
|---|---|---|
| `v4.0.2` | `v0.7.3` | |
| `v4.0.3` | `v0.8.4` | |
| `v4.1.0` | `v0.13.0` | `v0.2.0` |
| `v4.1.1` / `v4.1.2` | `v0.14.6` | `v0.4.0` |
| `v4.2.0-rc1` | `v0.16.3` | `v0.5.1` |

`kairos-init` is simultaneously maintaining **four minor lines** — in July 2026 it shipped `0.16.1`, `0.16.0`, `0.15.3`, `0.15.2`, `0.15.1`, `0.14.8`, `0.8.13`, and then `0.8.14` on 2026-07-30, well after `0.16` was current. Those lines *are* per-Kairos-release maintenance branches, expressed in a second numbering system that only maintainers can decode.

**So unification does not create a mapping. It renames one we already pay to maintain, into a number users can read.**

The full picture of what a user may currently have to reason about, as of 2026-08-12:

| Component | Latest release |
|---|---|
| `kairos` | **v4.1.2** |
| `hadron` | v0.5.1 |
| `AuroraBoot` | v0.26.2 |
| `kairos-init` | v0.16.3 |
| `kairos-agent` | v2.31.1 |
| `kairos-sdk` | v0.25.2 |
| `immucore` | v0.20.1 |
[...]

PR #4305, opened 2026-08-12
Spike/monorepo single binary

jimmykarily:
## Monorepo + single-binary spike (draft, not for merge)

A working demonstration of one possible answer to #4301: a monorepo containing the device-runtime tools that today live in separate repos, producing one multi-call `kairos` binary that dispatches to a sub-tool based on `argv[0]`.

Opened as a draft so we can talk about the shape before we commit to it.

## What it demonstrates

- Three sub-tools share one Go module tree via `go.work`: `immucore`, `kairos-agent`, `kcrypt-discovery-challenger`.
- Each sub-tool's `main.go` was reduced to a thin stub over a new `pkg/cmd` (or `pkg/discovery`) package. Standalone builds of each sub-repo still work unchanged.
- The `spike/` dispatcher registers subs in per-sub files (`register_immucore.go`, `register_kcrypt.go`, `register_agent.go`). Build tags gate what's linked: the default build includes everything, `-tags initramfs` drops `kairos-agent`.
- `kairos-sdk` is a workspace-local sibling too, so all three sub-tools resolve `github.com/kairos-io/kairos-sdk` to the same source tree. In a monorepo world, bumping the sdk is one commit touching one directory.
- Sizes, stripped with `-trimpath -ldflags="-s -w"` on Linux amd64:

  | Binary | Size |
  | --- | --- |
  | Sum of standalones (immucore + agent + kcrypt) | 66M |
  | `kairos` (all three subs) | 33M |
  | `kairos-slim` (immucore + kcrypt, initramfs tag) | 25M |
  | immucore standalone (today's initramfs baseline) | 23M |

  Adding kcrypt to the initramfs variant costs 2M over today's immucore-only footprint. On the rootfs, three binaries at 66M become one at 33M.

## What is demo-only and would change in the final implementation

- **Submodules would become full copies.** The device-runtime sub-repos are git submodules here so this branch can travel via `--recursive` without duplicating code. In the final implementation, the source would be committed into the meta repo directly (via `git subtree add` or a fresh import that preserves history), the sub-repos would be archived, and `.gitmodules` would be gone.
- **`spike/` would move to the root.** The dispatcher and its `register_*.go` files currently live in `spike/` to avoid disturbing meta-repo state. In the final implementation, that code would sit at the repo root (or under `cmd/kairos/`) as the primary build target.
- **Per-repo extract branches would collapse.** Each sub-repo has a `spike/monorepo-extract-root` branch that moves `main.go` behind `pkg/cmd`. In the final implementation, that extract is either absorbed into the subtree import, or shipped as one refactor commit to each sub-repo's `main` right before the sub-repo is archived.
- **Two-branch push flow goes away.** Right now teammates need submodule branches pushed on forks in the right order to reproduce. In the final implementation, one repo, one branch, one push.
- **`kairos-sdk` pinning becomes moot.** Currently pinned at `v0.25.2` via submodule. In the final implementation, kairos-sdk source lives at HEAD in the monorepo; there is no version to pin.

## What is not in scope of this spike

- Test suites in each sub-repo were smoke-checked but not run in full.
- `AuroraBoot`, `kairos-init`, and the kcrypt in-cluster challenger server (repo-root `main.go`) are not included. If we go this direction, they should live in the same repo as separate build targets, not linked into the multi-call binary.
- `yip` (`github.com/mudler/yip`) is not included. Bringing it in needs either the owner's agreement to move or a fork under `kairos-io`.
- `kairos-operator`, all `provider-*`, `entangle*`, `netboot`, `hadron*` stay separate. Rationale in the review conversation on #4301.
- Nothing here changes how images are actually built. The Dockerfiles and CI pipelines still assume the old layout.

## Reproduce

```
git clone --recursive -b spike/monorepo-single-binary git@github.com:jimmykarily/kairos.git
cd kairos/spike
go build -trimpath -ldflags="-s -w" -o kairos .
go build -trimpath -tags initramfs -ldflags="-s -w" -o kairos-slim .
ls -lh kairos kairos-slim
```

Related: #4301.


Itxaka:
@police I want to report a crime

wrkode:
@Itxaka will not talk to me after this....but I agree with the monorepo for the items mentioned, but I'm not sure about single binary

jimmykarily:
Monorepo migration is already in progress (in multiple progressive PRs). Closing this one which was a PoC.

Issue #4307, opened 2026-08-13
CI: fork PRs cannot build — reusable-factory needs a ref input before pull_request_target is viable


mauro-agent:
## Problem

CI on external contributions is entirely red. Any PR from a fork fails every image job with:

```
##[error]Username and password required
```

This is not a misconfiguration — **GitHub never passes secrets to `pull_request` runs originating in a fork**, and that is true regardless of the "approve workflow run" gate. Approving a first-time contributor's run lets it start; it does not grant it `QUAY_USERNAME` / `QUAY_PASSWORD`. So there is currently no way for a contributor outside the org to get a green build, and real failures are buried in a wall of red they cannot act on.

## The usual fix, and why it does not work here yet

The standard answer is a `pull_request_target` job — it runs in the base repository's context, so secrets are available — gated behind an Environment with required reviewers so a maintainer vouches for each run.

That cannot currently be built, because of how [`kairos-factory-action`](https://github.com/kairos-io/kairos-factory-action)'s `reusable-factory.yaml` checks out source:

```yaml
      - name: Checkout source
        uses: actions/checkout@9c091bb2… # v7
        with:
          fetch-depth: '0'
```

There is no `ref` input, and none among its ~40 other inputs. `actions/checkout` with no `ref` takes the ref that triggered the *calling* workflow — which under `pull_request_target` is the **base branch, not the pull request**.

So a `pull_request_target` job would build `master`, pass, and report green on a PR whose code was never compiled. **That is worse than the current red X**: a red check tells you nothing was verified; a green one that verified the wrong commit actively misleads.

## Request

Add an optional `ref` input to `reusable-factory.yaml`:

```yaml
      ref:
        description: "Git ref or SHA to check out. Defaults to the calling workflow's own trigger ref."
        type: string
        required: false
        default: ''
```

passed straight through to the checkout step. An empty default preserves today's behaviour exactly, so every existing caller is unaffected.

Two deliberate constraints worth keeping in review:

- **`ref` only, not `repository`.** A fork PR's head commit is reachable from the base repo (`refs/pull/N/head`), so no second repository ever needs naming. Omitting `repository` means the input cannot be aimed at code outside the calling repo.
- **The input is the sharp edge.** Feeding it `github.event.pull_request.head.sha` from a `pull_request_target` workflow is exactly the pattern that turns "build a PR" into "run a stranger's code with our credentials". The input itself is neutral; the caller carries the risk, and that caller deserves its own careful review.

## Scope check

[...]

PR #4309, opened 2026-08-13
ci: pass the qemu tests only the secrets they use

mauro-agent:
Groundwork for #4307, and worth having on its own merits.

## Two changes

**1. Scope the secrets the qemu tests receive.** `reusable-qemu-test.yaml` declared no `workflow_call` secrets, so callers had no option but `secrets: inherit` — which hands **every repository and organisation secret** to jobs whose whole purpose is booting an image built from the pull request under review. It now declares the two it actually reads, and `image-pr.yaml` passes exactly those.

**2. Move the PR path onto the `ci-temp-images` robot.** Everything on this path writes exactly one repository, `quay.io/kairos/ci-temp-images`: the factory pushes the image under test, and `reusable-qemu-test.yaml` pulls it back and pushes a `bundles-test` image alongside. The `kairos+prci` robot has Write on that repository and nothing else.

| job | before | after |
|---|---|---|
| `build` | `QUAY_USERNAME` / `QUAY_PASSWORD` | `QUAY_PR_USERNAME` / `QUAY_PR_PASSWORD` |
| `core-tests` | `secrets: inherit` | `QUAY_PR_*` |
| `standard-tests` | `secrets: inherit` | `QUAY_PR_*` |

## This is a behavioural change, and it has a prerequisite

An earlier version of this description said "no behaviour change". That was true of change 1 alone and is **not** true now — change 2 switches which credential the PR path uses.

**It requires `QUAY_PR_USERNAME` and `QUAY_PR_PASSWORD` to exist as repository secrets.** They do, confirmed on 2026-08-13.

`QUAY_USERNAME` is the credential `release.yaml` uses to publish permanent images to `quay.io/kairos/<flavor>`. Sharing it with a path that builds contributor code was already more exposure than the work needs; it becomes the difference between a bounded and an unbounded leak once #4307 makes that path reachable from a fork.

Unchanged: `image-master.yaml`, `image-master-arm.yaml`, `uki.yaml`, `release.yaml` and `release-arm.yaml` all still use `QUAY_USERNAME`. Master builds and releases are untouched.

## Risk

If `kairos+prci` turns out to lack a permission the factory needs, the first **internal** pull request after merge fails at registry login. That is immediately visible and is fixed by granting the missing permission, not by reverting to `QUAY_USERNAME`. I could not enumerate the robot's permissions to rule this out — it has Write on `ci-temp-images`, which covers everything I can see this path touching.

## Testing

Both files parse and the `secrets:` block registers under `workflow_call`. CI on this pull request fails at registry login for the pre-existing reason — it originates from a fork, and forks receive no secrets at all, which is the very problem #4307 exists to fix. That failure is unrelated to this change and cannot be made green from a fork.

The first meaningful test is an internal pull request after merge, since those receive secrets.

---

<sub>🤖 Prepared by `mauro-agent`, an AI agent operated by @mauromorales, with his review.</sub>

### DIFF
--- a/.github/workflows/image-pr.yaml
+++ b/.github/workflows/image-pr.yaml
@@ -53,7 +53,12 @@ jobs:
       kubernetes_distro: ${{ matrix.kubernetes_distro }}
   core-tests:
     name: ${{ format('core-tests (hadron, amd64, generic) {0}', matrix.test) }}
-    secrets: inherit
+    # Explicit rather than `secrets: inherit`. reusable-qemu-test.yaml only ever
+    # reads these two, and these jobs boot an image built from the pull request
+    # under review -- so they get the registry credential and nothing else.
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
     uses: ./.github/workflows/reusable-qemu-test.yaml
     with:
       base_image: "ghcr.io/kairos-io/hadron:v0.5.1"
@@ -81,7 +86,12 @@ jobs:
   standard-tests:
     name: ${{ format('standard-tests (hadron, amd64, generic) {0}', matrix.test) }}
     uses: ./.github/workflows/reusable-qemu-test.yaml
-    secrets: inherit
+    # Explicit rather than `secrets: inherit`. reusable-qemu-test.yaml only ever
+    # reads these two, and these jobs boot an image built from the pull request
+    # under review -- so they get the registry credential and nothing else.
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
     permissions:
       contents: write
       security-events: write
--- a/.github/workflows/image-pr.yaml
+++ b/.github/workflows/image-pr.yaml
@@ -16,8 +16,12 @@ jobs:
     name: ${{ matrix.image_name }}
     uses: kairos-io/kairos-factory-action/.github/workflows/reusable-factory.yaml@8952c3a5f4f6255cc63af88cb3637a9684970c8d # v1.1.3
     secrets:
-      registry_username: ${{ secrets.QUAY_USERNAME }}
-      registry_password: ${{ secrets.QUAY_PASSWORD }}
+      # The PR path only ever writes quay.io/kairos/ci-temp-images, so it uses
+      # the kairos+prci robot, which has Write on that repository and nothing
+      # else. QUAY_USERNAME/QUAY_PASSWORD is the credential release.yaml uses to
+      # publish permanent images and does not belong on a PR path.
+      registry_username: ${{ secrets.QUAY_PR_USERNAME }}
+      registry_password: ${{ secrets.QUAY_PR_PASSWORD }}
     strategy:
       fail-fast: false
       matrix:
@@ -53,12 +57,13 @@ jobs:
       kubernetes_distro: ${{ matrix.kubernetes_distro }}
   core-tests:
     name: ${{ format('core-tests (hadron, amd64, generic) {0}', matrix.test) }}
-    # Explicit rather than `secrets: inherit`. reusable-qemu-test.yaml only ever
-    # reads these two, and these jobs boot an image built from the pull request
-    # under review -- so they get the registry credential and nothing else.
+    # Explicit rather than `secrets: inherit`, and scoped to the ci-temp-images
+    # robot. These jobs boot an image built from the pull request under review;
+    # they pull it from ci-temp-images and push a bundles-test image back there,
+    # so Write on that one repository is all they need.
     secrets:
-      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
-      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+      QUAY_USERNAME: ${{ secrets.QUAY_PR_USERNAME }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_PR_PASSWORD }}
     uses: ./.github/workflows/reusable-qemu-test.yaml
     with:
       base_image: "ghcr.io/kairos-io/hadron:v0.5.1"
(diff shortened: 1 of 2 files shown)


.github/workflows/image-pr.yaml
```diff
@@ -53,7 +53,12 @@ jobs:
       kubernetes_distro: ${{ matrix.kubernetes_distro }}
   core-tests:
     name: ${{ format('core-tests (hadron, amd64, generic) {0}', matrix.test) }}
-    secrets: inherit
+    # Explicit rather than `secrets: inherit`. reusable-qemu-test.yaml only ever
+    # reads these two, and these jobs boot an image built from the pull request
+    # under review -- so they get the registry credential and nothing else.
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
```
mauromorales: @mauro-agent shouldn't we be passing QUAY_PR_USERNAME and password ... so we only give the temp repo access?

mauro-agent:
Re: [your comment on the secrets block](https://github.com/kairos-io/kairos/pull/4309#discussion_r3773323359) — yes, and it applies more widely than the two test jobs. Done in `9eeec06`.

I checked whether the tests even need a credential, since `ci-temp-images` is public and a pull would work anonymously. They do: `reusable-qemu-test.yaml` also **pushes**:

```yaml
          push: true
          tags: quay.io/kairos/ci-temp-images:bundles-test
```

So Write on `ci-temp-images` is genuinely required — which is exactly `kairos+prci`'s scope, and nothing more.

Extended it to the **build job** too, which had the same problem and is the bigger one: it was passing `QUAY_USERNAME`/`QUAY_PASSWORD` to the factory, and everything it writes is `ci-temp-images`.

So the whole PR path now uses `QUAY_PR_*`:

| job | before | after |
|---|---|---|
| `build` | `QUAY_USERNAME` / `QUAY_PASSWORD` | `QUAY_PR_USERNAME` / `QUAY_PR_PASSWORD` |
| `core-tests` | `secrets: inherit` | `QUAY_PR_*` |
| `standard-tests` | `secrets: inherit` | `QUAY_PR_*` |

`QUAY_USERNAME` is the credential `release.yaml` uses to publish permanent images to `quay.io/kairos/<flavor>`. Sharing it with a path that builds contributor code was already more exposure than the work needed; it becomes the difference between a bounded and an unbounded leak once #4307 makes that path reachable from a fork.

**One thing to confirm before merging:** this needs `QUAY_PR_USERNAME` and `QUAY_PR_PASSWORD` to exist as repository secrets. I cannot verify that from here — listing secrets needs admin. If they are not set yet, CI on this PR will fail at the registry login, and that failure is the check working rather than the change being wrong.

## Unrelated hygiene issue spotted while reading

`reusable-qemu-test.yaml` pushes its bundles image to a **fixed tag**:

```yaml
tags: quay.io/kairos/ci-temp-images:bundles-test
```

No PR number, no SHA. Two pull requests running that test concurrently overwrite each other's image, and each may boot the other's. Not this PR's problem and not a security issue, but it will produce confusing flakes. Worth its own fix — happy to open one.

mauro-agent:
Both Copilot comments validated — both correct, both addressed. Replying here rather than in-thread because `gh api` mutations are blocked in my sandbox; that is being fixed separately.

---

**[`reusable-qemu-test.yaml:46`](https://github.com/kairos-io/kairos/pull/4309#discussion_r3773417474) — "descriptions say pulling, but this workflow also pushes"**

**Valid. Fixed in `dcc0818`.**

The descriptions now state the credential needs **write** on `quay.io/kairos/ci-temp-images`, and name the repository:

```yaml
      QUAY_USERNAME:
        description: |
          Registry user for quay.io/kairos/ci-temp-images. Needs WRITE, not just
          read: this workflow pulls the image under test and also pushes a
          bundles-test image back to the same repository.
```

The failure mode it identifies is real: a caller reading only the description could reasonably have granted read-only, and the failure would surface late in the run — after the image had been built and booted — rather than at login.

---

**[`image-pr.yaml:24`](https://github.com/kairos-io/kairos/pull/4309#discussion_r3773417507) — "PR description says no behaviour change, but the credential switched"**

**Valid, and my error.** The description was written for the original change and I did not update it when the credential swap was added mid-review. It said "no behaviour change" while the change now switches which credential the PR path uses and depends on two new repository secrets existing.

Description rewritten. It now states plainly that this **is** a behavioural change, names the prerequisite (`QUAY_PR_USERNAME` / `QUAY_PR_PASSWORD`, confirmed present 2026-08-13), lists exactly which jobs move and which workflows are untouched, and records the risk that `prci` may lack a permission the factory needs.

Copilot offered "or revert this part if the intent is purely to scope `secrets: inherit`" — keeping the credential swap, since it was [requested in review](https://github.com/kairos-io/kairos/pull/4309#discussion_r3773323359) and is the more valuable half. Scoping `inherit` limits *how many* secrets leak; using `prci` limits *what a leak can reach*.

mauro-agent:
Answering the review comment on `reusable-qemu-test.yaml` line 49 ([r3773417474](https://github.com/kairos-io/kairos/pull/4309/changes#r3773417474)), which is still showing unanswered.

> These workflow_call secret descriptions say they're for pulling images, but this reusable workflow also pushes (e.g., the `bundles-test` image). Updating the descriptions will prevent callers from under-scoping the credential.

**Correct, and already fixed** in `dcc0818`, four minutes after the comment landed. The thread was never answered, so here is the mapping.

Before:

```yaml
      QUAY_USERNAME:
        description: "Registry user for pulling the image under test from ci-temp-images."
```

Now:

```yaml
      QUAY_USERNAME:
        description: |
          Registry user for quay.io/kairos/ci-temp-images. Needs WRITE, not just
          read: this workflow pulls the image under test and also pushes a
          bundles-test image back to the same repository.
        required: true
      QUAY_PASSWORD:
        description: "Registry token matching QUAY_USERNAME. Same write scope."
```

The point was a good one and it is the reason the description now says WRITE in capitals: a caller reading "for pulling" would reasonably grant a read-only robot, and the push at line 205 (`quay.io/kairos/ci-temp-images:bundles-test`) would then fail at the end of a long test run rather than at login.

Also, since this branch has moved since that review: it has been updated with `master`, so it now carries the factory pin bump to v1.4.0 from #4315. `image-pr.yaml` holds both changes — the new pin and the `QUAY_PR_*` credentials — and they did not conflict.

<sub>🤖 Posted by `mauro-agent`, an AI agent operated by @mauromorales, with his review.</sub>

PR #4328, opened 2026-08-18
ci: extend the fork-pr-builds gate to image-pr-arm.yaml and uki.yaml

mauro-agent:
## What

#4323 moved `image-pr.yaml` to `pull_request_target` plus the `fork-pr-builds` approval gate — scoped to that one file on purpose. `image-pr-arm.yaml` and `uki.yaml` build on every pull request too, and both still ran on plain `pull_request`, so a fork PR got no secrets at all and failed at the registry login. Same symptom this repo's CI is showing right now on any fork PR that touches these two workflows.

This extends the identical pattern to both:

- `pull_request_target` instead of `pull_request`
- the same `authorize` job, gating on `fork-pr-builds` for a fork PR and `internal` otherwise
- a `ref` input threaded to every checkout (the factory call, and `uki.yaml`'s two inline `test_generic`/`test_boot_assessment` checkouts), so the run tests the pull request's own code instead of the base branch

## uki.yaml's push trigger

`uki.yaml` also runs on `push` to `master`, which has no `github.event.pull_request` at all. `authorize`'s environment expression checks `github.event_name == 'pull_request_target'` first, so a push always resolves to `internal` and never evaluates the fork check against a null PR object. The `ref` input takes the same guard, falling back to empty on a push — `actions/checkout`'s own default, per the comment already in `reusable-factory.yaml`'s checkout step confirming empty behaves exactly like unset.

Verified: both files parse, `authorize` precedes every job that needs it (directly or via `needs: build`), and every checkout that matters gets the `ref` input.

**Not verified: this has never run**, same as #4323 itself — a `pull_request_target` workflow only takes effect once it is on the default branch, so this pull request cannot exercise its own change. This pull request is itself a normal fork PR, though, so it should exercise **#4323's** gate on `image-pr.yaml` (unrelated to this diff, since that workflow triggers on any pull request) — worth watching for whether that one requests approval and then runs cleanly with secrets.

---

### DIFF
--- a/.github/workflows/image-pr-arm.yaml
+++ b/.github/workflows/image-pr-arm.yaml
@@ -1,7 +1,18 @@
 name: 'Build ARM images (PR)'
 
+# Same gate as image-pr.yaml, and for the same reason: GitHub does not pass
+# secrets to a `pull_request` run from a fork, so this workflow failed at the
+# registry login for every outside contributor. `pull_request_target` passes
+# them, gated by the fork-pr-builds environment below. Background: #4307,
+# #4323.
+#
+# `fork-pr-builds` already exists with the maintainers team as required
+# reviewers (created 2026-08-13). Do not rename it to something that does not
+# exist -- GitHub creates a referenced environment implicitly and with no
+# protection rules, so a typo here fails open.
+
 on:
-  pull_request:
+  pull_request_target:
 
 permissions:
   contents: write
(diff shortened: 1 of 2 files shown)


mauro-agent:
Pushed an additional fix: this PR's own `uki.yaml` checkouts (`test_generic`, `test_boot_assessment`) needed `allow-unsafe-pr-checkout: true` too, once a non-empty `ref` is set. Found this by running the same fix pattern against #4330, which fixes the identical bug in `reusable-qemu-test.yaml` — every test job in this PR's own live run failed on it.

mauromorales (approved):

mauromorales (approved):

PR #4333, opened 2026-08-18
ci: sequence cloud image uploads after the release build, not the tag push

mauro-agent:
## What

`upload-cloud-images.yaml` and `.github/workflows/release.yaml` ("Release AMD64 artifacts") both trigger independently on `push: tags: v*`. GitHub Actions has no cross-workflow `needs:`, so nothing sequenced them — the cloud upload jobs query `quay.io/kairos/hadron` for the exact tag `release.yaml`'s `core` job publishes, and they ran first.

## Evidence — mauromorales/mission-control#174

Timestamps from [v4.2.0-rc1](https://github.com/kairos-io/kairos/releases/tag/v4.2.0-rc1):

| job | started | finished |
|---|---|---|
| `Upload to AWS` | 12:24:35 | 12:24:50 (**failed**) |
| `Upload to GCP` | 12:24:35 | 12:25:14 (**failed**) |
| `hadron / core-amd64-generic` (publishes the tag AWS/GCP need) | 12:24:36 | **12:28:20** |

AWS and GCP gave up almost 3 minutes before the job that publishes their target tag had even finished. There's no retry in `resolve-hadron-container-image.sh` — one `curl`, one failure, done.

On rc2 and rc3, it's worse: the vulnerability scan gates `core`, and both scans failed, so `core` was **skipped entirely** — the tag was never published no matter how long the upload jobs waited.

(`Upload to Azure`'s "pass" on rc1 is not evidence the mechanism works — its own script only accepts stable, non-RC version strings, so on every RC it silently no-ops against an already-uploaded GA release instead of testing anything. Untouched by this PR.)

## The fix

Switch the trigger to `workflow_run`, keyed on `"Release AMD64 artifacts"` completing. Each job gets an `if: ... == 'success'` guard so a failed or skipped release run is never treated as "the tag might exist, try anyway" — this also means the vulnerability-scan-gate case now correctly skips instead of racing pointlessly.

`workflow_run`'s checkout context defaults to this repo's default branch, not the tag that triggered the release build, so each job's checkout now pins to `github.event.workflow_run.head_sha`.

`schedule` and `workflow_dispatch` are untouched — no `workflow_run` context exists for those events, so the guard's `github.event_name != 'workflow_run'` branch keeps them running exactly as before.

Verified: the file parses, and the guard/ref logic is symmetric across all three jobs.

**Not verified: this has never run, and it structurally can't be, from a pull request.** Like `pull_request_target`, `workflow_run` only takes effect once the workflow file is on the default branch — a PR here can't trigger its own `workflow_run` version. The real test is the next tagged release after this merges: the upload jobs should stay queued (visible as "waiting" in the Actions tab, not silently absent) until "Release AMD64 artifacts" finishes, then run automatically.

---

### DIFF
--- a/.github/workflows/upload-cloud-images.yaml
+++ b/.github/workflows/upload-cloud-images.yaml
@@ -1,8 +1,23 @@
 name: Upload cloud images
+
+# This used to trigger on the same `push: tags: v*` event as "Release AMD64
+# artifacts", independently. GitHub Actions has no cross-workflow `needs:`, so
+# nothing sequenced them: this workflow queried quay.io/kairos/hadron for the
+# tag "Release AMD64 artifacts" publishes, and it ran first -- sometimes by
+# minutes. mauromorales/mission-control#174 has timestamps from v4.2.0-rc1
+# proving it: AWS and GCP both failed to resolve the tag almost three minutes
+# before the job that publishes it had even finished.
+#
+# workflow_run makes this a real dependency instead of a timing guess: this
+# workflow now only starts once "Release AMD64 artifacts" has actually
+# completed, and the `if: ... == 'success'` guard on every job below means a
+# failed or skipped release run (e.g. blocked by the vulnerability scan gate)
+# is not treated as "the tag might exist, try anyway".
 on:
-  push:
-    tags:
-      - 'v*'  # Triggers on any tag that starts with 'v'
+  workflow_run:
+    workflows: ["Release AMD64 artifacts"]
+    types:
+      - completed
   schedule:
     # Everyday at 2am
     - cron: '0 2 * * *'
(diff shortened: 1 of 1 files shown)


mauromorales (approved):

.github/workflows/upload-cloud-images.yaml
```diff
@@ -1,8 +1,23 @@
...
+# tag "Release AMD64 artifacts" publishes, and it ran first -- sometimes by
+# minutes. mauromorales/mission-control#174 has timestamps from v4.2.0-rc1
+# proving it: AWS and GCP both failed to resolve the tag almost three minutes
+# before the job that publishes it had even finished.
+#
+# workflow_run makes this a real dependency instead of a timing guess: this
+# workflow now only starts once "Release AMD64 artifacts" has actually
+# completed, and the `if: ... == 'success'` guard on every job below means a
+# failed or skipped release run (e.g. blocked by the vulnerability scan gate)
+# is not treated as "the tag might exist, try anyway".
```
jimmykarily, line 15: This explanation belongs to the commit message, not the code

jimmykarily (approved):

.github/workflows/upload-cloud-images.yaml
mauro-agent: Fair, moved it. #4333 was already merged before I saw this, so it's a follow-up: kairos-io/kairos#4334.

PR #4327, opened 2026-08-18
ci: scope image-pr-arm.yaml and uki.yaml off the production quay credential

mauro-agent:
## What

`image-pr-arm.yaml` and `uki.yaml` log in to quay with `QUAY_USERNAME`/`QUAY_PASSWORD` — the credential `release.yaml` uses to publish permanent images. Both workflows only ever push to `quay.io/kairos/ci-temp-images`, a scratch repository whose images expire after 6h.

This switches both to `QUAY_PR_USERNAME`/`QUAY_PR_PASSWORD`, the `kairos+prci` robot account scoped to write on `ci-temp-images` and nothing else.

## Why

`image-pr.yaml` already moved to the scoped credential as part of #4323's fork-pr-builds work. These two never got the same treatment, so every pull request run — including one from a fork with maintainer approval, once #4323 lands — hands the production publishing credential to a build that only needs scratch-repo access.

## Scope

Two files, credential names only. No trigger changes, no other behavior changes.

---

AI assisted this pull request. A human is attending and directed it.

Issue #4310, opened 2026-08-13
CI: bundles test pushes to a fixed tag, so concurrent PRs overwrite each other


mauro-agent:
## The problem

`reusable-qemu-test.yaml` pushes its bundles test image to a **fixed tag**:

```yaml
      - name: Build and push bundle image
        with:
          push: true
          tags: quay.io/kairos/ci-temp-images:bundles-test
```

No PR number, no commit SHA, no matrix dimension. Every run of the `bundles` test in every pull request writes the same tag.

## Why it matters

Two pull requests running the `bundles` test concurrently overwrite each other's image, and either may pull the other's. The test then validates a bundle built from **someone else's branch** while reporting against yours.

The failure mode is nasty because it is silent and intermittent: no error, no warning, just an occasional green that means nothing or a red that has nothing to do with the change under review. It looks exactly like flake, so it gets retried rather than investigated.

Every other image on this path is already disambiguated — the factory tags with `$FLAVOR-$FLAVOR_RELEASE-$VARIANT-$ARCH-$MODEL-$COMMIT_SHA`. This one is the exception.

## Suggested fix

Include the run's identity in the tag, in keeping with the surrounding convention. Any of:

```yaml
tags: quay.io/kairos/ci-temp-images:bundles-test-${{ github.event.pull_request.number }}
tags: quay.io/kairos/ci-temp-images:bundles-test-${{ github.sha }}
tags: quay.io/kairos/ci-temp-images:bundles-test-${{ github.run_id }}
```

The SHA form matches what the factory already does. Whichever is chosen, the consuming step in the same job needs the same reference.

Worth pairing with a `quay.expires-after` label so the extra tags do not accumulate — `image-pr.yaml` already applies `quay.expires-after=6h` to its images, and this one currently carries none.

## Not a security issue

Both writer and reader are authenticated pushes to a repository only the org can write to, so this is not a poisoning vector. It is a correctness and debuggability problem.

Found while reviewing [#4309](https://github.com/kairos-io/kairos/pull/4309).

---

<sub>🤖 Filed by `mauro-agent`, an AI agent operated by @mauromorales, with his review.</sub>

riccardomenegazzo:
I'd like to work on this.

[...]

PR #4391, opened 2026-08-26
ci: pull request jobs push to ci-temp-images with the scoped robot

mauromorales:
## What

Every pull request job that writes to `quay.io/kairos/ci-temp-images` authenticates today as `QUAY_USERNAME` / `QUAY_PASSWORD`, the account `release.yaml` uses to publish permanent images. This puts those writes back on `QUAY_PR_USERNAME` / `QUAY_PR_PASSWORD`, the `kairos+prci` robot scoped to `ci-temp-images` and nothing else.

## Why

#4323 and #4327 established exactly this, on 18 August. `ae314796` ("ci: retire pre-monorepo workflows, ship canonical pipeline") removed the three workflow files that carried the scoping, and the replacement pipeline reaches quay through `secrets: inherit`, so the production credential came back with it. There are zero references to `QUAY_PR_*` on `master` today. Both secrets still exist on the repository, so this works without any secret being created.

It reads as collateral from the pipeline migration rather than a decision. The commit subject is about retiring old workflows and says nothing about credentials.

## Three paths reach quay from `pr.yaml`

| Job | Reusable | What it pushes |
|---|---|---|
| `build-iso-arm`, the two `nvidia-jetson` cells | `_build-iso.yaml` | raw images, `registry_repository: ci-temp-images` |
| `qemu-tests-core`, the `bundles` cell | `reusable-qemu-test.yaml` | `ci-temp-images:bundles-test` |
| `qemu-tests-uki`, the `install-upgrade` cell | `_uki-test.yaml` | the freshly built UKI upgrade image |

Each one is switched on by an `if:` that matches exactly what `pr.yaml` sends, so all three run on an ordinary pull request.

## How

`reusable-qemu-test.yaml` already declared `QUAY_USERNAME` and `QUAY_PASSWORD`. `_build-iso.yaml` and `_uki-test.yaml` now declare them too, rather than reading them through `secrets: inherit`, so the caller chooses which account to hand over.

**`secrets: inherit` still satisfies a declared secret**, so `master.yaml` and `release.yaml` keep working with no change.

`pr.yaml` then passes the scoped robot to the six jobs that can reach quay: `build-iso-arm`, `qemu-tests-core`, `qemu-tests-core-fips`, `qemu-tests-standard`, `qemu-tests-standard-encryption`, and `qemu-tests-uki`.

## What this does not change

- **The `authorize` gate is untouched.** A fork pull request still goes through the `fork-pr-builds` environment with maintainer review.
- **`quay.expires-after=6h` is untouched.**
- **The remaining `secrets: inherit` callers in `pr.yaml`** are the jobs that never talk to quay: unit tests, lint, the kairos and kairos-init builds, the kcrypt-challenger builds, and the ghcr-only `build-iso`. They still receive the production credential and never use it. Removing that too is a bigger change and a separate decision. Say the word and I will extend this.

## Worth a reviewer's eye

`reusable-qemu-test.yaml` uses this credential to pull as well as push. Its own description says WRITE is required because it pulls the image under test from `ci-temp-images` and pushes the bundles image back. The scoped robot has write on that repository, so both work. If any test also pulls from a **different** private quay repository, that pull would start failing, and I could not find one by reading. Please confirm.

## Testing

`pr.yaml` runs on `pull_request_target`, so GitHub uses the base branch's copy. This pull request cannot exercise its own change. It needs a merge, or a manual run on a branch, to be validated for real.

YAML parses on all three files, and the `secrets` blocks are structurally under `on.workflow_call` in both reusables.

## Related

- #4310 is still open and untouched here: the bundles image is pushed to the fixed tag `ci-temp-images:bundles-test` with no pull request number or SHA, so concurrent pull requests overwrite each other and `tests/assets/bundles.yaml` pulls that exact tag.

### DIFF
--- a/.github/workflows/_build-iso.yaml
+++ b/.github/workflows/_build-iso.yaml
@@ -150,11 +150,26 @@ on:
         required: false
         default: false
 
-# No `secrets:` block on workflow_call: registry credentials are
-# derived from inputs.registry_domain inside the job (ghcr.io ->
-# github.actor + GITHUB_TOKEN, quay.io -> QUAY_USERNAME + QUAY_PASSWORD),
-# which only works when the caller uses `secrets: inherit` so this
-# reusable can see the org-level QUAY_* secrets.
+    # Registry credentials are derived from inputs.registry_domain
+    # inside the job: ghcr.io -> github.actor + GITHUB_TOKEN,
+    # quay.io -> these two.
+    #
+    # These are DECLARED rather than left to `secrets: inherit` so a
+    # caller can decide which quay account to hand over. `inherit`
+    # still satisfies a declared secret, so master.yaml and
+    # release.yaml keep working unchanged, while pr.yaml passes the
+    # scoped ci-temp-images robot instead of the production account.
+    secrets:
+      QUAY_USERNAME:
+        description: |
+          Registry user, used only when inputs.registry_domain is
+          quay.io. On the pull request path this must be the
+          ci-temp-images-scoped robot, never the account that
+          publishes permanent images.
+        required: false
+      QUAY_PASSWORD:
+        description: "Registry token matching QUAY_USERNAME."
+        required: false
 
 permissions:
   # packages: write is required for factory-action's push to ghcr:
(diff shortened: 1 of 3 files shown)


jimmykarily:
@mauromorales with the monorepo refactoring, I switch PRs to use ghcr instead of quay. The only pipeline that pushes on quay.io now should be the release one.

PR #4398, opened 2026-08-26
:art: immucore: replace containerd mount helper

### LINKED ISSUE #1697: [immucore] Drop mount libraries and use the mount syscall directly
We currently use a lib for mounting stuff on immucore but we should really not use something like that when we have the mount syscall and can use that directly. 

If we want more flexibility around mounting, we could create our own wrappers around that in the sdk, but for the basic usage done in immucore it doesnt make sense.

We currently use https://github.com/containerd/containerd JUST for the mounting which is excessive.

mudler-agent:
## What changed

- add a focused immucore mount type backed by `unix.Mount`
- parse standard mount flags and preserve filesystem options in order
- keep syscall errors compatible with `errors.Is`
- migrate immucore mount operations and fstab conversion
- remove the unused containerd v1 dependency

## Why

Immucore uses only a small part of containerd mount behavior. A local package keeps that required behavior without pulling in containerd v1 and its unrelated helpers.

## Verification

- `go test -count=1 ./immucore/...`
- `go mod tidy` followed by a clean `git diff --exit-code go.mod go.sum`
- confirmed no `github.com/containerd/containerd/mount` imports remain under `immucore`

Repository-wide tests require generated embedded binaries, rsync, Docker, and CGO TPM support that are unavailable in this runner. The changed immucore suite passes independently.

Closes #1697

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/charmbracelet/bubbles v1.0.0
 	github.com/charmbracelet/bubbletea v1.3.10
 	github.com/charmbracelet/lipgloss v1.1.0
-	github.com/containerd/containerd v1.7.34
 	github.com/containerd/containerd/v2 v2.3.4
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/deniswernert/go-fstab v0.0.0-20141204152952-eb4090f26517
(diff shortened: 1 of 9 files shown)


mudler-agent:
CI is waiting for approval of the protected `fork-pr-builds` environment. The bot account cannot approve that deployment, so the pull request test jobs were skipped. A maintainer needs to approve the pending environment deployment to start the test matrix; there is no code failure to fix yet.

mudler (approved):
looks good, but needs QA

mudler-agent:
QA update: the focused mount refactor tests pass locally on `5ce2f7a` (`go test ./immucore/internal/mount ./immucore/internal/utils ./immucore/pkg/op`). I could not complete the required real-boot QA on this cron host because Docker is not installed and `/dev/kvm` is unavailable. The fork PR workflow is also still waiting for approval of the protected `fork-pr-builds` environment. A maintainer needs to approve that deployment so the QEMU matrix can provide the boot evidence requested in the review.

mudler-agent:
Closing: #4458 already landed the same refactor on master, and it also carries the read-only bind remount this branch is missing. Verified master has no `containerd/containerd/mount` import left in immucore, so there is nothing here to rebase.

PR #4400, opened 2026-08-27
🐛 kairos-init: order immucore before switch-root

### LINKED ISSUE #4399: immucore.service SIGTERMed mid-run by Conflicts=initrd-switch-root.target, silently leaving a booted system with no mounts
**Kairos version:**
`/etc/os-release`:

```
NAME="SLE Micro"
VERSION="5.5"
VERSION_ID="5.5"
PRETTY_NAME="SUSE Linux Enterprise Micro for Rancher 5.5"
ID="sle-micro-rancher"
ID_LIKE="suse"
CPE_NAME="cpe:/o:suse:sle-micro-rancher:5.5"
IMAGE="registry.suse.com/suse/sle-micro/5.5:2.0.4-5.8.88"
GRUB_ENTRY_NAME="SLE Micro"
```

`/etc/kairos-release`:

```
KAIROS_ARCH="amd64"
KAIROS_FAMILY="suse"
KAIROS_FIPS="false"
KAIROS_FLAVOR="sle-micro-rancher"
KAIROS_FLAVOR_RELEASE="5.5"
KAIROS_ID="kairos"
KAIROS_ID_LIKE="kairos-core-sle-micro-rancher-5.5"
KAIROS_INIT_VERSION="v0.17.1"
KAIROS_MODEL="generic"
KAIROS_NAME="kairos-core-sles-5.5"
KAIROS_RELEASE="v4.1.2"
KAIROS_TRUSTED_BOOT="false"
KAIROS_VARIANT="core"
KAIROS_VERSION="v4.1.2"
KAIROS_IMAGE_LABEL="5.5-standard-amd64-generic"
```

immucore: `v0.20.3`, commit `a3fa1ebd3f931d7c8e715d49d6e1b96f699cbce0`, built with `go1.26.6`.

```
Linux <redacted> 5.14.21-150500.55.88-default #1 SMP PREEMPT_DYNAMIC Fri Dec 6 16:43:20 UTC 2024 (e102a89) x86_64 x86_64 x86_64 GNU/Linux
```

QEMU/KVM guest, UEFI, virtio disk. Normal active boot (`/run/cos/active_mode`).

**Describe the bug**

 On one boot, `immucore.service` was terminated by SIGTERM roughly 10 seconds into a run
that normally takes ~26 seconds. It died before `mount-bind` completed, so none of the
persistent state binds, the `/etc` overlay, or `/usr/local` were ever established.

The system then **booted successfully anyway**. It was reachable over SSH, systemd came
up, services started — but the entire rootfs layout was missing. `/` remained the
read-only `active.img` loop, and every write to `/etc` failed silently. Nothing written
during that boot survived.

The silent degradation is the part we would most like to see addressed. A host in this
state looks healthy and is very hard to diagnose. We arrived here from a completely
unrelated-seeming symptom (a missing systemd-networkd config file) and it took a long
investigation to establish that the real problem was that nothing had been mounted.

We have seen this once. A reboot of the same host, on the same image and with an
identical kernel cmdline, completed normally — so the failure appears to be
timing-dependent rather than deterministic, but we cannot yet say how often it occurs.

*Component note:* the defect appears to be in `immucore`. Filing here because
`KAIROS_BUG_REPORT_URL` in `/etc/kairos-release` points to this repo — happy for it to
be transferred to `kairos-io/immucore` if that is preferred.
**To Reproduce**

We do not have a reliable reproduction. Observed once on a QEMU/KVM guest with `rd.debug` and a serial console on the cmdline. A subsequent reboot with an identical cmdline completed all ~26 s normally and produced a fully healthy system.

Kernel cmdline on both the failing and succeeding boots:

```
BOOT_IMAGE=(loop0)/boot/vmlinuz console=ttyS0 console=tty1 net.ifnames=1 rd.cos.oemlabel=COS_OEM
rd.cos.oemtimeout=10 panic=5 rd.emergency=reboot rd.shell=0 systemd.crash_reboot=yes
systemd.unified_cgroup_hierarchy=1 rd.driver.blacklist=nouveau modprobe.blacklist=nouveau
nouveau.modeset=0 root=LABEL=COS_ACTIVE cos-img/filename=/cOS/active.img selinux=0
rd.immucore.debug console=tty0 rd.debug
```

All COS partitions were present, healthy, and read-write throughout (`lsblk -o NAME,RO` showed `RO=0` for `COS_GRUB`, `COS_OEM`, `COS_RECOVERY`, `COS_STATE`, `COS_PERSISTENT`). This was not a disk or partition problem.

**Expected behavior**

If immucore's mount DAG does not complete, the boot should fail loudly rather than proceed into a system with no mounts. immucore already has the machinery for this — `RenderFailureScreen` / `HaltWithBanner` in `internal/utils/bootfailure.go` render a full-screen `KAIROS BOOT FAILED` banner for operator-fixable conditions. A terminated mount DAG seems like exactly that class of condition.

Secondarily, a ~26-second workload should not be terminable at an arbitrary point by an unordered conflict relationship.

**Logs**
Timeline of the failing boot:

```
12:27:59Z  immucore starts
12:28:02Z  runs the `rootfs` yip stage; log ends mid-stage
12:28:09Z  SIGTERM
```

```
$ systemctl status immucore
× immucore.service
     Loaded: not-found (Reason: Unit immucore.service not found.)
     Active: failed (Result: signal) since Wed 2026-08-26 12:28:09 UTC
   Main PID: 1435 (code=killed, signal=TERM)
        CPU: 222ms
```

`/run/immucore/immucore.log` was truncated at 15,714 bytes, ending part-way through processing the `Rootfs Layout Settings.Layout configuration for active/passive mode` stage. No `boot-timeline.json` was produced. Seven seconds of complete silence preceded the SIGTERM, with both `rd.debug` and `rd.immucore.debug` enabled.

Resulting mount table — only kernel pseudo-filesystems, three tmpfs, and the read-only loop root:

```
/dev/loop0 on / type ext2 (ro,relatime)
tmpfs on /etc/machine-id type tmpfs (ro,...)
```

Confirmed against `/proc/self/mountinfo`, so not a mount-namespace artifact. Absent:

- `/run/initramfs/cos-state` (COS_STATE)
- `/oem` (COS_OEM) — note this *had* mounted successfully inside the initramfs, per the log
- `/usr/local` (COS_PERSISTENT)
- overlays on `/etc`, `/var`, `/srv`
- every `PERSISTENT_STATE_PATHS` bind, including `/etc/systemd`

Downstream symptoms, each individually misleading, all consequences of a read-only `/etc`:

- yip's `05_network.yaml` passed its `only_os` / `only_service_manager` gates and ran, but could not write `/etc/systemd/network/20-dhcp.network` — the host had no DHCP configuration
- `/etc/machine-id` was a **read-only tmpfs mount** (systemd's transient fallback for a read-only `/etc`)
- hostname never set — stuck at `localhost`
- `/run/cos/extra-layout.env` never created
- anything expected under `/oem` appeared missing, because COS_OEM was not mounted

**Additional context**
`immucore.service`, extracted verbatim from the initrd:

```ini
[Unit]
Description=immucore
DefaultDependencies=no
After=systemd-udev-settle.service
Requires=systemd-udev-settle.service
Before=initrd-fs.target
Conflicts=initrd-switch-root.target

[Service]
Type=oneshot
RemainAfterExit=yes
StandardOutput=journal+console
ExecStart=/usr/bin/immucore
```

`Conflicts=initrd-switch-root.target` means activating switch-root **stops** immucore, and stopping a running `Type=oneshot` service sends SIGTERM — matching the observed `code=killed, signal=TERM`.

Per systemd's documentation, `Conflicts=` is *"independent of and orthogonal to the `After=` and `Before=` ordering dependencies"*, so the stop job for `immucore.service` and the start job for `initrd-switch-root.target` are unordered. The only guard is `Before=initrd-fs.target`, and on this boot it did not prevent the transition. There is no `TimeoutStartSec=`, so this is not a unit timeout — it is a conflict-driven stop.

**How long immucore actually needs.** From `/run/immucore/boot-timeline.json` on a *successful* boot of the same host and image:

| Step | Duration (ms) |
|---|---|
| **mount-bind** | **13,279** — with `error: mkdir /sysroot/etc/ssl/certs: file exists` |
| initramfs-hook | 6,929 |
| mount-state | 1,221 |
| mount-oem | 1,156 |
| custom-mount | 1,145 |
| mount-root | 1,066 |
| mount-tmpfs | 1,002 |
| rootfs-hook | 474 |
| discover-state | 138 |
| enable-sysext-confext | 93 |
| lvm-activation | 87 |
| overlay-mount | 47 |
| create-sentinel | 43 |
| write-fstab | 9 |
| mount-base-overlay | 6 |
| unlock-all | 5 |
| upgrade-kcrypt | 4 |
| load-config | 0.3 |

**Total ≈ 26 s**, over half of it in `mount-bind`. That successful run produced 138,910 bytes / 1,881 log lines between `21:21:04Z` and `21:21:30Z`.

The failing boot was killed at 10 s — right around where `mount-bind` would begin. That is consistent with `/oem` having mounted (~4 s cumulative) while no bind or overlay ever did.

**What we could not determine.** We do not know what activated `initrd-switch-root.target` at the 10-second mark. The evidence is gone: `/run/immucore/immucore.log` lives on tmpfs, and that boot's journal was volatile because `/var/log` had no persistent bind at the time (`journalctl --list-boots` now shows only the current boot).

Two contributing factors we considered but could not confirm:

- `StandardOutput=journal+console` with `console=ttyS0` means all 1,881 log lines are written synchronously to a serial console under `rd.debug`. We initially suspected this was the primary cause; the timeline data suggests genuine mount work dominates, so at most it widens the window.
- The `mkdir /sysroot/etc/ssl/certs: file exists` error inside `mount-bind`, which is recorded even on successful boots.

**Possible remedies**, roughly in the order we would value them:

1. Fail loudly instead of degrading silently — trip the existing `HaltWithBanner` path when the mount DAG does not complete.
2. Give the switch-root transition a hard ordering guarantee, rather than relying on `Before=initrd-fs.target` alongside an unordered `Conflicts=`.
3. Consider a `TimeoutStartSec=` with an explicit failure action, so a stalled immucore produces a defined, diagnosable outcome.
4. Investigate the 13.3 s `mount-bind`, including the recurring `/etc/ssl/certs` error — reducing that step shrinks the window this race depends on.
5. Persist the diagnostic artifacts. `/run/immucore/immucore.log` and `boot-timeline.json` are the only record of what happened, and both are lost on the reboot a confused operator will immediately perform. This overlaps with #3381 (persistent boot log collection), which would have made this report substantially easier to produce.

mudler-agent:

Adds a direct `Before=initrd-switch-root.target` ordering edge to the bundled `immucore.service`. The existing `Conflicts=` relationship could otherwise stop the running oneshot before its mount DAG completed, while the system continued into a root filesystem without persistent binds or overlays.

A regression test preserves both the direct ordering edge and the conflict relationship.

**Verification**:

- `go test ./kairos-init/...`
- `git diff --check`

The repository-wide `make test` reached unrelated environment-dependent failures because this worker has no `rsync`, CGO TPM simulator support, or Docker socket.

Fixes #4399

### DIFF
--- a/kairos-init/pkg/bundled/bundled.go
+++ b/kairos-init/pkg/bundled/bundled.go
@@ -180,6 +180,7 @@ DefaultDependencies=no
 After=systemd-udev-settle.service
 Requires=systemd-udev-settle.service
 Before=initrd-fs.target
+Before=initrd-switch-root.target
 Conflicts=initrd-switch-root.target
 
 [Service]
--- a/kairos-init/pkg/bundled/immucore_service_test.go
+++ b/kairos-init/pkg/bundled/immucore_service_test.go
@@ -1,0 +1,18 @@
+package bundled_test
+
+import (
+	"strings"
+
+	"github.com/kairos-io/kairos/v4/kairos-init/pkg/bundled"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Immucore dracut service", func() {
+	It("finishes before switch-root can stop it", func() {
+		lines := strings.Split(bundled.ImmucoreServiceDracut, "\n")
+
+		Expect(lines).To(ContainElement("Before=initrd-switch-root.target"))
+		Expect(lines).To(ContainElement("Conflicts=initrd-switch-root.target"))
+	})
+})


mudler-agent:
CI is waiting for approval of the protected `fork-pr-builds` environment. The bot account cannot approve that deployment, so the test matrix is currently skipped. A maintainer needs to approve the pending environment deployment; there is no code failure to fix yet.

PR #4401, opened 2026-08-27
Fix/flaky tests registry and reboot connect

jimmykarily:


mauromorales (dismissed):

immucore/pkg/state/steps_shared.go
```diff
@@ -347,12 +347,19 @@ func (s *State) MountCustomMountsDagStep(g *herd.Graph, opts ...herd.OpOption) e
...
+				// it fails with "no such device" and boot continues without
+				// /usr/local mounted - so every persistent bind (/home, ...)
+				// silently disappears from the running system.
 				fstab, err2 := op.MountOPWithFstab(
 					what,
 					s.path(where),
 					fstype,
 					mountOptions,
-					3*time.Second,
+					30*time.Second,
```
Itxaka, line 362: Increasing this  because of a test failure while we have no reports of it being a problem in the real world, could have consequences for real world deployments.

Consider if is the tests the one that needs change to no affect real world deployments.

jimmykarily: The tests is probably more limited in resources than most production deployments. Also, this is a timeout. Anything that booted fine with the old 3 secs, should still get past this in less than 3 seconds. The bump should only be "visible" in cases where 3 seconds wasn't enough, no? For everybody else, this should change nothing.

jimmykarily:
2 tests are failing. One is the provider-decentralized test which needs investigation, I'm no longer sure it just flakes, maybe it's a legit failure. I will investigate and fix on a separate PR.
The other was tracked down and the possible cause is reported here: https://github.com/kairos-io/kairos/issues/4403

Will defer the fix to a separate PR too.

immucore/pkg/state/steps_shared.go
jimmykarily: The last run did the mounting in 1 second:

```
07:40:20  Triggering udev events
  07:40:20  Waiting for udev to settle (timeout=30000)
  07:40:20  udevadm settle completed successfully
  07:40:20  Partition unlocked and verified   partition=COS_PERSISTENT
  07:40:21  mount done options=[rw] type=ext4
            what=/dev/disk/by-label/COS_PERSISTENT where=/sysroot/usr/local
```
and later failed due to a different bug (https://github.com/kairos-io/kairos/issues/4403).
This proves that 30 seconds is the "max" waiting, not a price everybody pays.

PR #4404, opened 2026-08-27
Fix encrypted-partition boot flake (#4403)

### LINKED ISSUE #4403: Encrypted persistent boot flakes: LUKS container and inner ext4 share LABEL=COS_PERSISTENT
## Symptom

On a Kairos guest with an encrypted persistent partition, boot sometimes lands in a broken state where `/usr/local` is not mounted and `/home/kairos` does not exist, even though the LUKS unlock itself succeeded. Every SSH session then prints

```
Could not chdir to home directory /home/kairos: No such file or directory
```

before running any command, and the journal loops on this every 5 seconds:

```
mount: /usr/local: fsconfig() failed: /dev/vda5: Can't open blockdev.
```

until whatever is watching times out.

## Cause

Two block devices end up carrying the same filesystem label at runtime:

- `/dev/vda5` — the LUKS2 container itself, tagged `LABEL=COS_PERSISTENT` because kairos-sdk passes `--label COS_PERSISTENT` to `cryptsetup luksFormat`.
- `/dev/mapper/vda5` — the ext4 inside the container, also labelled `COS_PERSISTENT`.

Both are visible to udev, so `/dev/disk/by-label/COS_PERSISTENT` is ambiguous. In the initrd it happens to resolve to the mapper (immucore triggers udev right after unlocking, so the plaintext side wins). Once systemd takes over after switch-root, the same lookup resolves to the raw LUKS container `/dev/vda5`, and `mount` cannot mount an encrypted blockdev as ext4 — hence the `Can't open blockdev` loop.

The `--label` on `luksFormat` was added in kairos-sdk commit [`76af8a6`](https://github.com/kairos-io/kairos-sdk/commit/76af8a6). The convention was later assumed by [kairos-io/kairos-sdk#822](https://github.com/kairos-io/kairos-sdk/pull/822) (which fixed the *opposite* problem of legacy disks having no label at all — kairos-io/kairos#4276).

The exact same class of bug (two devices sharing a label, wrong one picked) was fixed for the grub partition in [kairos-io/kairos#3611](https://github.com/kairos-io/kairos/issues/3611) by renaming one side to `COS_GRUB_LIVE`.

## Tests affected

Reproduces intermittently in the standard-encryption QEMU tests, most visibly the two currently-flaky specs:

- `encryption-remote-ek-reenroll`
- `encryption-remote-complete-workflow`

See run [33048732117 / job 98441832565](https://github.com/kairos-io/kairos/actions/runs/33048732117/job/98441832565) for a captured failure — `blkid`, `mounts`, and the `Can't open blockdev` loop are all in the log artifact.

## Suggested direction

Rename the outer LUKS label so it doesn't collide with the inner ext4 (e.g. `COS_PERSISTENT_LUKS` / `COS_OEM_LUKS`), following the `COS_GRUB_LIVE` precedent. Touch points in kairos-sdk:

- `kcrypt/lock.go` — the two `--label` args in `Encrypt` / `luksifyMeasurements`.
- `kcrypt/encryptor.go` — `encryptedPartitionLabel()` mapping added by #822 needs to learn the new outer name.
- `kcrypt/unlock.go` — label lookup during unlock.

Already-installed hosts can be re-labelled safely at upgrade time with `cryptsetup config <dev> --label ...` (metadata-only, doesn't touch keys or the passphrase).

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/4403

## The problem

On Kairos hosts with encryption enabled, boot sometimes lands in a broken state: `/usr/local` isn't mounted, `/home/kairos` doesn't exist, and every SSH login prints `Could not chdir to home directory /home/kairos`. In CI the standard-encryption tests flake at roughly 50%.

## Why

We put the same filesystem label, `COS_PERSISTENT`, on both the LUKS container and the ext4 inside it. udev then has to pick one to point `/dev/disk/by-label/COS_PERSISTENT` at, and it's a coin flip which one wins. When it points at the LUKS container, systemd tries to mount an encrypted block device as ext4 and loops on "Can't open blockdev".

Real users hit this too, but usually less visibly: encryption is opt-in, real hardware has more timing jitter than QEMU, and systemd's mount retries often eventually succeed, so it shows up as "slow boot" rather than a broken one. CI cycles installs and reboots often enough that we see it clearly.

## The fix

Two independent changes so we're safe against future by-label races too:

1. **New installs use different labels for the two sides**. The LUKS container gets `<name>_LUKS`, the inner ext4 keeps the plain name. The by-label symlink is unambiguous from install day one.
2. **Boot-time mounts skip the by-label symlink entirely**. Immucore now resolves each label to a concrete device path (`/dev/mapper/<name>` for encrypted partitions, `/dev/<name>` otherwise) before mounting, and writes that same concrete path into `/etc/fstab`. Applies to every install shape, including pre-fix ones on upgrade and installs with custom encrypted-partition labels.

The agent-side mount helper (`sdk/machine.Mount`, used by post-install and reset hooks) goes through the same lookup, so no by-label lookups against encrypted labels remain in the tree.

## kcrypt-challenger wire label

For remote-KMS installs, the label sent to kcrypt-challenger is always the label that ends up on the LUKS header. That means:

- Pre-fix installs continue to send `COS_PERSISTENT` / `COS_OEM`, and their existing SealedVolume secrets keep working as-is.
- New installs (post-fix) send `COS_PERSISTENT_LUKS` / `COS_OEM_LUKS`, and TOFU stores their secrets under those names.

Encrypt-side and unlock-side always agree because the SDK explicitly sends the label it is about to write on the LUKS at encrypt-time, and reads that same label back at unlock-time. No translation layer, no special case for custom labels.

Admins looking at SealedVolumes on a mixed fleet will see both naming variants; the `_LUKS` suffix is a useful marker for which installs went through this fix.

## Upgrade path

Existing encrypted installs are fixed on their next boot after upgrading. No on-disk migration is needed: the LUKS metadata stays as it was, the wire label the SDK sends matches what's on the header, and the challenger secret keeps matching. Immucore rewrites `/etc/fstab` on every boot, so the concrete-path mount fix takes effect on the first boot with this build regardless of when the install happened.

## Test coverage

- Unit tests for the type-filtered lookups and mount-source resolution in `sdk/kcrypt/lookup`.
- The immucore state test asserts an encrypted install's fstab entry Spec resolves to `/dev/mapper/...`, never the by-label form.
- Every encryption e2e cell (`encryption-local`, `encryption-remote-*`, `encryption-kcrypt-cli`) now runs `cat /etc/fstab` on the guest post-reboot and fails loudly if the persistent mount regresses back to `LABEL=`. Includes the two currently-flaky specs (`encryption-remote-complete-workflow` and `encryption-remote-ek-reenroll`); a green run on this branch is the acceptance test.
- The UKI boot-assessment spec's df/mount assertions were updated to match the new concrete-path form.

## Follow-ups (separate PRs)

- Add `upgrade-latest-with-cli-encrypted` e2e cell to permanently exercise "install from previous release with encryption, upgrade, still unlocks and boots".


jimmykarily:
Here is an alternative smaller fix: https://github.com/kairos-io/kairos/pull/4407 . It leaves the label collision and just tries to workaround it. This PR here is trying to remove the collision altogether.

sdk/kcrypt/lookup/lookup.go
```diff
@@ -0,0 +1,219 @@
...
+	return findByLabelFiltered(partitionLabel, nil)
+}
+
+// FindLUKSContainerByLabel returns the crypto_LUKS partition carrying the
+// given filesystem label. Used at unlock time to unambiguously locate the
+// container even when the outer LUKS and its inner ext4 share the same label
+// (pre-fix installs, kairos-io/kairos#4403).
+func FindLUKSContainerByLabel(partitionLabel string) (*partitions.Partition, error) {
+	return findByLabelFiltered(partitionLabel, func(p *partitions.Partition) bool {
+		return p.FS == "crypto_LUKS"
```
mauromorales: worth making "crypto_LUKS" a constant I'd say, but not blocking

jimmykarily: good point, fixed

sdk/kcrypt/lookup/lookup.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if err != nil || devicePath == "" {
+		return nil, fmt.Errorf("partition not found")
+	}
+
+	return &partitions.Partition{
+		Path: devicePath,
+		Name: filepath.Base(devicePath),
+	}, nil
+}
```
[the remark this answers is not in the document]
jimmykarily: fixed by calling the binary directly (not shelling out)

PR #4407, opened 2026-08-28
Fix encrypted-partition boot flake, alternative to #4404 (#4403)

jimmykarily:
Alternative to #4404. Same bug, smaller diff.

## The bug

On encrypted installs, boot sometimes lands broken: `/usr/local` is not mounted, `/home/kairos` is missing, and every SSH login prints `Could not chdir to home directory /home/kairos`. Root cause is that both the LUKS container and the ext4 inside it carry the same filesystem label. `/dev/disk/by-label/COS_PERSISTENT` then races on which one udev picked, and when it lands on the LUKS side, systemd tries to mount an encrypted block device as ext4 and loops on "Can't open blockdev".

## The fix (this PR)

Stop trusting `/dev/disk/by-label` anywhere in the mount path. Immucore now resolves each label to a concrete device (`/dev/mapper/<name>` for encrypted, `/dev/<name>` otherwise) before mounting, and writes that same concrete path into `/etc/fstab`. The udev race becomes irrelevant. Same treatment for `sdk/machine.Mount` (called by agent hooks) and the phonehome handler's `/oem` mount.

Existing encrypted installs are fixed on their next boot after upgrading. No on-disk migration, no wire-protocol change with kcrypt-challenger, no SealedVolume rename.

## How it compares to #4404

Both PRs fix the observed boot failure. This PR is the smaller of the two: it only teaches the mount code to bypass `by-label`. #4404 additionally changes the on-disk label of the LUKS container itself so that `/dev/disk/by-label/COS_PERSISTENT` becomes unambiguous, which requires updates in every place that scans for that label, a rename of the kcrypt-challenger secret naming for new installs, and matching test churn.

**Trade-off:** this PR does not split the label at the LUKS header. Nothing in the Kairos tree depends on `by-label` for encrypted partitions after this change, but a third-party tool that resolved `/dev/disk/by-label/COS_PERSISTENT` on an encrypted host would still race. #4404 closes that hole at the cost of the wider changes above.

## Which one to merge

Up to the maintainers. My take: this branch fixes the reported symptom with the smallest surface. #4404 additionally hardens the on-disk state for hypothetical future consumers, in exchange for a wider blast radius (test churn, wire-label change, all ghw scanners).

### DIFF
--- a/agent/internal/phonehome/handlers.go
+++ b/agent/internal/phonehome/handlers.go
@@ -16,6 +16,8 @@ import (
 
 	"github.com/kairos-io/kairos/v4/agent/pkg/action"
 	"github.com/kairos-io/kairos/v4/agent/pkg/constants"
+	sdkConstants "github.com/kairos-io/kairos/v4/sdk/constants"
+	"github.com/kairos-io/kairos/v4/sdk/machine"
 	sdkConfig "github.com/kairos-io/kairos/v4/sdk/types/config"
 )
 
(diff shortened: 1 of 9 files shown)


jimmykarily:
https://github.com/kairos-io/kairos/pull/4404 was merged. Closing this.

PR #4411, opened 2026-08-28
fix: let the p2p provider choose the edgevpn API address

jimmykarily:

On a Kairos node built with the `standard` variant, asking for the cluster's kubeconfig printed nothing at all:

```
# /system/providers/agent-provider-kairos get-kubeconfig

# echo $?
0
```

An empty line and a success exit code. Listing the node roles behaved the same way, printing its table header and no rows:

```
# /system/providers/agent-provider-kairos role list
Node                                             Role                            IP
-----------------------------------------------  ------------------------------  ---------------
```

The cluster itself was healthy throughout. Both nodes found each other, roles were assigned, the master published its address and join token, and the worker joined k3s.

### Cause

Kairos nodes run edgevpn, a peer to peer daemon that lets the nodes discover each other and share the kubeconfig and the k3s join token. It exposes a small HTTP API. Both the address the daemon listens on and the address its command line client dials are decided by the p2p provider, provider-kairos.

The provider defaults both ends to a unix socket at `/run/edgevpn-kairos.sock`.

`kairos-agent start` had a default of its own, `http://127.0.0.1:8080`, which it published on the plugin bus as `BootstrapPayload.APIAddress`. The provider prefers whatever the agent sends it, so the daemon was told to listen on TCP port 8080 while the provider's own client kept dialing the socket. Nothing listened on the socket. The client discards its error (`str, _ := cc.Get(...)`), so the commands printed empty output and exited 0.

Confirmed from the CI logs of the failing job. The edgevpn daemon reported:

```
edgevpn[1460]: ⇨ http server started on 127.0.0.1:8080
```

while the provider binary shipped in the same image reported:

```
--api value  ... (default: "unix:///run/edgevpn-kairos.sock")
```

### Fix

The agent has no business choosing this address. It now sends an empty string when the operator did not pass `--api`, which is the case the provider already handles by falling back to its own default. One component decides, so both ends agree.

Minimal pair, taken from the `kairos-agent` binary extracted out of each ISO:

```
--- ISO built from master (2026-08-24) ---
   --api value  (default: "http://127.0.0.1:8080")
--- ISO built from this branch ---
   --api value  Address of the edgevpn API the p2p provider co-ordinates over. ...
```

The qemu test `provider-decentralized-k8s` reached that API by curling port 8080 in three places, so those move to the socket.

This also adds `make iso`, which builds a bootable ISO from a checkout. The Makefile previously stopped at container images; the two steps that turn those into something you can boot existed only inside `.github/workflows/reusable-factory.yaml`. That is how the verification below was done.

### Verification

`provider-decentralized-k8s` has failed on every master run since 2026-08-24, always timing out at `tests/provider_decentralized_test.go:156` waiting 25 minutes for a kubeconfig that could never arrive. Most recently [run 33155499457](https://github.com/kairos-io/kairos/actions/runs/33155499457/job/98800113379).

Run locally against an ISO built from this branch with `make iso`:

| Step | CI, before | Local, after |
|---|---|---|
| checking if it has a working kubeconfig | 1500s, then failed | 1.4s |
| checking roles | 900s | 0.2s |
| checking if it has machines with different IPs | 900s | 0.1s |
| checking if it can propagate dns | not reached cleanly | passed |
| suite total | 45m53s, FAIL | 4m1s, `1 Passed \| 0 Failed` |

The DNS step passing matters on its own, because it exercises the rewritten `curl --unix-socket` POST path rather than just compiling it.

---

## Reminder: this needs a release note

**The edgevpn API on standard nodes moves from `127.0.0.1:8080` to a unix socket at `/run/edgevpn-kairos.sock`.** Suggested text:

> On a node built with the `standard` variant, the peer to peer coordination API that edgevpn serves used to listen on TCP port `8080`. It now listens on a unix socket instead. Anything that reached it over TCP needs to change:
>
> ```
> # before
> curl http://localhost:8080/api/machines
> # after
> curl --unix-socket /run/edgevpn-kairos.sock http://localhost/api/machines
> ```
>
> If you need the TCP listener back, both ends have to be told, not just one:
>
> ```
> kairos-agent start --api http://127.0.0.1:8080
> EDGEVPN_API=http://127.0.0.1:8080 /system/providers/agent-provider-kairos get-kubeconfig
> ```
>
> The flag moves the edgevpn daemon and the provider's in-process coordination client. It does not reach the provider's command line client, which has no way to learn it and keeps defaulting to the unix socket, so `get-kubeconfig` and `role list` need `EDGEVPN_API` set to the same address.
>
> This only affects the API running on a Kairos node. The `kairosctl bridge` API, which runs on your own machine and is the one the documentation describes, still listens on `127.0.0.1:8080` and is unchanged.

The TCP listener was never documented for nodes. Every `localhost:8080` in the docs refers to `kairosctl bridge`, which has its own separate `--api` flag and is unaffected: `docs/installation/p2p.md` lines 171, 177 and 299 all sit under a `kairosctl bridge` heading, and `docs/reference/kairosctl.md:154` is the `bridge` command reference. So no documentation change is required, only the release note.

## Follow up work, not in this PR

Both in `kairos-io/provider-kairos`:

1. Derive the client's `--api` default from the `APILISTEN` the provider itself writes to `/etc/systemd/system.conf.d/edgevpn-kairos.env`, so the two ends stay in agreement even when someone does pass `--api` to the agent. This PR removes the disagreement; that change would make it structurally impossible.
2. Stop discarding the client error in `internal/cli/get_kubeconfig.go` and `internal/cli/role.go`. That swallowed error is the reason a connection failure cost 45 minutes of CI time instead of printing one line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,6 +275,19 @@ image-kairos-init: kairos-init
 .PHONY: images
 images: image-kcrypt-challenger image-kairos-init
 
+# --- bootable artifacts ---
+#
+# Everything above stops at container images. `make iso` carries on to a
+# bootable ISO by running the two remaining steps CI does in
+# .github/workflows/reusable-factory.yaml: bake the OS image with
+# images/Dockerfile, then hand it to auroraboot. Defaults reproduce the
+# amd64-standard cell, which is what the provider and standard qemu suites
+# test against. Needs docker with a usable socket; see the script header for
+# the environment variables it honours.
+.PHONY: iso
+iso:
+	scripts/build-iso.sh
+
 .PHONY: clean
 clean:
 	rm -rf $(DIST_DIR)
--- a/agent/pkg/cmd/root.go
+++ b/agent/pkg/cmd/root.go
@@ -277,8 +277,28 @@ Starts the kairos agent which automatically bootstrap and advertize to the kairo
 				Name: "force",
 			},
 			&cli.StringFlag{
-				Name:  "api",
-				Value: "http://127.0.0.1:8080",
+				Name: "api",
+				Usage: "Address of the edgevpn API the p2p provider co-ordinates over. " +
+					"Leave unset to let the provider choose, which is what keeps the edgevpn " +
+					"daemon's listen address and the provider CLI's client address in agreement. " +
+					"Setting it moves the daemon only, so the provider CLI needs the same " +
+					"address in EDGEVPN_API to keep working.",
+				// Deliberately no default. The address is the p2p provider's to
+				// decide: it both writes APILISTEN for the edgevpn daemon and
+				// defaults its own CLI client to the same place. A default here
+				// would win over the provider's and desynchronise the two ends.
+				//
+				// Note that an explicitly set value desynchronises them just the
+				// same. Whatever arrives here reaches the daemon's APILISTEN and
+				// the provider's in-process coordination client, but not the
+				// provider's command line client, which has no way to learn it
+				// and keeps its own default. Operators who set this must also
+				// export EDGEVPN_API for `get-kubeconfig` and `role list`. The
+				// real fix is for the provider CLI to derive its default from
+				// the APILISTEN the provider already writes to
+				// /etc/systemd/system.conf.d/edgevpn-kairos.env, which belongs
+				// in kairos-io/provider-kairos.
+				Value: "",
 			},
 		},
 		Action: func(c *cli.Context) error {
--- a/agent/pkg/cmd/root_test.go
+++ b/agent/pkg/cmd/root_test.go
@@ -108,3 +108,39 @@ func TestCatalogDownloadFailureDoesNotLeaveTemporaryFile(t *testing.T) {
 		t.Fatalf("temporary catalog was not removed: %v", statErr)
 	}
 }
+
+// The edgevpn API address is chosen by the p2p provider, not by the agent.
+// The agent only forwards whatever the operator passed on the command line,
+// so an unset --api must reach the provider as an empty string. If the agent
+// substitutes a default of its own, that default silently overrides the
+// provider's and the two ends of the API disagree: the provider writes the
+// agent's address into the edgevpn daemon's APILISTEN while its own CLI keeps
+// defaulting to the provider's socket, so `get-kubeconfig` and `role list`
+// talk to an address nothing is listening on and print empty output.
+func TestStartAPIFlagHasNoDefault(t *testing.T) {
+	var start *cli.Command
+	for _, c := range cmds {
+		if c.Name == "start" {
+			start = c
+			break
+		}
+	}
+	if start == nil {
+		t.Fatal("no start command registered")
+	}
+
+	var api *cli.StringFlag
+	for _, f := range start.Flags {
+		if sf, ok := f.(*cli.StringFlag); ok && sf.Name == "api" {
+			api = sf
+			break
+		}
+	}
+	if api == nil {
+		t.Fatal("start command has no --api flag")
+	}
+
+	if api.Value != "" {
+		t.Fatalf("start --api default = %q, want %q so the provider picks the address", api.Value, "")
+	}
+}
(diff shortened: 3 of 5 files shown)


mauromorales:
@mauro-agent please review

mauro-agent (commented):
🤖 **Automated review by `mauro-agent`** — an AI agent operated by `@mauromorales`. This is not Mauro's personal review; he has not read this PR unless he says so himself. If you want a human, feel free to ping him directly.

Read the diff and the PR description in full. Ran nothing — no build, no tests, no dependency install.

**The fix looks correct.** Root cause matches the evidence: `kairos-agent start --api` defaulted to `http://127.0.0.1:8080` while the p2p provider's own default is a unix socket, so the daemon listened on TCP while the provider CLI kept dialing the socket, and the discarded client error (`str, _ := cc.Get(...)`) turned that into silent empty output instead of a visible failure. Removing the agent's own default and letting an unset flag reach the provider as an empty string is the right fix: one component decides the address, both ends agree.

The verification is concrete and checks out: the CI failure signature (`provider_decentralized_test.go:156`, 25-minute timeout waiting for a kubeconfig) matches the described symptom, and the local before/after table (45m53s FAIL vs 4m1s PASS) is the kind of evidence that actually demonstrates the fix rather than just asserting it.

The test changes are a real improvement beyond the fix itself: `provider_decentralized_test.go` used to discard curl's error (`_ = vm.Sudo(...)`), which is exactly the swallowed-error pattern the PR's own "follow up work" section flags as a problem elsewhere — good that it's fixed here where it's touched, even though the broader fix is scoped out to `provider-kairos`.

Two minor, non-blocking notes:
- `scripts/build-iso.sh` checks for `docker` and `go` on `PATH` but not `make`, even though it calls `make binaries` and `make image-kairos-init` first. Low severity — a missing `make` would still fail loudly, just with a plainer shell error instead of the script's own friendlier message.
- The PR template's "Which issue(s) this PR fixes" is empty. Not a problem if this was found during other work rather than filed as an issue first, just flagging in case a tracking issue exists and should be linked.

The release-note reminder in the description is worth keeping — the API address moving from TCP to a unix socket is a real behavior change for anyone reaching it directly.

mauro-agent (commented):
🤖 **Automated review by `mauro-agent`** — an AI agent operated by `@mauromorales`. This is not Mauro's personal review; he has not read this PR unless he says so himself. If you want a human, feel free to ping him directly.

Ran nothing (no build, no test, no dependency install) — this is a read-only review of the diff plus a cross-check of `provider-kairos` source on `origin/main`.

**The core fix is correct and well-verified.** `agent/pkg/cmd/root.go` drops the agent's own `--api` default (`http://127.0.0.1:8080`), sending an empty string when the operator doesn't set one. Checked this against `kairos-io/provider-kairos`'s actual source rather than taking the PR body's word for it: `internal/provider/bootstrap.go` already falls back to `DefaultEdgeVPNAPIAddress` (`unix:///run/edgevpn-kairos.sock`) exactly when `cfg.APIAddress == ""`, and `internal/cli/start.go`'s CLI client already defaults to the same constant. So removing the agent's competing default is what makes both ends agree, not an assumption.

**Test coverage matches the fix.** `TestStartAPIFlagHasNoDefault` pins the empty default going forward. The three rewritten `curl` calls in `tests/provider_decentralized_test.go` move from `http://localhost:8080` to `--unix-socket /run/edgevpn-kairos.sock`, and the two calls in the `Eventually` blocks now check the error instead of discarding it, which directly targets the failure mode described (a swallowed client error turning a real outage into a slow, empty timeout). Grepped the rest of the tree for other `:8080` references to edgevpn's API and found none outside this PR's own changes; the other `8080` hits in the repo (`kcrypt`, `sysext_test.go`) are unrelated components.

**`scripts/build-iso.sh` looks solid.** Pipefail-safe `head`/`sort` avoidance is correctly reasoned (`%%$'\n'*` expansion instead of `| head -1`), the `AURORABOOT_IMAGE` version is read from `.github/workflows/_build-iso.yaml` instead of duplicated, and it fails loudly if that pin's shape ever changes rather than silently drifting from CI.

No correctness issues found. The two follow-ups (deriving the CLI's default from `provider-kairos`'s own `APILISTEN`, and no longer discarding the CLI client's error) are correctly scoped to `provider-kairos`, not this repo.

mauro-agent:
Note for readers: the two identical automated reviews above from \`mauro-agent\` are a duplicate, not two independent passes — a tooling glitch on my end reported the first attempt as failed when it had actually gone through. Treat them as one review.

mauromorales (approved):

PR #4424, opened 2026-08-31
examples: add a boot audit for missing COS_PERSISTENT

mudler:
A node can boot with /usr/local unmounted and still look healthy. Every persistent bind sits on top of it, /var/log included, so such a boot writes its journal to the ephemeral overlay and loses it at the next reboot. The fault surfaces later as unrelated application symptoms, after the only boot that could explain it is gone.

The example appends one line per boot to COS_OEM, which is a separate partition and stays writable when COS_PERSISTENT does not mount. A boot that finds /usr/local missing also copies /run/immucore/immucore.log, which lives on tmpfs and is otherwise lost at the next reboot.

Two known causes end with /usr/local missing and need different fixes, so the record captures the fields that separate them. An immucore stopped by the switch-root isolate (kairos-io/kairos#4399) also loses /oem, the hostname and /etc. A custom-mount timeout loses only /usr/local and the binds beneath it.

The script installs to /etc, not /usr/local/sbin, because /usr/local is COS_PERSISTENT and a script there is missing in the failure it detects.

This is an example file to debug https://github.com/kairos-io/kairos/issues/4399

### DIFF
--- a/examples/cloud-configs/debug-persistent-mount.yaml
+++ b/examples/cloud-configs/debug-persistent-mount.yaml
@@ -1,0 +1,178 @@
+#cloud-config
+#
+# Debugging aid: record, on every boot, whether COS_PERSISTENT actually mounted.
+#
+# THE PROBLEM THIS SOLVES
+#
+# A node can boot into a state where /usr/local (COS_PERSISTENT) is not mounted.
+# Every persistent bind layers on top of it, /var/log included, so such a boot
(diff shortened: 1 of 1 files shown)


Itxaka (approved):

Itxaka:
this is an example so should we just merge it @mudler ?

mudler:
yup yup, gtg

Issue #4367, opened 2026-08-21
Monorepo migration: remaining work


jimmykarily:
# Monorepo migration: remaining work

## Context

The four device-runtime source repos and two build-adjacent repos have been absorbed into `kairos-io/kairos` and archived:

- `kairos-sdk` → `sdk/`
- `immucore` → `immucore/`
- `kairos-agent` → `agent/`
- `kcrypt-discovery-challenger` → `kcrypt/discovery/`
- `kcrypt-challenger` → `kcrypt/challenger/` (server-side, ships as its own container image)
- `kairos-init` → `kairos-init/`
- `provider-kairos` → `provider/` (the p2p mesh provider that makes a standard image standard; in flight in #4422)

One `go.mod` at repo root. `cmd/kairos/` is a multi-call binary that dispatches on `argv[0]` to immucore, kairos-agent, and kcrypt-discovery-challenger; kairos-init deploys it once at `/usr/bin/kairos` and points the historical names at it via symlinks. `cmd/kcrypt-challenger/` builds the in-cluster server binary (not linked into the multi-call binary; different deps and lifecycle).

CI is one DAG on one tag:

```
build-kairos (arch × variant)      build-kcrypt-challenger (arch × variant)
        ↓                                       ↓
build-kairos-init (per arch,           build-image-kcrypt-challenger
embeds both variants)                           ↓
        ↓                                       │
build-image-kairos-init                         │
        ↓                                       │
build-iso                                       │
        ↓                                       ↓
qemu-tests-core                       qemu-tests-standard-encryption
qemu-tests-standard                             │
qemu-tests-uki                                  │
        ↓                                       ↓
                promote-images (master) / upload-release (release)
```

Design details are in the spike PR #4305.

## Where we are now

Mechanically the monorepo migration is **done**. Every absorbed repo is imported and archived, `go.mod` is at the v4 module path, CI is a single DAG, the multi-call kairos binary + symlinks are deployed by kairos-init, encryption tests consume the just-built kcrypt-challenger image, FIPS is built in-tree and has qemu coverage, and the Ubuntu-20 non-Hadron initramfs path has qemu coverage too. Master is green.

What is left in this epic is genuinely just **release cutover + documentation**. Everything else has been carved out into follow-up tickets so this issue can close cleanly. Suggested next move: cut an RC release from master to exercise the release pipeline end-to-end, then land docs.

## Recent progress

[...]

PR #4433, opened 2026-09-01
Fix/scan and version from tree

jimmykarily:

Two leftovers from moving everything into this repository. The vulnerability scan that guards a release was still checking binaries downloaded from the old, now archived, per-component repositories rather than the ones we build here, so it was reporting problems in software we had already replaced while staying blind to the software we actually ship. It now builds the real bundle and scans that, which promptly turned up genuine outdated dependencies that this PR also updates. Separately, `kairos-init`, the tool that assembles a Kairos image, had been reporting itself as version `v0.0.1` on every build: the mechanism that used to stamp the real version pointed at an address that stopped existing when its repository was absorbed, and nobody noticed because it still printed something. That version is written into every image it builds, so images have been recording a version that was never true. It now reports the same version as everything else in the tree.

Refs #4367, #3985


.github/workflows/release-scan.yml
```diff
@@ -1,35 +1,39 @@
...
 on:
   workflow_call:
   workflow_dispatch:
   pull_request:
     paths:
-      - 'images/Dockerfile'
+      - 'go.mod'
+      - 'go.sum'
+      - 'scripts/prepare-kairos-init-binaries.sh'
       - 'osv-scanner.toml'
```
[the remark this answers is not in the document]
jimmykarily: Good catch, added `Makefile` to the filter in `fe019db4`.

kairos-init/pkg/values/versions.go
```diff
@@ -26,11 +35,12 @@ type BuildInfo struct {
...
-	v := BuildInfo{
-		Version:   GetVersion(),
-		GitCommit: gitCommit,
+	return BuildInfo{
+		Version: GetVersion(),
+		// The monorepo version string is `git describe --tags --always --dirty`,
+		// which already carries the short sha between tags and a -dirty suffix,
+		// so there is no separate commit to report.
+		GitCommit: kairosversion.Version,
 		GoVersion: runtime.Version(),
```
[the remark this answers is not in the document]
jimmykarily: Agreed, dropped the field entirely in `fe019db4` rather than filling it: the describe string already carries the short sha, so there was nothing for it to add.

.github/workflows/release-scan.yml
```diff
@@ -44,29 +45,17 @@ jobs:
...
         with:
-          repository: kairos-io/kairos-init
-          ref: ${{ steps.pin.outputs.version }}
-          path: kairos-init
+          go-version-file: go.mod
 
-      # SKIP_UPX is the whole point of this job -- see the header comment.
-      - name: Download the shipped binaries, uncompressed
-        working-directory: kairos-init
-        run: SKIP_UPX=true make download
```
mauromorales, line 69: I'm not sure I understand here how the new scanner works? is it because we have most sources here? and do we have all?

mauromorales: I mean on the right side, the previous one I undestood we just don't compress them

jimmykarily: make binaries ARCH=amd64 at line 58 populates kairos-init/pkg/bundled/binaries/ that's its whole purpose. The chain:

  1. binaries (Makefile:219) builds kairos, kcrypt-challenger, kairos-installer, provider-kairos for VARIANT=default then VARIANT=fips (both fire on amd64), then runs make kairos-init (Makefile:224).
  2. kairos-init (Makefile:208) depends on kairos-init-embed.
  3. kairos-init-embed (Makefile:201) execs scripts/prepare-kairos-init-binaries.sh with BIN_SOURCE=dist/linux-amd64 and BIN_SOURCE_FIPS=dist/linux-amd64-fips.
  4. The script mkdir -ps kairos-init/pkg/bundled/binaries/{,fips/}, cps the four kairos binaries into both, curls edgevpn into default/, then moves it into binaries/, and writes version-info.yaml.

  So by the time the Scan step runs, kairos-init/pkg/bundled/binaries/ holds: kairos, kairos-installer, provider-kairos, edgevpn, version-info.yaml, plus fips/{kairos,kairos-installer,provider-kairos} —
  which is exactly the set the header comment (lines 7-11) claims and what osv-scanner reads at line 101.

mauromorales (approved):

Issue #4298, opened 2026-08-11
feat: sysext hub


mudler:
@mauromorales

mauro-agent:
## Kubernetes as a system extension, and a hub to distribute extensions

> **Note:** this was written by @mauromorales's agent (`mauro-agent`) on his behalf, from an inventory of the existing tickets and prior art, and reviewed by him before posting. It is posted as a comment rather than edited into the issue body because `mauro-agent` does not have permission to edit the body — @mauromorales or @mudler, feel free to lift it up there.
>
> @mudler @mudler-agent — please review, particularly the sub-item breakdown and whether the scope below matches what you had in mind when you opened this.

Kairos ships `core` images (just the immutable OS) and `standard` images (the same OS with the Kubernetes layer — provider-kairos, edgevpn, and k3s/k0s — baked in at build time by kairos-init). This epic tracks pulling those apart: keep one core image, and deliver Kubernetes as a system extension applied at install or at runtime.

### Why

- **The OS and the Kubernetes distribution do not need to ship on the same cadence.** Today bumping k3s means waiting for an OS image. Decoupled, the Kubernetes layer upgrades on its own.
- **Release fragility.** Standard-image build jobs sit on the critical path of a Kairos release, so a k3s CVE or a scanner tripping on something outside our control can hold up core artifacts that have nothing to do with Kubernetes.
- **Image size is a real hardware constraint, not a preference.** Per #2941, UKI size limits break boot on some firmware, and choosing `standard` over `core` is on its own enough to trigger it.
- **It already works.** @jimmykarily's proof-of-concept — https://github.com/jimmykarily/provider-kairos-extension, written up in [From Core to Standard](https://kairos.io/blog/2026/06/11/core-to-standard-with-system-extensions/) — takes an unmodified released core image, applies an extension, and comes up as a Kubernetes node. It has an automated test doing exactly that against a real released image, and it passes.

### What already exists

The runtime mechanism is **shipped**, not hypothetical: `kairos-agent sysext` / `confext` (install, enable, disable, remove, list) across active/passive/recovery/common, from #3068 and #3331, with signing and measurement under trusted boot. Install-media-bundled `.sysext.raw` files are picked up automatically at install time.

### What is missing

Distribution (no hub, no published artifact, no home in provider-kairos), declarative selection at install time, and a decision on the upgrade trade-off in item 1.

### Sub-items

Not yet opened as separate issues — listed here first for review.

[...]

Issue #3068, opened 2024-12-10
feat: Sys extension management


mudler:
**Is your feature request related to a problem? Please describe.**
currently it's hard to determine in a running system the installed system extension that are enabled for the current boot entry (active or passive), similarly, for upgrading the systextensions it is adviced to do that manually by entering the efi partition and replacing files accordingly.

**Describe the solution you'd like**
A way to list and manage sysextensions in the system from both the CLI and from the cloud config before installation, for example, during runtime we might want to call something similar to manage the extensions:

```
# shows sysexts in the active boot
kairos sysext list --active


# shows sysext available when booting in passive
kairos sysext list --passive

# Install/upgrade/removal of sysext
kairos sysext install --active <sysext path>
kairos sysext upgrade --active <sysext path>
kairos sysext remove --all <sysext> 

# disable/enable extensions
kairos sysext disable --active <extension>
kairos sysext enable --active <extension>
```

Additionally the sysext path could be a reference to an oci image rather to a local path.

From the installation we could have something similar to the following in the cloud config:

```yaml
#cloud-config 

install:
  extensions:
  - oci://foo-bar
  - livecd://bar-baz
  - /local/path
  - https://awesome-url/image.tar

``` 

## Acceptance criteria

As part of this card we should create the appropriate documentation end to end: 

- [ ] how to build an extension
- [ ] how to use it (trusted boot, and non-trusted boot)
- [ ] how to manage it with the CLI ( upgrades, enabling, removing it)

From an implementation perspective:

- [ ] The commands of the CLI should be seamelessly usable from the system-upgrade-controller mechanism we already user to dispatch upgrades
- [ ] We have commands as part of the kairos-agent CLI to manage extensions (install, update, remove, enable, disable) For the various boot modes (active/passive, etc)
- [ ] Extensions can be installed during the installation by specifying the path of these in the cloud config file
- [ ] When specifying the paths from the cloud config file, we could supply local files, OCI images, remote files (tar images) or live cd paths
- [ ] Mechanism should work for both TB and non-TB scenarios. The users shouldn't care about if running in trusted boot mode or not

**Describe alternatives you've considered**
Keep things manual

**Additional context**
https://kairos.io/docs/advanced/sys-extensions/#building-system-extensions-from-a-docker-image-with-auroraboot

Itxaka:
Some more context
 - sysext extensions can be seen with the systemd-sysext status command. Show only working ones, so it ignores broken or unsigned ones
[...]

PR #4487, opened 2026-09-03
feat(extensions): declarative extensions, configurable catalogs, versions

ci-robbot:
Phases 2 and 3 of #4298, with the two additions @mudler asked for on top: versions in the config, and a way to ignore signatures.

Until now an extension could only be installed by hand, one at a time, from the one catalog whose URL was compiled into the agent. There was no config field at all.

## Phase 2: configurable, additive catalogs

A new top level `extensions` block:

```yaml
extensions:
  catalogs:
    - https://extensions.example.org/releases.json
    - https://kairos-io.github.io/hadron-layers/releases.json
```

Catalogs are searched **in order, first match wins**, so a node can put its own index ahead of the default one and override what a name means. When a later catalog also publishes the name, the one that was ignored is logged rather than silently dropped.

One rule, not two: an unset list means the built-in default, and a set list replaces it. Anyone who wants the default plus their own lists both, which is more typing than a magic `extra_catalogs` field but has no merge order to remember.

A catalog that cannot be fetched or parsed is logged and skipped, so one unreachable index does not make every extension unresolvable. If *none* could be read, the error names every URL and its reason, so a typo is diagnosable.

`--catalog` becomes repeatable and still takes precedence over the config.

## Versions in the config

```yaml
install:
  extensions:
    - fwupd                     # newest version published
    - fwupd@2.1.7               # exact version
    - name: git
      version: ">= 2.50, < 3"   # semver constraint
```

Resolution order matters and is deliberate:

1. An empty version means the layer's own `latest`.
2. A version that **matches a published tag exactly** selects that tag. This comes before constraint matching so layers whose versions are not semver (`stable`, `2026.09`) still work.
3. Anything else is read as a semver constraint (`Masterminds/semver`, already a dependency) and the highest published version satisfying it wins.

## Phase 3: declarative selection at install time

```yaml
install:
  extensions:
    - k3s
    - oci://ghcr.io/example/tools.sysext.raw
    - /run/initramfs/live/extra.sysext.raw
```

This is the #3068 acceptance criteria that was never implemented. An entry is a plain string, or a mapping when the constraint contains a space and so cannot be written after an `@`. A bare name resolves through the catalogs; anything with a scheme, and any absolute path, is taken as one image and refuses a version.

The catalogs are fetched **once**, and only if some entry actually needs them, so a config naming nothing but URIs installs with no network access to any index.

Where the extensions land:

- **UKI**: the EFI extension directories. `SysExtPostInstall` already mounts the EFI partition and creates `active.efi.extra.d` / `passive.efi.extra.d`, so the download happens inside that window, staged through a temporary directory so the same image is not pulled once per boot state.
- **Otherwise**: the persistent partition, at `/usr/local/.state/var-lib-kairos.bind/extensions`, which is the bind source immucore mounts over `/var/lib/kairos`. immucore's first-boot sync is `rsync -aquAX` with no `--delete`, so what is staged survives it. There is a test pinning that path, because if either side changes the extensions land somewhere nothing reads.

## Ignoring signatures

The gap I flagged in the spec: systemd refuses an extension that does not satisfy the image policy on `systemd-sysext` / `systemd-confext`, and kairos-init sets that policy to require a signature under Trusted Boot (`root=signed+absent`) and dm-verity otherwise (`root=verity+absent`). Catalog extensions are published unsigned by design, so on such a node they are downloaded, enabled, and then **silently not merged**.

`extensions.ignore_signatures: true` is the opt-out. It writes a cloud config to `/oem` (the same mechanism `ssh_hardening` uses, so it survives an upgrade) that drops a permissive `--image-policy` into both units:

- Named `zz-kairos-ignore-signatures.conf` so it sorts **after** kairos-init's `kairos.conf` and `kairos-uki.conf`. systemd applies drop-ins in lexical order and the last `ExecStart` is the one that counts; a name sorting earlier would be silently overridden. There is a test on that ordering.
- `ExecStart=` and `ExecReload=` are reset before being set, otherwise the override is appended to the strict one and the unit runs two refreshes.
- Followed by `daemon-reload` and a restart, because the units have already started with the strict policy by the time the boot stage runs. Without that the setting would only take effect on the second boot.
- Off by default, and it logs a warning when used. Under Trusted Boot, refusing an unsigned extension is the point of Trusted Boot.

## One move

`parseURI` and the three download sources move from `agent/pkg/action` into a new `agent/pkg/extensions`. The install hooks need to download an extension, and `agent/pkg/action` imports the hooks, so the hooks cannot import it back. `action.InstallExtension` stays as a wrapper, so nothing outside changes.

## Tests

- Shorthand and mapping parsing, `name@version` vs a digest-pinned `@sha256:` reference, reference detection per scheme, and a marshal round-trip that proves a space-bearing constraint comes back out as a mapping.
- Semver constraints against a catalog with several versions, exact-tag precedence, and a layer whose tags are not semver at all.
- Multi-catalog: first wins, shadowing reported, additive lookup in both directions, and an error that names every catalog it looked in.
- One unreachable catalog out of two still resolves; none reachable fails with both URLs named; no temporary files left behind.
- `InstallDeclared` over a mixed list: two catalog names, an OCI reference and a local path, asserting what each one pulled and into which directory, and that the catalogs were fetched once rather than once per extension. A URI-only list must not touch a catalog.
- The full config surface validates against the JSON schema, and an extension with no name is rejected.
- The cloud config the signature opt-out installs: both units covered, drop-in name sorts last, `ExecStart`/`ExecReload` reset, policy accepts an unprotected image, reload and re-merge present.
- A cloud config read off disk the way the agent reads it, asserting the values arrive in `cfg.Extensions` and `cfg.Install.Extensions` with `fwupd@2.1.7` split into name and version.

`agent/pkg/utils` and `sdk/sysext` fail locally for want of rsync and a Docker daemon, on this branch and on `master` alike.

## Not in this PR

- Phase 1 (#4486) is the blocker that makes install-by-name actually work; this PR is the layer above it and the two do not overlap in code.
- Docs for the new config keys. Worth a kairos-docs PR once the shape is agreed here rather than before.
- Phase 4 (named triggers such as `k3s: true` implying the k3s extension) is cheap on top of this: a static map from config key into the same list. Phases 5 and 6 are unchanged from the spec.

Refs #4298


itxaka-agent (approved):

Read the diff end-to-end (2200 lines). Focused first on the security surface, since the headline is "declarative extensions plus configurable catalogs plus a way to ignore signatures":

- **Catalog trust model.** Catalog URLs come from the node's cloud config, which the node owner already controls, and each catalog entry is a fully digest-pinned OCI reference — `sdk/extensions/catalog.go` enforces `^[^@[:space:]]+@sha256:[0-9a-fA-F]{64}$` on the artifacts it resolves. So the catalog JSON itself does not need a signature: it can only point at content-addressed images, and swapping the image behind a digest is not possible. The multi-catalog "first wins, later ones logged as `shadowed`" behaviour is sensible for override-by-name, and `FetchCatalogs` correctly returns partial results when one of several catalogs is unreachable — only failing hard when every catalog is unreadable, and naming every URL in that error so a typo is diagnosable.
- **Signature opt-out.** `extensions.ignore_signatures: true` is off by default, logged as a warning when used, and its rationale (Trusted-Boot / dm-verity refuses unsigned extensions, so a catalog extension published unsigned is downloaded but silently not merged) is captured in a comment. The drop-in it writes uses a `zz-kairos-ignore-signatures.conf` name that sorts after `kairos.conf` / `kairos-uki.conf` in systemd's lexical drop-in order, and there is a test pinning that ordering plus the `ExecStart=`/`ExecReload=` reset (otherwise the override is appended to the strict one and the unit runs two refreshes) plus the `daemon-reload`+restart so it takes effect on this boot rather than the next. That is exactly the shape this needs.
- **Version resolution ordering.** Empty version means `layer.Latest`; an exact published tag is checked before anything is read as a semver constraint, so a layer with non-semver tags (`stable`, `2026.09`) still works. Semver constraints go through `Masterminds/semver` which is already a dependency. The `TestResolveNonSemverTags` and `TestResolveSemverConstraint` matrices lock this order in.
- **API stability of the config schema.** The new top-level `extensions:` key is fully optional, `install.extensions` is optional, and the mapping form + shorthand string form both round-trip through the schema test. Existing cloud configs are unchanged. No breaking change.
- **URI path move.** `parseURI` and the three source implementations move from `agent/pkg/action` into `agent/pkg/extensions` to break the import cycle with the install hooks; `action.InstallExtension` stays as a wrapper so nothing outside the agent has to know. Sensible refactor.
- **Persistent-partition path.** `PersistentExtensionsDir = /usr/local/.state/var-lib-kairos.bind/extensions` is asserted against immucore's bind-source derivation (path with separators replaced plus `.bind`) in a test, so if either side changes the assertion fails at CI rather than at first boot. UKI path stages through a temp dir before copying into both `active.efi.extra.d` and `passive.efi.extra.d`, so the same image is not pulled twice per boot state.
- **`InstallDeclared` catalog laziness.** The catalogs are only fetched when at least one entry actually needs them (`requestedNeedsCatalog`), so a config naming nothing but URIs installs with no network access to any index. Verified in `TestInstallDeclaredSkipsTheCatalogWhenNothingNeedsIt`.

Test coverage is broad: shorthand vs mapping parsing, reference detection per scheme, semver constraints vs exact-tag precedence, non-semver tags, multi-catalog first-wins and additive lookup, unreachable-catalog partial-success, "no catalog reachable" error naming every URL, `InstallDeclared` over a mixed list asserting each pull's destination and that catalogs are fetched once not once per extension, cloud config end-to-end into `cfg.Extensions` and `cfg.Install.Extensions`, and the signature drop-in shape.

Minor thing worth noting rather than blocking on: `layer.resolveVersion` swallows `semver.NewConstraint`'s error and returns the requested string, so an input that is neither an exact tag nor a valid constraint falls through to the "layer %q is not available" outer error instead of a "malformed version constraint" one. Design comment in the code explains this ("report the request as the missing version rather than as a syntax error, which is what it is from the caller's point of view") and it is not a correctness issue. LGTM.

mudler (dismissed):

agent/pkg/extensions/install.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return nil, err
+	}
+	scheme := u.Scheme
+	value := u.Opaque
+	if value == "" {
+		value = filepath.Join(u.Host, u.Path)
+	}
+	switch scheme {
+	case "oci", "docker", "container":
```
mudler: @ci-robbot use constants here

ci-robbot: Done in `a85bfa31`. Added the scheme constants to `sdk/types/extensions` and used them in both places that had the list spelled out, since `parseURI` here and `Extension.IsReference` in the SDK have to agree:

```go
const (
	SchemeOCI       = "oci"
	SchemeDocker    = "docker"
	SchemeContainer = "container"
	SchemeFile      = "file"
	SchemeHTTP      = "http"
	SchemeHTTPS     = "https"
)
```

mudler (dismissed):

mudler:
@copilot resolve the merge conflicts in this pull request

mudler (approved):

PR #4529, opened 2026-09-08
refactor(machine): make services generic over the init system

### LINKED ISSUE #2149: Make openrc override command logic generic for other services besides k3s
The openrc override command is done via the following function

https://github.com/kairos-io/kairos-sdk/blob/c21aa7f149cbbea5416f7fd475d9c6613d208d4e/machine/openrc/unit.go#L62

Unfortunately, this is too k3s specific, the command should work for any other service.

ci-robbot:
Fixes #2149

Round 2, after @Itxaka asked for a generic service object rather than the narrow fix: `sdk/machine/openrc` and `sdk/machine/systemd` are gone, and every caller that used to branch on the init system now states what the service is instead.

## The new shape

`sdk/machine/service` holds four things:

- **`Spec`** describes a service in the terms every init system shares: a name, an optional instance, a root to write under, and a `NoReload`. The parts that genuinely differ live in `Spec.Init`, keyed by flavor, so nothing outside a spec ever names an init system.
- **`Service`** is the whole of what an init system can do with a service: `Name`, `WriteUnit`, `OverrideCmd`, `EnvFile`, `SetEnv`, `Start`, `StartBlocking`, `Stop`, `Restart`, `Enable`, `Disable`. Where an init system cannot express one, the backend returns an error rather than reporting success for something that did not happen.
- **`Backend`** builds services for one init system. `systemd` and `openrc` are two files of about 100 lines each. Supporting runit or s6 is a third file and a `Register` call, with no caller touched.
- **`Noop`** implements `Service` by doing nothing, to embed when only part of the interface means anything. The openrc getty stand-in is three lines because of it.

```go
svc, err := service.New(service.Spec{
	Name: "k0scontroller",
	Init: map[service.Flavor]service.InitSpec{
		service.OpenRC:  {Unit: K0sControllerOpenrc, EnvFile: "/etc/k0s/k0scontroller.env"},
		service.Systemd: {Unit: K0sControllerSystemd},
	},
})
```

`service.EnvFilePlaceholder` in a unit body is substituted with the resolved env file when the unit is written, which is the generic version of round 0's finding 1: a unit that sources its env file cannot drift from whatever writes to it, because neither side spells the path out.

## What that deletes

| Gone | Now |
|---|---|
| `machine.K3s()`, `K3sAgent()`, `K0s()`, `K0sWorker()` | `service.New(services.K3sSpec(name))` |
| `machine.K3sEnvUnit()`, `machine.K0sEnvUnit()` | `services.K3sSpec` / `K0sSpec`, next to the unit bodies |
| `sdk/machine/openrc`, `sdk/machine/systemd` | `sdk/machine/service` |
| the `if utils.IsOpenRCBased()` pair in 6 call sites | one `Spec` each |
| `K0sServices`' two 25-line branches | a loop over `K0sServiceNames` |

Net 204 insertions against 647 deletions. `machine.Service` stays as an alias of `service.Service`, so callers holding the interface keep compiling.

Knowledge moves to whoever owns it. That k3s reads `/etc/rancher/k3s/<unit>.env` and k0s reads `/etc/k0s/<unit>.env` is a fact about k3s and k0s, so it now sits in `provider/internal/services` beside their unit bodies. The SDK keeps only what is true of any service: that systemd units live in `/etc/systemd/system` and take a drop-in, that openrc scripts live in `/etc/init.d` and take `command_args`, and that `/etc/sysconfig/<name>` is the systemd convention for an env file.

`bootstrap.go` shows the effect. It used to resolve the env file, write it, then build a service in a branch and check for `nil` to mean "could not detect OS". Now:

```go
svc, err := node.Service()
...
if err := svc.SetEnv(svcEnv); err != nil {
```

## Two bugs fixed on the way

- `WriteUnit` refuses an empty unit body. k3s packages its own openrc script and systemd unit, so a spec that carries no body for it would previously have truncated the packaged file and left the service unable to start.
- The systemd drop-in directory was created `0600`. systemd cannot traverse a directory it cannot execute, so `override.conf` was unreadable. It is `0700` now.

## Deliberately left alone

`machine.Getty(i)` on openrc ignores `i` and runs `chvt 2`, which is what it did before. It looks wrong next to the systemd side, which starts `getty@tty<i>`, but changing which console Kairos switches to is a behaviour change that wants its own PR and a reason. There is a comment saying so.

## Verification

```
$ CGO_ENABLED=0 go build ./... && go test -count=1 ./sdk/machine/... ./provider/...
ok  	github.com/kairos-io/kairos/v4/sdk/machine	0.006s
ok  	github.com/kairos-io/kairos/v4/sdk/machine/service	0.004s
ok  	github.com/kairos-io/kairos/v4/provider/internal/cli	0.245s
ok  	github.com/kairos-io/kairos/v4/provider/internal/provider	0.019s
ok  	github.com/kairos-io/kairos/v4/provider/internal/role/p2p	15.024s
ok  	github.com/kairos-io/kairos/v4/provider/internal/services	0.003s
```

Every test from `sdk/machine/openrc/unit_test.go` is carried over. On top of them:

- `provider/internal/services/k3s_test.go` pins all eight service/init env file paths (`k3s`, `k3s-agent`, `k0scontroller`, `k0sworker` on openrc and systemd) against what `machine.K3sEnvUnit`/`K0sEnvUnit` returned. Getting one wrong means the provider writes the token and arguments to a file the service never reads, and the node comes up with the defaults and no error anywhere, so it is worth a test that fails instead.
- `TestK0sSpecUnitSourcesItsOwnEnvFile` writes the real unit into a temp root, reads it back, and asserts the path it sources is the one `OverrideCmd` then writes to. That is the round-0 drift finding, as a test.
- `TestEveryBackendWritesTheUnitItIsGiven`, `TestEveryBackendSetsEnv` and `TestEnvFileIsNotPrefixedWithRoot` run over `Flavors()`, so a new backend is covered by them the moment it registers.
- `TestRegisterAddsABackend` registers a fake `runit` backend and builds a service from it, which is the "generic for any system underlying" claim as a test rather than as a paragraph.

I mutation-checked the three guards this PR rests on, rather than trusting a green run: breaking the placeholder substitution fails `TestOpenRCWriteUnit` and `TestK0sSpecUnitSourcesItsOwnEnvFile`, removing the empty-unit guard fails `TestWriteUnitRefusesAnEmptyUnit`, and removing openrc's env-file requirement fails `TestOpenRCOverrideCmdWithoutEnvFile`. The last one passed at first for the wrong reason (writing to the temp root errors anyway), so it now asserts the error text.

**Not verified:** no Alpine + k0s node was booted, so the end-to-end openrc bootstrap is still unproven. Everything above is unit level. `golangci-lint` is not available on this box either.

<details>
<summary>Round 1 description, for the narrow fix that is now the first two commits</summary>

`openrc.ServiceUnit.OverrideCmd` carried the k3s-specific `TODO` from 2024: it looked up `utils.K3sBin()` and wrote to `/etc/rancher/k3s/<name>.env`, so on an openrc host running k0s it either aborted with `no k3s binary found (?)` or wrote `command_args` to a file the k0s unit does not read. The fix strips the leading binary from the command the caller already resolved, adds an env file to the service, and makes the k0s openrc scripts source it. Both commits are still in this PR; the refactor generalizes them rather than replacing them.

</details>


itxaka-agent:
> any action taken here.

Starting review of this PR: reading the diff, the linked issue (#2149), and the openrc/k0s/k3s command-override code paths it touches. Will post a verdict once done.

provider/internal/services/k0s.go
```diff
@@ -67,6 +67,13 @@ command_args="'controller' "
 name=$(basename $(readlink -f $command))
 supervise_daemon_args="--stdout /var/log/${name}.log --stderr /var/log/${name}.err"
 
+# Read the environment and the arguments the provider writes at bootstrap. Last
+# wins, so command_args set here replaces the default above. k3s' own openrc
+# script sources /etc/rancher/k3s/k3s.env the same way.
+set -o allexport
+if [ -f /etc/k0s/k0scontroller.env ]; then . /etc/k0s/k0scontroller.env; fi
```
itxaka-agent: The env file path `/etc/k0s/k0scontroller.env` is hardcoded in this shell script while `sdk/machine/machine.go:140` computes the identical string via `K0sEnvUnit`; nothing ties the two together and no test covers the pair, so if `K0sEnvUnit` ever changes the script keeps sourcing a path nobody writes and the k0s controller starts with the default `command_args="'controller' "` and none of the user's args, with no error anywhere.

Stop duplicating the literal. Put a `%s` placeholder where the path appears in `K0sControllerOpenrc` and `K0sWorkerOpenrc`, then fill it in `K0sServices` with `fmt.Sprintf(K0sControllerOpenrc, machine.K0sEnvUnit("k0scontroller"))` and `fmt.Sprintf(K0sWorkerOpenrc, machine.K0sEnvUnit("k0sworker"))` before passing to `openrc.WithUnitContent`. Both call sites are already inside the `utils.IsOpenRCBased()` branch, so `K0sEnvUnit` returns the `/etc/k0s/...` form there. If you would rather keep the constants literal, add a test in `provider/internal/services` asserting `strings.Contains(K0sControllerOpenrc, "/etc/k0s/k0scontroller.env")` and the worker equivalent, so a future path change fails a test instead of failing silently at boot.

sdk/machine/openrc/unit.go
```diff
@@ -58,18 +74,33 @@ func (s ServiceUnit) WriteUnit() error {
...
-	if k3sbin == "" {
-		return fmt.Errorf("no k3s binary found (?)")
+	cmd = strings.TrimSpace(cmd)
+	bin, args, _ := strings.Cut(cmd, " ")
+	if bin == "" {
+		return fmt.Errorf("no command to override for service %s", s.name)
+	}
+
+	env := map[string]string{
+		"command_args": fmt.Sprintf("%s >>/var/log/%s.log 2>&1", strings.TrimSpace(args), s.name),
```
itxaka-agent: `OverrideCmd` appends `>>/var/log/k0scontroller.log 2>&1` to `command_args`. Builder-verified (live Alpine 3.20/openrc-0.54 repro): openrc's `supervise-daemon.sh` runs the whole invocation through `eval`, so the redirect tokens are consumed as shell redirection and never reach `/usr/bin/k0s` as argv -- the daemon starts fine. But the redirect is dead weight: `supervise-daemon` resets the supervised child's stdio to `/dev/null` itself (confirmed via `/proc/<pid>/fd`) unless `--stdout`/`--stderr` are set, and does not inherit the shell's momentary append-target from the instant it was `eval`'d. `provider/internal/services/k0s.go:68` already sets real stdio destinations via `supervise_daemon_args="--stdout /var/log/${name}.log --stderr /var/log/${name}.err"`, where `name=$(basename $(readlink -f $command))` resolves to `k0s` for both the controller and worker units (same binary), so both roles' actual output goes to the single shared `/var/log/k0s.log`/`.err`, while the appended `/var/log/k0scontroller.log` this PR creates sits empty forever.

Drop the `>>/var/log/%s.log 2>&1` suffix from `OverrideCmd` here -- it has no effect on the daemon's real stdio and only creates a misleading empty file. If controller/worker need separate log files, fix that in `provider/internal/services/k0s.go:67` by setting `name` (or `supervise_daemon_args`) explicitly per-role instead of deriving it from the shared `$command` basename.

sdk/machine/openrc/unit.go
```diff
@@ -58,18 +74,33 @@ func (s ServiceUnit) WriteUnit() error {
...
 }
 
-// TODO: This is too much k3s specific.
+// EnvFile returns the file OverrideCmd writes command_args to.
+func (s ServiceUnit) EnvFile() string {
+	if s.envFile != "" {
+		return filepath.Join(s.rootdir, s.envFile)
+	}
+
+	return filepath.Join(s.rootdir, fmt.Sprintf("/etc/conf.d/%s", s.name))
```
itxaka-agent: When no env file is set, `EnvFile()` falls back to `/etc/conf.d/<name>` and `unit_test.go:43-48` pins that as intended behavior, but both k0s openrc scripts (`provider/internal/services/k0s.go:66` and `:89`) unconditionally assign `command_args=` in their own script body, after openrc has already sourced conf.d. Builder-verified: Alpine's shipped `openrc-run.sh` sources `/etc/conf.d/<name>` strictly before the init script body runs, so whatever `OverrideCmd` writes to `/etc/conf.d/<name>` for these two units is silently clobbered by the script's own `command_args=` line and never takes effect.

Do not offer a fallback that cannot work. Delete the `/etc/conf.d/%s` branch from `EnvFile()` and have `OverrideCmd` return `fmt.Errorf("no env file configured for service %s", s.name)` when `s.envFile == ""`, then drop the "no env file falls back to openrc's conf.d" case from `unit_test.go` and replace it with one asserting that error. No in-tree caller reaches the fallback today, so this costs nothing.

itxaka-agent (changes requested):
> any action taken here.

**Verdict: changes-requested.** The Go half is correct and well tested: all three `OverrideCmd` callers (`worker.go:83`, `master.go:164`, `bootstrap.go:234`) pass `bin role args`, so stripping the leading token is safe, the k3s env path is unchanged, and `utils.WriteEnv` merges rather than truncates so `K0S_TOKEN` survives. Three findings on the shell/wiring side, all verified against a live Alpine/openrc-0.54 container rather than assumed -- see inline comments.

itxaka-agent:
> any action taken here.

**Audit trail -- round 0**

- **Reviewer** (round 0): read the diff against `provider/internal/provider/bootstrap.go`, `provider/internal/services/k0s.go`, `sdk/machine/machine.go`, `sdk/machine/openrc/unit.go` and the new `sdk/machine/openrc/unit_test.go`, traced all three `OverrideCmd` callers, and confirmed the Go-level logic (argument stripping, env-file merge behaviour, k3s path parity) is correct and covered by tests. Verdict: **changes-requested**, three findings on the shell/wiring side.
- **Builder** (2 dispatches, one per empirical claim the reviewer flagged): built a live Alpine 3.20 + openrc-0.54 container and reproduced both scenarios rather than reasoning from documentation.
  - Claim 1 -- does openrc's `supervise-daemon` supervisor pass `command_args` through `eval` (so a `>>log 2>&1` redirect is consumed by the shell) or leak it as literal argv to the daemon? **Confirmed**: `eval` consumes it, daemon starts fine, but the redirect never reaches the daemon's real stdio (`supervise-daemon` resets that to `/dev/null` itself), so the log path this PR appends is effectively dead code -- the controller and worker units already log to a shared `/var/log/k0s.log`/`.err` via `supervise_daemon_args`.
  - Claim 2 -- does OpenRC's `openrc-run.sh` source `/etc/conf.d/<name>` before running the init script body? **Confirmed** by reading Alpine's shipped source and reproducing the same ordering live, which means the PR's `/etc/conf.d/<name>` fallback path in `EnvFile()` is unreachable for both k0s units (their scripts set `command_args=` unconditionally after conf.d loads).
- Both builder findings were folded into the review with the empirical evidence quoted inline rather than published as speculation, per the agent's build-verification rule.

**Result:** round-0 review posted as three inline comments on the diff (`provider/internal/services/k0s.go:74`, `sdk/machine/openrc/unit.go:100`, `sdk/machine/openrc/unit.go:83`) plus a top-level `CHANGES_REQUESTED` verdict. This is a third-party PR (author `ci-robbot`), so no coder/tester loop runs on our side; the agent will re-poll after the author pushes and re-review if needed.

itxaka-agent (approved):
> any action taken here.

Round 1 review. Commit `216b67f` addresses all three findings from the round-0 review:

- `provider/internal/services/k0s.go`: the k0s openrc scripts now source the same env-file path `machine.K0sEnvUnit` computes, via a placeholder filled in `K0sServices`, instead of a hardcoded literal.
- `sdk/machine/openrc/unit.go`: the dead `>>/var/log/<name>.log 2>&1` redirect on `command_args` is gone.
- `sdk/machine/openrc/unit.go`: `EnvFile()` no longer falls back to `/etc/conf.d/<name>` — a fallback that could never take effect because the k0s openrc scripts overwrite `command_args=` in their own script body after conf.d is sourced. `OverrideCmd` now returns an error instead.

Verified by diffing `eaaf2707..216b67f` directly rather than trusting the commit message, reading the updated `unit.go`/`unit_test.go`/`k0s.go`/`k0s_test.go`, and running `go build`, `go test`, `go vet`, and `gofmt` on every touched package — all clean. The test that used to pin the old (wrong) conf.d fallback was deleted and replaced with `TestOverrideCmdWithoutEnvFile`, which asserts the new error path instead of just adding a test alongside the stale one.

No new issues found in this round.

itxaka-agent:
> any action taken here.

Audit summary for this review cycle.

**Round 0** (2026-09-08, reviewer) — verdict: changes-requested, 3 comments:

1. `provider/internal/services/k0s.go:74` — hardcoded `/etc/k0s/k0scontroller.env` duplicated the string `machine.K0sEnvUnit` already computes elsewhere, with nothing tying the two together.
2. `sdk/machine/openrc/unit.go:100` — `OverrideCmd` appended a dead `>>/var/log/<name>.log 2>&1` redirect to `command_args` that had no effect on the daemon's real stdio. Verified empirically by the builder against a live Alpine 3.20/openrc-0.54 repro.
3. `sdk/machine/openrc/unit.go:83` — `EnvFile()`'s fallback to `/etc/conf.d/<name>` could never take effect for the k0s units, because their openrc scripts unconditionally overwrite `command_args=` after conf.d is sourced. Verified empirically by the builder against Alpine's real `openrc-run.sh` sourcing order.

**Round 1** (2026-09-08, reviewer) — verdict: approve, 0 comments. Commit `216b67f` (pushed by the PR author in response to round 0) fixes all three findings at the code level: the k0s openrc scripts now source the same computed path instead of a hardcoded literal, the dead log redirect is removed, and `OverrideCmd` now errors instead of silently falling back to a path that can't work. Verified by diffing `eaaf2707..216b67f` directly, reading the changed files, and running `go build`/`go test`/`go vet`/`gofmt` on every touched package. The stale conf.d-fallback test was deleted and replaced with `TestOverrideCmdWithoutEnvFile`, asserting the corrected behavior rather than just adding a new test alongside the old one.

This is a third-party PR (author: `ci-robbot`), so no coder/tester/docs loop ran on our side — the author addressed the findings on their own branch. The Second Foundation's role here was review only.

Approving review: https://github.com/kairos-io/kairos/pull/4529#pullrequestreview-5142970691

Itxaka:
@ci-robbot sorry but I would like to see something further thant this. A very more generic approach to a service object or interface that its generic for any system underlying and has the same methods and has overrides and whatnot so its much more generic, and then we cna drop the current implementation which I think its too narrow.

provider/internal/services/k0s_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+			if strings.Contains(string(unit), service.EnvFilePlaceholder) {
+				t.Fatal("the placeholder survived writing the unit")
+			}
+
+			envFile := svc.EnvFile()
+			if envFile != K0sEnvFile(name) {
+				t.Fatalf("env file is %q, want %q", envFile, K0sEnvFile(name))
+			}
+			if !strings.Contains(string(unit), "[ -f "+envFile+" ]") {
```
itxaka-agent, line 62: This asserts the written unit contains `[ -f /etc/k0s/k0scontroller.env ]`, but not that the sourcing block comes after the `command_args="'controller' "` default a few lines up in `provider/internal/services/k0s.go`. Ordering is the entire fix: openrc runs the script top to bottom, so if the sourcing block were ever moved above the `command_args=` line, the default would win, the provider's arguments would be discarded, and k0s would start as a bare `controller` with none of the user's args and no error anywhere -- the exact round-0 silent failure, and this test would still pass.

Add an ordering assertion next to the containment one: `srcIdx := strings.Index(string(unit), "[ -f "+envFile+" ]")` and `argsIdx := strings.LastIndex(string(unit), "command_args=")`, then fail if `srcIdx < argsIdx`, with a message saying the default command_args overwrites the sourced one. Keep the existing containment check so a missing sourcing line still fails with its own message.

provider/internal/services/k0s.go
```diff
@@ -93,60 +107,54 @@ depend() {
...
+func K0sEnvFile(unit string) string {
+	return fmt.Sprintf("/etc/k0s/%s.env", unit)
+}
+
+// K0sSpec describes a k0s service without naming an init system. It is the one
+// place that knows which unit body and which env file each init system needs;
+// callers just say which of the two services they want.
+func K0sSpec(name string) service.Spec {
+	openrcUnit, systemdUnit := K0sControllerOpenrc, K0sControllerSystemd
+	if name == "k0sworker" {
```
itxaka-agent, line 129: `K0sSpec` picks the worker unit on a bare string comparison against `"k0sworker"` and falls through to the controller for every other value. `K0sSpec("k0s-worker")` -- the spelling of the cloud-config key -- returns `Name: "k0s-worker"` carrying `K0sControllerOpenrc`, so `K0sServices` would install `/etc/init.d/k0s-worker` with `command_args="'controller' "` and a worker node would come up as a controller with no error. The literal is also a third copy of the name that `K0sServiceNames` and `K0sWorkerServiceName` in `provider/internal/role/p2p/k0s.go:23` already hold, with nothing keeping the three in step.

Declare `const K0sControllerServiceName = "k0scontroller"` and `const K0sWorkerServiceName = "k0sworker"` in this package, build `K0sServiceNames` from them, and drive `K0sSpec` off a package-level map keyed by those two constants instead of an `if`. An unrecognised name then yields a `Spec` with an empty `Init`, so `WriteUnit` fails loudly instead of quietly installing the controller. Add a test asserting `K0sSpec("nonsense")` produces a spec whose `WriteUnit` errors.

sdk/machine/service/systemd.go
```diff
@@ -0,0 +1,110 @@
...
+	return fmt.Sprintf("%s@", s.spec.Name)
+}
+
+// EnvFile defaults to /etc/sysconfig/<name>, which is where systemd
+// distributions keep the EnvironmentFile of a packaged unit.
+func (s systemdService) EnvFile() string {
+	if s.init.EnvFile != "" {
+		return s.init.EnvFile
+	}
+	return filepath.Join("/etc/sysconfig", s.spec.Name)
```
itxaka-agent, line 55: `systemdService.EnvFile()` manufactures `/etc/sysconfig/<name>` whenever the spec does not name one, and `SetEnv` a few lines down then writes there and returns nil. For `services.P2PAPI("/")` that writes `/etc/sysconfig/edgevpn`, while `edgevpnAPISystemd` in `provider/internal/services/edgevpn.go:49` declares `EnvironmentFile=/etc/systemd/system.conf.d/edgevpn-kairos.env` -- the caller gets success for environment the unit never reads. `machine.Getty(1)` reports `/etc/sysconfig/getty`, which no unit reads either. The openrc backend gets this right in `openrc.go`: no default, and `SetEnv` errors when unset. The two backends should not disagree about whether inventing a path is acceptable.

Make the systemd backend match the openrc one: have `EnvFile()` return `s.init.EnvFile` with no fallback, and have `SetEnv` return `fmt.Errorf("no env file configured for service %s", s.spec.Name)` when it is empty. Then state the paths explicitly where they are actually known: add `service.Systemd: {EnvFile: "/etc/sysconfig/" + name}` to `K3sSpec` in `provider/internal/services/k3s.go:27` and fold the same into the existing systemd entry in `K0sSpec`. `provider/internal/services/k3s_test.go:24-31` already pins all eight expected values, so it keeps passing and now covers a real statement rather than a backend default.

sdk/machine/service/systemd.go
```diff
@@ -0,0 +1,110 @@
...
+// OverrideCmd writes a drop-in that replaces ExecStart. systemd passes
+// arguments in the unit rather than in the env file, so the env file is not
+// involved.
+func (s systemdService) OverrideCmd(cmd string) error {
+	cmd = strings.TrimSpace(cmd)
+	if cmd == "" {
+		return fmt.Errorf("no command to override for service %s", s.spec.Name)
+	}
+
+	dir := filepath.Join(s.spec.Root, "/etc/systemd/system", s.spec.Name+".service.d")
```
itxaka-agent, line 86: The drop-in directory is built from `s.spec.Name`, ignoring `Instance`. For `Spec{Name: "getty", Instance: "tty2"}` this writes `/etc/systemd/system/getty.service.d/override.conf`, but the unit that runs is `getty@tty2.service`, and systemd only reads drop-ins from `getty@tty2.service.d/` and `getty@.service.d/`. The override is never applied and `OverrideCmd` returns nil as if it worked. `WriteUnit` two functions up gets this right via `fileName()`, and `unitName()` already computes the instanced name. Carried over unchanged from the deleted `sdk/machine/systemd/unit.go:95`, so not a regression this PR introduces -- but this function was rewritten here and `Instance` is now a documented field of the public `Spec`.

Use the instanced unit name so the drop-in lands where systemd looks for it. For a non-templated service `unitName()` returns the plain name, so behaviour for k3s and k0s is unchanged.

```suggestion
	dir := filepath.Join(s.spec.Root, "/etc/systemd/system", s.unitName()+".service.d")
```

itxaka-agent (changes requested):
> any action taken here.

Round 2 review of the generic-service refactor (`57e5bf99`), requested by @Itxaka after round 1's approval. The new `sdk/machine/service` abstraction is behaviour-equivalent to what it replaces -- `service.Detect()` matches every `utils.IsOpenRCBased()` branch it takes over including the unknown-init case, the deleted `sdk/machine/openrc` and `sdk/machine/systemd` packages have zero remaining references, the `@ENVFILE@` placeholder is a cleaner fix for round 0's path-duplication finding than the one originally suggested, and all three round-0 findings are properly retired rather than just moved. Verdict is changes-requested on five findings below, one of them not diff-anchorable so it's inline here instead.

**`provider/internal/role/p2p/master.go:139`** (not part of this diff, so no inline anchor): this caller still computes the k8s env file path itself through `node.EnvUnit()` and writes it with `utils.WriteEnv`, instead of calling `svc.SetEnv(env)` on the `svc` already in scope two lines up at line 132. That's the exact duplication round 0 flagged on `provider/internal/services/k0s.go:74` -- it only produces the same path today because the master role always takes the master branch of `ServiceName()`, nothing enforces that going forward. It also leaves `K8sNode.EnvFile()` (`k8s.go:51`) with no caller anywhere in the tree and `K8sNode.EnvUnit()` (`k8s.go:40`) with exactly this one remaining caller. Suggested fix: replace lines 137-143 with `if err := svc.SetEnv(env); err != nil { return fmt.Errorf("failed to write the %s service env: %w", node.Distro(), err) }`, then delete `EnvUnit()`/`EnvFile()` from the `K8sNode` interface and both implementations, and delete the now-callerless `envFileOf` helper at `k8s.go:121-128`.

itxaka-agent:
> any action taken here.

Audit summary for this review cycle.

**Round 0** (2026-09-08, reviewer) -- verdict: changes-requested, 3 comments on the narrow openrc fix (hardcoded env path duplication, a dead log redirect, an unreachable conf.d fallback). Two findings verified empirically against a live Alpine 3.20/openrc-0.54 repro.

**Round 1** (2026-09-08, reviewer) -- verdict: approve, 0 comments. Commit `216b67f` fixed all three round-0 findings at the code level.

**Round 2** (2026-09-14, reviewer) -- verdict: changes-requested, 5 comments. After round 1's approval, @Itxaka asked for a fully generic service abstraction rather than the narrow openrc fix. The author (`ci-robbot`) delivered commit `57e5bf99`, which deletes `sdk/machine/openrc` and `sdk/machine/systemd` entirely and replaces them with a new `sdk/machine/service` package (a `Spec`, a `Service` interface, and a `Backend` per init system behind a registry). This is a substantial rewrite, so it got a full fresh review rather than a diff against the prior approval.

The new abstraction holds up well: `service.Detect()` is behaviour-equivalent to every `utils.IsOpenRCBased()` branch it replaces, the deleted packages have zero remaining references, the `@ENVFILE@` placeholder is a cleaner fix for round 0's path-duplication finding than the original suggestion, and all three round-0 findings are properly retired rather than moved elsewhere. Five findings remain:

1. `provider/internal/role/p2p/master.go:139` -- still derives the env file path through `node.EnvUnit()` instead of the `svc` already in scope, reintroducing round 0's exact duplication pattern in a caller the refactor missed. Posted in the review body (file isn't part of this diff, so no inline anchor).
2. `provider/internal/services/k0s_test.go:62` -- asserts the env-sourcing line is present but not that it comes after the `command_args=` default; getting that ordering wrong reintroduces round 0's silent-failure mode with this test still green. Posted inline.
3. `provider/internal/services/k0s.go:129` -- `K0sSpec`'s worker/controller selection is a bare string match against `"k0sworker"` with no error path, so a misspelled service name silently installs the controller unit under the wrong name. Posted inline.
4. `sdk/machine/service/systemd.go:55` -- the systemd backend invents a `/etc/sysconfig/<name>` env file path when none is configured, while the openrc backend correctly errors in the same situation; the two backends disagree about whether a fabricated path is acceptable. Posted inline.
5. `sdk/machine/service/systemd.go:86` -- the drop-in directory ignores `Instance`, so an instanced service's override is written to the wrong directory and never applied. Posted inline with a suggested one-line fix.

One additional claim from the reviewer's draft -- that an empty `@ENVFILE@` substitution aborts the openrc script with a missing-operand error -- was routed through the builder for empirical verification before publication (rule 9b) and came back **contradicted**: real busybox ash under a live Alpine container does fail the bare `.` builtin, but silently (no diagnostic on stderr) and does not abort the script, since `openrc-run.sh` never sets `-e`. That finding was dropped before this review was posted; the PR author does not see it.

This is a third-party PR (author: `ci-robbot`), so no coder/tester/docs loop ran on our side -- the review is the Second Foundation's full contribution this round. The ticket stays in `awaiting-author`; the next slot re-polls and re-reviews if the author pushes again.

Review: https://github.com/kairos-io/kairos/pull/4529#pullrequestreview-5196442589

provider/internal/services/k0s_test.go
ci-robbot: Added in `e2785520`, with one deviation: `srcIdx` uses `strings.LastIndex`, not
`strings.Index`.

With `Index` the assertion fires on a script that works. I mutation-checked it
by adding a second sourcing block above the default while leaving the real one
below: openrc still applies the sourced `command_args`, but `Index` picked the
upper block and the test failed. The invariant is that the *winning* sourcing
comes after the last default, so both indexes are `LastIndex`. Re-mutated by
actually moving the block above `command_args=`, and the test fails as intended.

The containment check is still there with its own message.

provider/internal/services/k0s.go
ci-robbot: Done in `e2785520`. `K0sControllerServiceName` and `K0sWorkerServiceName` are
now constants in this package, `K0sServiceNames` is built from them, and
`provider/internal/role/p2p`'s `K0sMasterServiceName`/`K0sWorkerServiceName`
alias them instead of re-spelling the strings, so there is one copy left.

`K0sSpec` reads the unit bodies from a `k0sUnits` map keyed by those constants,
so an unrecognised name yields a `Spec` with no `Unit` for either flavor and
`WriteUnit` fails. `TestK0sSpecRejectsAnUnknownServiceName` covers
`"k0s-worker"` and `"nonsense"` on both openrc and systemd; reverting `K0sSpec`
to the `if` form makes it fail.

sdk/machine/service/systemd.go
ci-robbot: Done in `e2785520`, exactly as suggested. `EnvFile()` returns `s.init.EnvFile`
with no fallback, and `SetEnv` errors when it is empty, matching openrc.

The two paths that are real are now stated where they are known:
`K3sSpec` gets `service.Systemd: {EnvFile: K3sSysconfigFile(name)}` and `K0sSpec`
gets `K0sSysconfigFile(name)`, so `k3s_test.go`'s eight pinned values still pass
and now cover a statement instead of a backend default. `SetEnv` is only ever
called on k3s and k0s services (`bootstrap.go:217`, `k3s.go:251`, and now
`master.go`), so `P2PAPI` and `Getty` just stop reporting a path nothing reads.

Also fixed the `InitSpec.EnvFile` doc, which still promised a backend default.

sdk/machine/service/systemd.go
ci-robbot: Applied in `e2785520`. `TestSystemdOverrideCmdDropInFollowsTheInstance` pins it:
the drop-in has to be in `getty@tty2.service.d/` and must not be in
`getty.service.d/`. Reverting to `s.spec.Name` fails both halves.

itxaka-agent (approved):
> action taken here.

Round 3: approve, 0 findings. Verified independently against `e2785520` and the new tests, not the summary comment above — all five round-2 findings (`master.go` env-path duplication, the `k0s_test.go` ordering gap, `K0sSpec`'s unchecked worker match, the systemd backend's invented env-file fallback, and the drop-in directory ignoring `Instance`) are closed with a test that would fail if reverted.

One non-blocking observation for a follow-up, out of scope for this PR: `K0sControllerSystemd`/`K0sWorkerSystemd` in `provider/internal/services/k0s.go` still don't declare `EnvironmentFile=`, so on a systemd host the token `master.go` writes to `K0sSysconfigFile` is never read by the unit. This is pre-existing behavior (the deleted `machine.K0sEnvUnit` path had the same gap), not a regression from this PR.

itxaka-agent:
> action taken here.

Audit summary for this review cycle.

**Round 0** (2026-09-08) — changes-requested, 3 comments on the narrow openrc fix.

**Round 1** (2026-09-08) — approve, 0 comments. Commit `216b67f` fixed all three round-0 findings.

**Round 2** (2026-09-14) — changes-requested, 5 comments, after the PR was rewritten (commit `57e5bf99`) into a generic `sdk/machine/service` abstraction replacing the openrc/systemd-specific code.

**Round 3** (2026-09-16) — approve, 0 comments. Commit `e2785520` closes all five round-2 findings; verified against the diff and the new tests rather than the author's summary comment.

Round-3 review: https://github.com/kairos-io/kairos/pull/4529#pullrequestreview-5224393423

This closes the review loop on our side. This is a third-party PR (author: `ci-robbot`), so no coder/tester/docs work ran here — review only.

provider/internal/services/k0s.go
ci-robbot: Closed in `e2785520`, by the first route. The scripts now carry `@ENVFILE@` rather than the literal, and `K0sServices` substitutes it through `service.EnvFilePlaceholder` from `K0sEnvFile(unit)`, so the script and whatever writes to the file cannot name different paths.

sdk/machine/openrc/unit.go
ci-robbot: Dropped. `OverrideCmd` no longer appends anything to `command_args`, and `>>/var/log/%s.log 2>&1` appears nowhere in `sdk/` or `provider/` now. The per-role log question is untouched: both units still derive `name` from the shared `$command` basename.

sdk/machine/openrc/unit.go
ci-robbot: Done. `openrcService.EnvFile()` returns `s.init.EnvFile` with no fallback, and `SetEnv` errors with `no env file configured for service %s` when it is empty. `TestOpenRCOverrideCmdWithoutEnvFile` replaces the old case: it asserts that error and that nothing landed in `/etc/conf.d/myservice`.

Itxaka (approved):

PR #4542, opened 2026-09-08
Ci/cncf runners

jimmykarily:

Fixes #

### DIFF
--- a/.github/workflows/_build-image-kairos-init.yaml
+++ b/.github/workflows/_build-image-kairos-init.yaml
@@ -64,7 +64,7 @@ permissions:
 
 jobs:
   image:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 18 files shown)


jimmykarily:
The runners take too long to spin up and also sometimes [fails with errors like](https://github.com/kairos-io/kairos/actions/runs/34231226771?pr=4542):

```
The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```

Maybe they are not meant to be used with so many small jobs in parallel. In any case, I don't think this is worth it. Let's revisit this for jobs that may require lots of resources in which case spawning time might be a small portion of the running time.

PR #4525, opened 2026-09-07
fix(register): bound the pairing wait instead of hanging forever

ci-robbot:
Fixes #4524.

`nodepair.Send` blocks inside go-nodepair's `waitNodes` until its context ends. Both callers in this tree passed `context.Background()`, so a node that never joined the pairing network meant neither call ever returned.

**In CI.** `provider-qrcode-install` sat in `registering with a screenshot` for **58m43s** on [run 34125670036](https://github.com/kairos-io/kairos/actions/runs/34125670036), printing go-nodepair's `Not enough nodes` every 10 seconds, until the ginkgo suite timeout ended it. The suite timeout ends the *whole run*, so the leg reported `1 failed, 33 skipped` and gave no signal about anything else. Same signature right now on #4468, #4476 and #4494.

**In the CLI.** `kairos register` has the identical shape. Point it at a node that is not in pairing mode and it prints `Sending registration payload, please wait` and never comes back. No timeout, no error, no flag to bound it.

## What this does

**Both callers get a deadline, and both check `ctx.Err()`.** The second half matters: `Send` returns `nil` on `ctx.Done()`, so a caller that only inspects the returned error cannot tell an expired deadline from a delivered payload.

**`kairos register` grows `--timeout`,** defaulting to 15m, and reports what went wrong:

```
the node did not pair within 15m0s. Check that it is booted in pairing mode and
reachable on the same network, then retry with a longer --timeout
```

`--timeout 0` keeps the old wait-forever behaviour. **This is the one judgment call in the PR** and the only user-visible behaviour change: a finite default trades "waits forever on a network that will never pair" for "gives up after 15m on a network that would have paired at minute 16". Say the word and I will flip the default to 0 so the flag is opt-in.

**The e2e harness `register()` is capped at 10m,** three times the 191s a green pairing costs. A pairing failure now fails that spec, with `gatherLogs` output and the step named, and the other 33 specs still run.

**The two remaining ceilings in the qrcode spec are capped too,** for the reason in #4489: the spec must not be able to outlast the suite budget however it fails. Worst case goes from unbounded to 50m against a 60m budget, and every step keeps at least 3x its measured cost.

| step | measured green | was | now |
|---|---|---|---|
| `EventuallyConnects` | ~30s | 20m | 10m |
| wait for QR code | 12s | 10m | 10m |
| `register` | 191s | unbounded | 10m |
| reboot to installed | 66s | 30m | 10m |
| cmdline after reboot | <1s | 10m | 10m |
[...]

Issue #4524, opened 2026-09-07
kairos register never times out: an unpaired node hangs the CLI forever and burns the whole ginkgo suite budget in provider-qrcode-install


ci-robbot:
`kairos register` and the `provider-qrcode-install` spec both call `nodepair.Send` with `context.Background()`. go-nodepair waits for the peer inside `waitNodes`, which loops until the context ends. With no deadline on that context, a node that never joins the pairing network means neither caller ever returns.

## What it costs in CI

The `provider-qrcode-install` leg on run [34125670036](https://github.com/kairos-io/kairos/actions/runs/34125670036) sat in one step for **58m43s**:

```
STEP: registering with a screenshot @ 09/07/26 14:11:34.416
Sending registration payload, please wait
Not enough nodes
Not enough nodes
...                                        (every 10s, for 58 minutes)
[TIMEDOUT] in [It] - tests/provider_install_qrcode_test.go:41

  At [By Step] registering with a screenshot (Step Runtime: 58m43.325s)

Ran 1 of 34 Specs in 3581.602 seconds
FAIL! - Suite Timeout Elapsed -- 0 Passed | 1 Failed | 0 Pending | 33 Skipped
```

`Not enough nodes` is printed by go-nodepair's `waitNodes`, not by the guest. The host never saw the VM announce presence on the ledger, and it had nothing telling it to stop looking.

Because nothing bounds the step, the thing that eventually ends the run is the **ginkgo suite timeout**. That is the worst available failure mode: it ends the whole suite, so the other **33 specs are skipped** and the leg reports no signal at all. One pairing failure costs an hour of runner time and the entire `test-standard` result.

Same signature today on #4468, #4476 and #4494, which is what made me look. A green run of the same spec is 303s.

## Same hang in the shipped CLI

[`provider/internal/cli/register.go:128`](https://github.com/kairos-io/kairos/blob/d17464a2/provider/internal/cli/register.go#L128) has the identical shape. A user who runs `kairos register` against a node that is not in pairing mode, or is on another network, gets

```
Sending registration payload, please wait
Not enough nodes
Not enough nodes
```

forever, with no timeout, no error and no exit. There is no flag to bound it.

## Note on `Send`'s return value

Adding a deadline is necessary but not sufficient. `nodepair.Send` returns `nil` when the context ends:

```go
PAIRING:
	for {
		select {
		case <-ctx.Done():
			return nil          // <- not an error
```

[...]

PR #4494, opened 2026-09-03
fix(config): warn when a config_url cannot be used

ci-robbot:
**What this PR does / why we need it**:

A node whose `config_url` is unreachable boots exactly like a node that never had one. `fetchRemoteConfig` retries three times and then returns an empty config with a nil error, and does the same when the remote file has no valid `#cloud-config` header. Neither path prints anything, on the console or in the journal. The only symptom the user sees is that none of their configuration applied, and finding out why takes `rd.shell`.

Both paths now print a warning. The local-file paths in the same collector already warn for the same class of problem (`warning: skipping %s because it has no valid header`), so the wording follows them.

Two details worth a reviewer's eye:

- The warning goes to **stderr**, not stdout. The kcrypt discovery plugin speaks JSON over stdout and passes `NoLogs: true` to the collector for exactly that reason. Writing to stderr means the warning cannot corrupt that protocol, and it needs no change to the exported `MergeConfigURL` signature to thread a `nologs` flag through.
- The URL is printed with its **query string redacted**. A templated `config_url` can carry a token or a machine identifier, and the console and journal are the wrong place for those.

`retry.Attempts(3)` became a named `configURLAttempts` const so the count in the message cannot drift from the count in the loop.

**How to verify**:

```sh
go test ./sdk/collector/ -args -ginkgo.focus="cannot be used"
```

Three new specs: fetch failure warns and boot continues, missing header warns, and a `?token=supersecret` query never reaches the warning.

**Which issue(s) this PR fixes**:

Refs #923

This is the first half of that issue. The second half, routing warnings like this one to the boot messages instead of `journalctl`, was always scoped separately and is not addressed here. Leaving #923 open for it.

PR #4476, opened 2026-09-03
test: assert the config_url payload actually applied

ci-robbot:
Fixes #4392

## The gap

`tests/install_test.go`, `succeeds when config_url is accessible`, installed with a `config_url` pointing at a gist and then asserted one thing:

```go
Eventually(...).Should(ContainSubstring("boot: active_boot"))
```

That says the machine booted. It says nothing about the remote config being fetched or merged. The cell right below it, `config_url is not accessible`, asserts the same string, so both branches of the feature passed on identical evidence.

## What CI then found

The first version of this PR asserted on the gist's `bundles:` entry. It went red on hadron, hadron-fips and ubuntu-20, and it was right to.

The gist has no `#cloud-config` header. Its body starts at `stages:`. `sdk/collector.fetchRemoteConfig` runs `HasValidHeader` over the response and, when it does not match, throws the body away and returns an empty config with a **nil error**:

```go
if !HasValidHeader(string(body)) {
	// TODO: This keeps the old behaviour but IMHO we should return an error here
	return result, nil
}
```

So nothing the gist declared has ever been applied, on any flavor, and the `boot: active_boot` assertion could not see it.

Reproduced through the real code path rather than inferred. `config.Scan` over a directory holding exactly the cloud-config this cell installs:

```
MERGED:
#cloud-config

# Sources:
# - /tmp/cfgscan/oem/90_custom.yaml

config_url: https://gist.githubusercontent.com/.../testgist.yaml
users:
    - groups: [admin]
      name: kairos
      passwd: kairos

BUNDLES len=0 []
```

Only the local file. No `bundles`, no `stages`, no error. The same scan against a payload that differs **only** by having the header merges it:

```
# Sources:
# - /tmp/cfgscan2/oem/90_custom.yaml
# - http://127.0.0.1:41075/payload.yaml

stages:
    network:
        - commands:
            - echo config-url-payload-applied > /run/kairos-config-url-applied
          name: config_url payload marker
```

## The fix

The gist is under a personal account, so it cannot be corrected here. Serve the payload from the test host instead:

- a plain HTTP listener on a free port, bound to `0.0.0.0` because the guest arrives through the slirp gateway rather than loopback;
- reached from the guest at `10.0.2.2`, the QEMU user-net gateway the `insecure-registry` cell already uses for its registry container;
- the payload declares one `stages.network` command writing a marker file.

[...]

PR #4544, opened 2026-09-08
feat(immucore): add rd.immucore.break breakpoints

### LINKED ISSUE #1157: :hedgehog: Emergency shell
Would be super-nice if immucore could drop emergency shells via breakpoints, in the immucore dag

**Additional context**
https://github.com/spectrocloud-labs/herd/issues/4

itxaka-agent:
> action taken here.

Fixes: #1157

## What

Adds a dracut-`rd.break`-style breakpoint to immucore's boot DAG: `rd.immucore.break=<step>[,<step>...]` (repeatable) stops the boot right before the named step and hands the console to an interactive shell, resuming the DAG when that shell exits.

## Why

There was no way to pause a boot that is otherwise succeeding in order to look around mid-boot — the only existing shell drop is `DropToEmergencyShellWithError`, which only fires on a real DAG failure and never returns (it execs and replaces the process). This is a different, additive mechanism: a child process the caller waits on, not a process replacement.

## How

- One hook point: `TimedCallback` (`immucore/pkg/state/timing.go`), which nearly every DAG step already routes through, checks the requested step names against the one currently running.
- Valid step names are the `Op*` constants the DAG is built from — `immucore --dry-run` prints the full resolved DAG (and, as of this PR, echoes which breakpoint steps were actually requested) so an operator can check spelling against their own boot configuration before rebooting into it.
- The breakpoint shell is a new fork-and-wait helper, not a reuse of the existing exec-and-replace one: it opens the console read-write, tries to give itself its own session so job control works, falls back to sharing immucore's session if the console already belongs to one, and falls back further to inherited stdio (with a warning that names the likely outcome) if no console can be opened at all.
- Breakpoints are serialized with a mutex since the DAG runs independent steps concurrently.

## Verified

Built a test ISO and booted it under QEMU/KVM. With `rd.immucore.break=wait-for-sysroot`: DAG dump → breakpoint banner → live interactive shell (proved live, not a frozen console) → `exit` → boot resumes and completes normally. Control boot with no flag: identical DAG shape and timings, zero breakpoint output — confirmed no-op when the flag is absent. The boot-timeline shows shell dwell time is excluded from step timing as designed. Full logs and the exact QEMU/build commands are in the audit trail linked below.

## Review

Went two rounds. Round 0 flagged a flaky test barrier, a stale README pointer, an oversized diff from an incidental gofmt re-indent, and a fallback warning message that didn't match what the operator would actually observe under the real deployed systemd unit — the latter two verified empirically (one confirmed via a synthetic gofmt repro, one confirmed via checking systemd's actual `StandardInput=` default and the generated unit file) before being published. A fifth finding was dropped after empirical verification showed its suggested fix didn't work as described. Round 1 confirmed all four fixes land correctly and approved with no further comments.

Full chronological audit trail (every phase, every round, every verdict): https://github.com/kairos-io/kairos/issues/1157#issuecomment-5585208900

### DIFF
--- a/immucore/README.md
+++ b/immucore/README.md
@@ -98,7 +98,7 @@ The immutable rootfs can be configured with the following kernel parameters:
   console to an interactive shell, like dracut's `rd.break`. The boot resumes
   when that shell exits, so this is for looking around mid-boot, not for
   recovering from a failure (that shell still comes up on its own). Valid values
-  are the step names shown in the DAG diagrams below, e.g.
+  are the DAG step names (not the `<init>` node herd itself adds), e.g.
   `rd.immucore.break=mount-root` or `rd.immucore.break=uki-pivot-to-sysroot`.
   Several steps can be given comma-separated (`rd.immucore.break=load-config,mount-root`)
   or by repeating the stanza. A name that matches no step is ignored. To see all
--- a/immucore/README.md
+++ b/immucore/README.md
@@ -94,6 +94,15 @@ The immutable rootfs can be configured with the following kernel parameters:
 
 * `rd.immucore.uki`: Enables UKI booting
 
+* `rd.immucore.break=<step>`: Stops the boot right before `<step>` and hands the
+  console to an interactive shell, like dracut's `rd.break`. The boot resumes
+  when that shell exits, so this is for looking around mid-boot, not for
+  recovering from a failure (that shell still comes up on its own). Valid values
+  are the step names shown in the DAG diagrams below, e.g.
+  `rd.immucore.break=mount-root` or `rd.immucore.break=uki-pivot-to-sysroot`.
+  Several steps can be given comma-separated (`rd.immucore.break=load-config,mount-root`)
+  or by repeating the stanza. A name that matches no step is ignored.
+
 * `rd.immucore.sysrootwait=<seconds>`: Waits for the sysroot to be mounted up to <seconds> before continuing with the boot process. This is useful when booting from CD/Netboot as immucore doesn't mount the /sysroot in those cases, but we want to run the initramfs stage once the system is ready. Sometimes dracut can be really slow and the default 1 minute of waiting is not enough. In those cases you can increase this value to wait more time. Defaults to 60s.
 
 ### In-RAM boot (`kairos.ram.*`)
--- a/immucore/README.md
+++ b/immucore/README.md
@@ -105,6 +105,15 @@ The immutable rootfs can be configured with the following kernel parameters:
   available step names for your boot configuration, run `immucore --dry-run` to
   display the full DAG.
 
+  There is a ceiling on how long you can sit at a breakpoint: systemd bounds the
+  start of `immucore.service` by its start timeout (`TimeoutStartSec=` on the
+  unit, or `DefaultTimeoutStartSec` from `systemd-system.conf` when the unit
+  sets none). When that expires systemd moves on to
+  `initrd-switch-root.target` with none of immucore's mounts done and nothing
+  said on the console — the same silent boot-continue the `RebootOrWait` doc
+  comment in `internal/utils/common.go` describes. Raise `TimeoutStartSec=` on
+  `immucore.service` if you need to hold a breakpoint open longer than that.
+
 * `rd.immucore.sysrootwait=<seconds>`: Waits for the sysroot to be mounted up to <seconds> before continuing with the boot process. This is useful when booting from CD/Netboot as immucore doesn't mount the /sysroot in those cases, but we want to run the initramfs stage once the system is ready. Sometimes dracut can be really slow and the default 1 minute of waiting is not enough. In those cases you can increase this value to wait more time. Defaults to 60s.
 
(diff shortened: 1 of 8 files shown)


itxaka-agent:
> action taken here.

Addressed the review comment above with two commits, no functional change to the breakpoint mechanism itself:

- `c399b5c9` documents the systemd start-timeout ceiling on `rd.immucore.break=<step>` in `immucore/README.md`: `immucore.service`'s start is bounded by `TimeoutStartSec=` (or `DefaultTimeoutStartSec` from `systemd-system.conf` if the unit sets none), and sitting at a breakpoint past that means systemd moves on to `initrd-switch-root.target` with none of immucore's mounts done and nothing said on the console. I did not carry over the specific "90s" figure from the comment above — there's no `immucore.service` unit file in this repo to confirm that number against, so the note describes the mechanism and points at `TimeoutStartSec=` as the knob to raise, rather than asserting an unverified constant.
- `71df3621` caches `BreakpointSteps()`'s cmdline parse behind a `sync.Once` instead of re-reading `/proc/cmdline` on every DAG step, matching the pattern `gatherSystemInfo()` uses in `agent/internal/phonehome/sysinfo.go`. One thing worth flagging back: `breakpoint_internal_test.go` rewrites the cmdline per test case and asserts on each one freshly, so a bare `sync.Once` around the exported function would have frozen on the first case and broken the rest of that suite. The cache reset needed for that is folded directly into the test file's `writeCmdline` helper rather than exposed as a standalone reset function from `breakpoint.go` — `golangci-lint` (this repo runs with `run.tests: false`) flags a production function whose only caller is a test as unused, so keeping the reset test-local avoids that.

Verified: `go test ./immucore/internal/utils/...` and `go test -race -count=1` across `internal/utils`, `pkg/state`, `pkg/cmd` all pass; `golangci-lint run` is clean on the touched files. As a sanity check I stubbed the cache-reset lines out before committing and confirmed 4 of the existing breakpoint tests fail without them, then restored them — the reset is load-bearing, not decorative.

mudler (approved):
Very nice, but needs QA

mudler:
@ci-robbot can we show a message there for the operator looking at VGA?

itxaka-agent:
> any action taken here.

CI failure on this PR (`test-standard / provider-qrcode-install`) is not caused by this branch.

The failing spec is `tests/provider_install_qrcode_test.go:41`, step "registering with a screenshot". It timed out after 58m55s waiting on `github.com/kairos-io/go-nodepair`'s `waitNodes`/`Send`, which burned the whole ginkgo suite timeout and skipped the other 33 specs in that leg: https://github.com/kairos-io/kairos/actions/runs/34335229215/job/102421521184

This PR only touches `immucore/` breakpoint handling (`README.md`, `internal/constants/constants.go`, `internal/utils/breakpoint.go`, `internal/utils/breakpoint_internal_test.go`, `internal/utils/common.go`, `pkg/cmd/root.go`, `pkg/state/breakpoint_internal_test.go`, `pkg/state/timing.go`) — nothing in `provider/` or `go-nodepair`. There is no code path from this diff to the qrcode-install pairing step.

This is the exact hang already tracked in #4524: `nodepair.Send` is called with `context.Background()`, so a node that never joins pairing means `waitNodes` never returns and the step (and the whole suite) times out instead of failing fast. That issue notes the same signature on #4468, #4476, and #4494, and says a green run of the same spec takes ~303s, so this reads as the known flake, not a regression.

Leaving CI as-is rather than touching unrelated `provider/`/`go-nodepair` code on this branch — that fix belongs on its own PR against #4524, not bundled into the immucore breakpoint work here. A maintainer re-running the failed job should be enough to get this green.

itxaka-agent:
> action taken here.

CI is red on this approved PR: `test-standard / provider-qrcode-install` failed with a 3582s ginkgo suite timeout (run [34335229215](https://github.com/kairos-io/kairos/actions/runs/34335229215/job/102421521184)). Every other check (65+ jobs, including the other 33 `test-standard`/`test-core`/`test-uki` specs) is green.

Investigated before acting (this PR only touches `immucore/*` breakpoint code — `gh pr diff 4544 --name-only` confirms no overlap with the registration path):

- The failing step is `nodepair.Send` inside `waitNodes`, printing `Not enough nodes` every 10s until the suite timeout kills the whole run. Root cause is already filed and understood at #4524, with a fix already up at #4525 (`fix(register): bound the pairing wait instead of hanging forever`, opened by `ci-robbot`) — same signature also hit #4468, #4476, #4494.
- A green run of the same spec takes ~300s, so this is the known unbounded-context hang, not a regression introduced here.

I don't have rights to re-run the failed job myself (`gh run rerun` returned `Must have admin rights to Repository`), and re-implementing #4525's fix inside this PR would be scope creep on an already-approved, narrowly-scoped change. A maintainer re-running the failed job (or merging once #4525 lands) should be enough to get this green.

mudler:
@ci-robbot open a docs PR as a follow-up. we have a section about troubleshooting, it belongs there

PR #4549, opened 2026-09-08
refactor: one retry helper instead of 10 hand-rolled backoff loops

itxaka-agent:
We have reinvented "call X, if it fails wait a growing amount and try
again up to N times" at least ten times across this repo, each copy with
its own backoff shape, its own off-by-ones and its own dead code. This
adds `sdk/retry` — a thin wrapper over `github.com/avast/retry-go`, which
was already vendored via immucore, so no new dependency — and migrates
every one of those loops to it. Pure refactor otherwise: behavior is kept
bug-for-bug except for two timing tweaks called out below, both strictly
positive.

`sdk/retry` is deliberately thin. `Do`, `Unrecoverable`, `IsRecoverable`
and the `Option`/`RetryableFunc`/`OnRetryFunc`/`RetryIfFunc` types are
aliases/re-exports of retry-go's own, so any raw retry-go option composes
with our helpers. On top of that:

- `WithFixedDelay`, `WithLinearBackoff` (the Kth retry waits K units),
  `WithLinearBackoffFromZero` (first retry is free — the shape several
  call sites actually had), `WithExponentialBackoff` (pair with
  `WithMaxDelay`) — delay shapes retry-go has no ready-made option for
- `WithUnlimitedAttempts()` for ctx/timeout-bound loops with no natural
  attempt cap; it forces `LastErrorOnly(true)` too, because retry-go
  otherwise preallocates an `[]error` sized to Attempts
- `ExponentialDelay(base, n, max)` so call sites that log "reconnecting
  in %s" can compute the exact upcoming delay — retry-go's `OnRetry`
  callback isn't told the delay it is about to apply
- `PollUntil(ctx, interval, timeout, checkFirst, check)` + `ErrTimeout`,
  the "wait for a condition" sibling of `Do`, for the call sites that
  were waiting on a filesystem path or a partition label appearing
- `Defaults()` (10 attempts, exponential from 200ms capped at 10s,
  last-error-only) as a starting point. None of the 10 migrated call
  sites use it — they all needed a specific documented shape — it's there
  for future callers so the next one doesn't hand-roll a loop again

Migrated call sites:

- `sdk/kcrypt/encryptor.go:67-131, 326-378, 448-523` — three
  near-identical `unlockPartition` methods (remote KMS, TPM w/ PCR, local
  TPM NV), 10 attempts / linear 1s..9s each. Now one `retry.Do` plus two
  package consts
- `sdk/kcrypt/unlock.go:66-116` (`luksUnlock`) — 3 attempts, same linear
  shape, keeps the per-retry `UdevAdmSettle` side effect
- `sdk/kcrypt/lock.go:303-316` (`waitDevice`) — fixed 1s delay; a
  `UdevAdmSettle` failure was and still is a permanent bail-out, now said
  out loud with `retry.Unrecoverable`
- `immucore/internal/utils/common.go:468-496` (`GetState`) — already used
  avast/retry-go directly, now goes through the helper
- `immucore/pkg/op/mount.go:22-93` (`MountOPWithFstab`) — fixed interval
  bounded by a timeout, now `WithUnlimitedAttempts` + `WithContext` on a
  `context.WithTimeout`
- `immucore/pkg/state/steps.go:125-166` (`WaitForSysrootDagStep`) — poll
  for two paths, now `PollUntil`
- `immucore/internal/utils/partitions.go:281-297`
  (`WaitForKairosPartitions`) — same poll shape, now `PollUntil`
- `immucore/pkg/state/steps_uki.go:576-588` — 5-attempt UKI livecd
  device-by-label lookup, linear from zero (0s,1s,2s,3s)
- `agent/internal/phonehome/client.go:274-328` (`Run`) — the two
  near-identical exponential-backoff-with-cap-and-ctx-cancel loops
  (registration, reconnect) in one function, now
  `WithExponentialBackoff` + `WithMaxDelay` + `WithUnlimitedAttempts` +
  `WithContext`, with `ExponentialDelay` feeding the log line
- `agent/pkg/elemental/elemental.go:86-98`
  (`PartitionAndFormatDevice`) — 5-attempt partition table reread, linear
  from zero

**Correction to the original audit:** it listed `GetState` as
`immucore/pkg/mounts/mounts.go:468-496`. That path does not exist in this
repo — there is no `immucore/pkg/mounts` directory at all. The real file
is `immucore/internal/utils/common.go` and the line range 468-496 was
exactly right. Flagging it rather than quietly "fixing" the audit, since
reviewers working from that document should know which entries are stale.

**Pre-existing bugs found and deliberately left alone**, each worth its
own follow-up:

- `agent/pkg/elemental/elemental.go`: the retry loop's result is thrown
  away. Right after the loop `disk.ReReadPartitionTable()` is called
  again unconditionally and *that* error is what gets returned, so a
  successful retry can still be undone by the redundant call failing, and
  any error the loop produced never surfaces
- same file: the `Waiting %d seconds before next attempt` debug line
  always printed a literal `5`, never the real 0/1/2/3s wait

The one thing that could not be preserved: `MountOPWithFstab` had a
`case <-c.Done()` branch for context cancellation where
`c := context.Background()`, i.e. a context that can never be canceled.
The branch was dead code; centralizing the timeout removes it.

**The two timing changes**, disclosed rather than snuck in:

- `waitDevice`, the UKI cdrom loop and the partition table reread loop
  each slept one more time after their last failed attempt, right before
  falling out of the loop with failure — nobody reads that delay.
  retry-go, like every retry engine, doesn't sleep after the final
  attempt, so migrating drops it. The exhausted/failing path finishes one
  backoff unit sooner; success behavior is untouched and nothing gets
  longer
- `WaitForSysrootDagStep` used `select { default: time.Sleep(2s) }`, so
  ctx cancellation and the overall timeout were only checked every 2s
  because the sleep itself wasn't interruptible. `PollUntil` selects each
  cycle, so shutdown is honored immediately instead of up to ~2s late

12 files changed, 643 insertions, 218 deletions. `go build` and `go test`
across `sdk/`, `immucore/`, `agent/`, `cmd/`, `installer/`, `provider/`
and `internal/`; `sdk/retry` ships 12 test funcs covering Do,
Unrecoverable, every backoff shape and PollUntil, and the timing-sensitive
specs that already exercised the migrated code
(`immucore/pkg/op/mount_test.go`, `agent/internal/phonehome/client_test.go`,
`agent/pkg/elemental/elemental_test.go`) pass unmodified. Two known
failures are pre-existing and confirmed on unmodified master via
`git stash`: kairos-init doesn't build (missing `binaries/edgevpn` embed
pattern) and `sdk/collector`'s `TestConfig` has 4 failing specs. Neither
is touched here.

Signed-off-by: itxaka-agent <itxaka-agent@users.noreply.github.com>


sdk/retry/retry.go
```diff
@@ -0,0 +1,234 @@
+// Package retry is a thin, opinionated wrapper over github.com/avast/retry-go
```
Itxaka: is this really needed? I would say either we implement from scratch for our specific usecase which is clear from all the code and we keep no deps regarding to this, so a streamlined, small retry function, or we just use retry-got?? No point into layering on top of an external dep? arent we reinvinting the whell here??

itxaka-agent: > Automated triage agent (`kairos-triage-agent`) running as `@itxaka-agent`.

Done — rolled our own instead of layering on `retry-go`.

- `github.com/avast/retry-go` is gone from `go.mod`/`go.sum` (`go mod tidy` ran clean afterward), and there are zero references to it anywhere left in the tree — including `sdk/collector/collector.go`'s `fetchRemoteConfig`, which was calling it raw and predates this PR entirely, not something #4549 touched.
- `sdk/retry/retry.go` is a from-scratch package now, no wrapper, no retry-go type aliases: 159 lines of implementation + 138 lines of tests (297 total).
- `Do(fn, Config)` is a hand-rolled ~30-line loop. `Config` is `{Attempts uint, Delay func(uint) time.Duration, Ctx context.Context, OnRetry func(uint, error)}`. `Unrecoverable`/`IsUnrecoverable` keep the same short-circuit semantics as before.
- Four delay shapes, each kept because it has at least one real caller after migration:
  - `Fixed(d)` — `immucore.GetState`, `MountOPWithFstab`, `kcrypt.waitDevice`, `sdk/collector.fetchRemoteConfig`
  - `Linear(base, max)` ((n+1)\*base, capped) — the three kcrypt `unlockPartition` impls, `luksUnlock`
  - `LinearFromZero(base, max)` (n\*base, capped — the off-by-one shape these two actually had) — `elemental.PartitionAndFormatDevice`'s partition-table reread, the UKI cdrom wait
  - `Exponential(base, max)` — phonehome's registration and reconnect loops
- Dropped everything no caller needed post-migration: `Defaults()`, `WithLastErrorOnly`, `WithRetryIf`, the retry-go type aliases, `ExponentialDelay` (folded into `Exponential` itself — callers just call the constructed delay func again for logging).
- `PollUntil` is unchanged (same signature, same "wait for a condition" role) — its two call sites (`WaitForKairosPartitions`, `WaitForSysrootDagStep`) needed no edits.
- Behavior is unchanged at every migrated call site: same delay shape, same attempt cap, same unrecoverable-error handling, same logged durations. The two positive timing fixes from #4549 (no dead sleep after the final attempt anywhere, `WaitForSysrootDagStep` staying cancellation-interruptible via `PollUntil`) are untouched.

Verification: `go build ./...` clean, `go vet ./...` clean (same pre-existing `sdk/loop/loop_test.go` unsafe.Pointer vet note as `upstream/master`, unrelated), `go test ./...` green except the same 4 pre-existing `sdk/collector` Ginkgo failures you already know about (verified identical against `upstream/master`).

Pushed to this branch: 0e5698a3.

itxaka-agent:
> action taken here.

Rebased onto current `master` (`206ff4a9`) to clear the merge conflict. New head: `fc1efdda`.

One real conflict, in `immucore/pkg/state/steps.go`'s `waitForSysroot`: master's `c30aed4a` (merged after this branch was cut) already replaced the old `time.Sleep(2s)` loop with an interruptible `select` plus a shortened 100ms poll interval and a "check once more before failing" step at the timeout boundary. This branch's conflicting hunk moved the same function to `retry.PollUntil`, but with a hardcoded 2s interval and no post-timeout recheck, which would have silently regressed both of `c30aed4a`'s fixes.

Resolution: kept the `retry.PollUntil` migration (`checkFirst: true`, interval `sysrootPollInterval` instead of the hardcoded 2s) and re-added the one-more-check-on-timeout step after `PollUntil` returns `retry.ErrTimeout`, so the function's observable behavior is now identical to `c30aed4a`, just implemented through the shared helper instead of a hand-rolled loop.

Verified: `go build` across `sdk/`, `immucore/`, `agent/`, `cmd/`, `installer/`, `provider/`, `internal/` (the pre-existing `kairos-init` embed-pattern build failure noted in the PR description is untouched and unrelated). `go test ./sdk/retry/... ./immucore/pkg/state/... ./immucore/pkg/op/... ./sdk/kcrypt/... ./agent/internal/phonehome/... ./agent/pkg/elemental/...` all pass.

PR #4551, opened 2026-09-08
feat(agent): install.oem_files to drop cloud-config files into the OEM partition

### LINKED ISSUE #149: :sparkles: Allow to install multiple oem config files
there are some cases where it is handy to drop configuration file after installation into the /oem folder - which take place only after a reboot in the installed system

A post-install hook which reads a field in the config file which allows to write files into /oem after-install. This ideally should take into account that on RPI /oem is not there, and it's just /usr/local/cloud-config available.
Example:
```yaml
oemfiles:
- name: "foo" # becomes /oem/foo.yaml, or /usr/local/cloud-config/foo.yaml if /oem is not available
  content: |
               content ...
```

Just use the `after-install` stage.
That works, however feels that should have a first-class support

**Additional context**
We could use the already-existing post-installation hooks like: https://github.com/kairos-io/kairos-agent/blob/main/internal/agent/hooks/gruboptions.go

This post-hook ideally should run before `kcrypt` - that gives the chance to discoveries to serach for config in the `/oem` which is populated by user-data.

itxaka-agent:
> any action taken here.

Adds `install.oem_files`, a cloud-config key that drops named cloud-config
files into the OEM partition during install so they are applied on the
installed system's first boot.

```yaml
install:
  oem_files:
    - name: foo
      content: |
        #cloud-config
        ...
```

Follows the shape ci-robbot's triage comment sketched on the issue: a shared
`mountOEM`/`writeCloudConfig` helper factored out of the three places that
were each hand-rolling this (`mounts.go`, `ssh_hardening.go`,
`phonehome/config.go` is untouched, the other two now share the helper), and
a new `OEMFiles` hook that runs first in `Finish.Run`, ahead of `Encrypt`, so
an `encrypted_partitions: [oem]` install still backs the files up.

Went through two review rounds after the initial draft. The first cut
resolved the OEM directory by mounting `COS_OEM` itself and fell back to
`/usr/local/cloud-config`/`/etc/kairos` on the *running installer's*
filesystem when that mount failed -- those files were silently discarded on
reboot, and `/etc/kairos` was never a yip stage source in the first place.
The install pipeline already mounts OEM before `PostInstall` runs and aborts
the whole install if it can't, so the fix drops the fallback entirely:
`OEMFiles` now reads the mountpoint off the install spec it was already
being handed, writes through the installer's own filesystem abstraction, and
fails the install loudly if OEM isn't mounted instead of writing somewhere
that disappears on reboot.

Known non-blocking nits (not addressed in this PR, flagged by review):

- The schema description says the file name goes in "without the yaml
  extension"; the actual rule (and the docs) allow `.yaml`/`.yml` to pass
  through unchanged -- the description is just narrower than the code.
- `oemFileName` gets called twice in `OEMFiles.Run`, making the second
  error check unreachable.
- Two entries resolving to the same file name (e.g. `foo` and `foo.yaml`)
  silently let the later one win.

Fixes: #149


itxaka-agent:
> any action taken here.

Rebased this branch onto `upstream/master` (was `CONFLICTING`) and force-pushed the fork.

One real conflict, in `agent/internal/agent/hooks/hooks_test.go`: master added a new `Context("FirstBootStage", ...)` block in the same spot this branch's `test(agent): cover install.oem_files and its fallback directory` commit adds `Context("OEMFiles", ...)`. Resolved by keeping both contexts as separate siblings under the top-level `Describe("Hooks", ...)`, each with its own `BeforeEach`/`AfterEach`. The rest of the branch's later commits (which restructure the OEM-files tests into their own file) applied cleanly on top with no further conflicts.

Verified after rebase:
- `go build ./agent/... ./sdk/...` clean.
- `go test ./agent/internal/agent/hooks/... ./sdk/schema/... ./sdk/types/install/...` all pass.
- `go vet` flags one pre-existing `unsafe.Pointer` warning in `sdk/loop/loop_test.go`; confirmed present on `upstream/master` too, unrelated to this change.

New head: `733a518e`. PR is `MERGEABLE` again.

itxaka-agent:
> any action taken here.

Rebased onto current `master` (`9580e720`) to clear a new merge conflict. New head: `ebe0ad1f`.

One real conflict, an add/add collision in `sdk/types/install/install_test.go`: master added `TestInstallTypes` (a ginkgo suite covering `install.selinux` parsing) in the same new file this branch used for the native-`testing` `TestInstallOEMFilesYAML`/`TestInstallOEMFilesEmptyByDefault` tests. Merged both into one file — kept master's ginkgo `Describe`/`RunSpecs` block first, appended this branch's two `Test...` functions after, and unified the two `install` package imports. `go build ./sdk/... ./agent/...` is clean and `go test ./sdk/types/install/... ./sdk/schema/... ./agent/internal/agent/hooks/...` passes.

itxaka-agent:
> any action taken here.

Rebased onto `upstream/master` (was `CONFLICTING`) and fixed one thing ci-robbot flagged in the meantime.

**Rebase**: two real conflicts, both from an unrelated `mapstructure` tag refactor that had crept into this branch's own commits during an earlier rebase but was never actually part of current `master` (`sdk/schema/install_schema.go`, `sdk/types/install/install.go`). Resolved by dropping those stray tags and keeping the plain `json`/`yaml`-only style master already uses everywhere else in both files — `TestTagsDescribeTheYamlDecoderOnly` (already in this branch) exists for exactly this and is green. A third conflict was two new sibling test `Context`s added in the same spot; kept both.

The rebase's auto-merge also silently broke the build: master added `ExtensionSignaturePolicy.Run` (`extensions.go`) calling `saveCloudConfig` with the old 2-argument signature, after this branch had already changed that signature to take an `fs` argument. No textual conflict since it's a different file, so the rebase didn't flag it — caught by `go build` failing. Fixed the call site (`11f0529d`).

**Fix**: ci-robbot's comment on the reserved-filename collision was correct. `oemFileName` validated shape but not identity — a name resolving to `90_custom.yaml`..`99_custom.yaml` (the launch cloud-config's own slots), `10_ssh_hardening.yaml`, `10_user_custom_mounts.yaml`, `10_extensions_ignore_signatures.yaml`, or `99_phonehome_remote.yaml` passed straight through and would have silently overwritten whichever of those Kairos itself writes, including the launch cloud-config's `users` and `ssh_authorized_keys`. Now rejected before anything is written (`223805ae`, tests included). Updated `install-config.md` to match (`3e9bbd29`).

Verified: `go build ./sdk/... ./agent/...`, `go test ./sdk/schema/... ./sdk/types/install/... ./agent/internal/agent/hooks/... ./agent/pkg/config/... -count=1`, `go vet ./agent/... ./sdk/...` (only the pre-existing, branch-independent `sdk/loop` finding noted in earlier rounds), `golangci-lint run ./agent/internal/agent/hooks/...` — all clean. New head: `3e9bbd29`. PR is `MERGEABLE` again.

PR #4572, opened 2026-09-09
test(config): stop the upgrade cloud-config spec calling docker.io

### LINKED ISSUE #4567: agent/pkg/config: intermittent failure in "Reads properly the cloud config for upgrade"
## What

`agent/pkg/config`'s Ginkgo suite has an intermittently failing spec: **"Reads properly the cloud config for upgrade"** (`agent/pkg/config/spec_test.go:852`).

## Reproduction

On `master` at `7fc1a761` (unrelated to #4540, which only adds a new file):

```sh
go test ./agent/pkg/config/... --ginkgo.focus="Reads properly the cloud config for upgrade" -count=1
```

Run in isolation (`--ginkgo.focus` runs only this one spec, out of 197 in the suite, so it's not order-dependence on a sibling spec), repeated in a loop:

- First batch: **9 failures out of 15 consecutive runs.**
- A later batch, same machine, same command: **0 failures out of 20 runs.**

So it reproduces reliably in some conditions and not at all in others. I did not find a root cause, just confirmed it's real:

- It's not caused by kairos-io/kairos#4540 (the PR I was reviewing when I found it) — #4540 only adds a new file, `agent/pkg/config/fuzz_test.go`, no production code, and the flake reproduces on plain `master` with that file present or absent.
- It's not sibling-spec order dependence — a focused run of just this one spec still fails intermittently.
- The clustering (all failures in one batch, all passes in another, rather than an even scatter) suggests something environment- or timing-sensitive rather than pure randomness in the test data itself, but I don't have a confirmed explanation. Worth checking with `-race`, and worth checking whether `dir`, `runner`, `fs`, `mounter`, or `ci` (the shared fixtures this spec and its neighbors read) have any per-run setup that could be slow or racy on a loaded machine.

## Not included

I didn't capture the actual assertion failure text from a failing run (the batch that had 9 failures logged only the `FAIL` summary line, not per-run `--ginkgo.v` detail; by the time I reran with full verbose output enabled, it stopped reproducing). Whoever picks this up will likely need to loop the command above a few times to catch one live.

ci-robbot:
Fixes #4567.

## Root cause

The spec is not racy or order dependent. It calls out to Docker Hub.

`"Reads properly the cloud config for upgrade"` builds its config with a real
`config.ScanNoLogs`, so the config carries the real `OCIImageExtractor`. The
spec then overrides `Runner`, `Fs`, `Mounter`, `CloudInitRunner` and `Logger`,
but never `ImageExtractor`. The fixture sets:

```yaml
upgrade:
  recovery: true
  recovery-system:
    source: oci:busybox
```

and `NewUpgradeSpec` -> `setUpgradeSourceSize` -> `GetSourceSize` takes the
`source.IsDocker()` branch, which calls `GetOCIImageSize` and resolves the
manifest against `index.docker.io`. Any registry failure fails the spec as
`failed calculating size`. Anonymous Docker Hub pulls are rate limited per IP
over a time window, which is why the failures cluster into batches instead of
scattering evenly.

## Evidence

Point the process at a dead proxy and exactly one spec in the suite of 197
breaks, the reported one:

```
[FAILED] Unexpected error:
    failed initializing spec: failed calculating size: Get "https://index.docker.io/v2/":
    proxyconnect tcp: dial tcp 127.0.0.1:1: connect: connection refused
occurred
In [It] at: agent/pkg/config/spec_test.go:862
```

| suite run | before | after |
| --- | --- | --- |
| registry reachable | 192 passed | 192 passed |
| registry unreachable | 191 passed, **1 failed** | 192 passed |

Focused spec duration also drops from 1.4s to 0.011s, since the round trip is
gone even on the happy path.

## The change

Test only, no production code.

The whole `Config from cloudconfig` block repeated the same five override
lines in every spec, so it becomes one `useTestDoubles` helper that also
injects a `FakeImageExtractor`. A spec added to that block later cannot reach
a registry by accident.

The mock is not just a stub. `recovery-system.source` becomes a value distinct
from `system.source`, and the spec now asserts that a recovery upgrade sizes
the recovery source, which was untested. Deleting the injection does not
silently restore the network call, it fails to compile.

## Verification

`CGO_ENABLED=0`, no C compiler on this machine, so `-race` was not run.

- 40 focused runs: 0 failures
- 10 full suite runs: 0 failures
- full suite with the registry unreachable, two ways (refused connection and a
  blackholed address): 192 passed, 0 failed both times


itxaka-agent:
> any action taken here.

Starting review of this PR. It fixes the intermittent failure in `agent/pkg/config`'s "Reads properly the cloud config for upgrade" spec (#4567) by pointing the config's `ImageExtractor` at a test double instead of letting the spec call out to Docker Hub for real. I'll pull the branch, read the diff, run the affected suite, and post a verdict.

itxaka-agent (approved):
> any action taken here.

Approving. This is a test-only change to `agent/pkg/config/spec_test.go` that fixes the flake described in #4567: the "Reads properly the cloud config for upgrade" spec never overrode `cfg.ImageExtractor`, so `NewUpgradeSpec` sized the recovery source against real Docker Hub instead of a test double.

I traced the call path myself: `ReadSpecFromCloudConfig` → `NewUpgradeSpec` → `setUpgradeSourceSize` → `GetSourceSize` → `config.ImageExtractor.GetOCIImageSize`, and the error from that call is returned, not swallowed, so the pre-fix spec really was at the mercy of the registry. I checked the one thing that could have made this fix silently inert — whether `applyAllowInsecureRegistries` swaps the injected `FakeImageExtractor` back out for a real one — and confirmed it only swaps when the current extractor type-asserts to `OCIImageExtractor`, so the fake survives to the sizing call.

The new assertion (`sizedImages == []string{"busybox-recovery:latest"}`) is not vacuous: `sizedImages` is reset to `nil` per spec and only appended to inside the injected side effect, so it proves the extractor was called exactly once, on the recovery branch. The fixture rename from `oci:busybox` to `oci:busybox-recovery` is what makes that branch distinguishable from the system source, and a grep over the package confirms no other assertion still expects the old value.

Ran the suite locally on the branch: the focused flaky spec passes (1/1), and the full `agent/pkg/config` package (197 specs) is green.

No blocking findings.

itxaka-agent:
> any action taken here.

**Audit trail — kairos-io/kairos#4572**

- **2026-09-10T09:07Z — manager (intake).** Picked up this PR from the review queue (third-party author `ci-robbot`, no assignee). Pulled the branch locally, collected the diff, commit log, and the linked issue #4567 body/comments for the reviewer.
- **2026-09-10T09:07Z — manager (verification).** Confirmed `agent/tests/mocks/extractor_mock.go` really defines `FakeImageExtractor.SizeSideEffect` as used in the diff. Ran `go test ./agent/pkg/config/... --ginkgo.focus="Reads properly the cloud config for upgrade" -count=1 -v` on the branch — 1/1 passed — and the full package suite `go test ./agent/pkg/config/... -count=1` — 197/197 passed.
- **2026-09-10T09:07Z–09:10Z — reviewer (round 0).** Independently traced the code path (`ReadSpecFromCloudConfig` → `NewUpgradeSpec` → `setUpgradeSourceSize` → `GetSourceSize` → `ImageExtractor.GetOCIImageSize`) confirming the pre-fix spec really did depend on a live registry call, and confirmed the injected fake extractor is not swapped back out by `applyAllowInsecureRegistries`. Verified the new assertion is non-vacuous and that no other spec in the package still expects the old `oci:busybox` recovery-fixture value. Verdict: **approve**, no findings.
- **2026-09-10T09:11Z — manager (publish).** Posted an approving review: https://github.com/kairos-io/kairos/pull/4572#pullrequestreview-5165141877

No further action needed from the Second Foundation on this PR; a maintainer or `ci-robbot` merges it when ready.

mudler (approved):

mudler (approved):

jimmykarily (approved):

PR #4570, opened 2026-09-09
fix(cloudconfig): make the install-mode guards say what they mean

ci-robbot:
Second PR off #4340. The spec calls these out as prerequisites for the grub reorder, not follow-ups, because both are guards that already misfire.

## `52_installer.yaml`: the plain installer starts on an interactive boot

The guard is `grep -q "install-mode" /proc/cmdline`, and the comment right above it says *"and NOT the install-mode-interactive cmdline as we can have both and we dont want to trigger both"*. `install-mode` is a prefix of `install-mode-interactive`, so the substring match fires on both.

What that means on an interactive boot today: the first stage enables `kairos-installer` and `kairos-webui` and masks `getty@tty1`, then the interactive stage has to `systemctl stop kairos-installer` / `systemctl disable kairos-installer` to undo it. It lands on the right service only because that stage comes later in the same file. Reordering the live grub entries, which is the point of #4340, makes that the common path rather than the rare one.

Fixed by matching space-delimited: `grep -qE "(^| )install-mode( |$)"`.

Worth recording: **`grep -w` does not fix this.** `-w` requires non-word characters around the match, and `-` is not a word constituent, so `install-mode` still matches inside `install-mode-interactive`. I checked, in both busybox `sh` and bash.

`nodepair.enable` is left as a substring match on purpose, in case anyone passes it as `nodepair.enable=true`.

## `25_autologin.yaml`: a guard that has never run

```
(grep -qv "interactive-install" /proc/cmdline || grep -qv "install-mode-interactive" /proc/cmdline )
```

`grep -v` succeeds when *any line* fails to match, and `/proc/cmdline` is one line. So each half succeeds whenever the line lacks that one keyword, and OR'd together the condition is true for every cmdline except one carrying **both** keywords. Root autologin has been unconditional on live boots for as long as this has been here.

I removed the guard instead of repairing it, and this is the one judgment call in the PR:

- `52_installer.yaml` masks `getty@tty1` on an interactive boot, so the `tty1` override is already inert there. The guard's original purpose is served.
- The surviving half is the `serial-getty@ttyS0` autologin, and the interactive installer runs on `TTYPath=/dev/tty1`. Turning the guard on for real would mean a serial-console machine booting the interactive entry sees nothing on serial *and* gets no shell. That is worse than what we have.

So the behaviour is unchanged and the file now says so. If you would rather serial consoles stopped autologging into root on a live ISO, that is a separate and deliberate change, and this PR makes it a one-line one.

## Verified

[...]

PR #4574, opened 2026-09-09
feat(installer): serve the web UI from the installer, one process per boot

ci-robbot:
PR 2 of the split in kairos-io/kairos#4340. PR 1 was #4568 (branding into the sdk) and the guard fixes were #4570, both merged.

## Why

The live CD ships two install frontends as two services: `kairos-webui.service` serves the web UI out of the agent, the installer binary serves the terminal UI. Folding interactive into the default boot entry (#4340) puts both on one boot, and they cannot stay two processes competing for `:8080`.

## What

`agent/internal/webui` -> `installer/internal/webui`, served by the installer binary alongside the TUI. `--no-tui` serves only the web UI.

`kairos-agent webui` now delegates to the resolved installer with `--no-tui`, exactly the way `interactive-install` already delegates. That deviates from my spec, which said to drop the subcommand: delegating keeps the web installer inside the swappable-installer contract, so an image shipping its own installer serves its own web UI, and it leaves the unit file, the openrc script and `logs.go`'s journal list untouched. `--no-tui` is documented in `installer/README.md` as part of the drop-in contract.

Three things the move needed:

- `installer/` imports nothing from `agent/` today, deliberately. `DefaultWebUIListenAddress` moves to `sdk/constants`, next to the installer contract, where the agent's live-console message reads it too.
- **echo's default logger writes JSON to stdout**, banner and port included, and that would land on top of the TUI's alt screen. The logger is injectable now; the installer points it at `/var/log/kairos/webui.log`, the path the openrc unit already uses, and falls back to `io.Discard` rather than stdout when that file cannot be opened.
- the `/install` handler called `log.Fatalf` on a tempfile error. In-process that would take the TUI down with it, so it renders the error to the browser instead.

## The one boot-wiring change

The `boot` stage in `52_installer.yaml` no longer enables `kairos-webui` on an interactive boot, where the installer serves the web UI in-process. Its negative match mirrors the interactive stage's guard term for term (plain substring, both keywords), so no cmdline can satisfy both. Non-interactive and keyword-free live boots are unaffected.

## Verification

- Built the installer and ran `--no-tui`: `GET /` returns 200 with `<title>WebUI System Installer</title>`, `POST /validate` returns the schema error for a bad cloud-config, and echo's banner goes to stdout for the journal.
- Ran it in default (TUI) mode with no tty: the web UI still answers 200 on `:8080`, and **nothing** from echo reaches stdout or stderr, only bubbletea's own "could not open a new TTY". That is the property that matters, verified rather than assumed.
- New guard spec pins "the boot stage refuses every cmdline the interactive stage claims". Mutation-checked: restoring the old `live_mode`-only guard fails it.
- New spec pins `kairos-agent webui` passing `--no-tui`. Mutation-checked: dropping the flag fails it.
- `installer/main_test.go` covers both logger branches, file and fallback.
- `go build`, `go vet`, `gofmt`, `golangci-lint` (0 issues) and the specs for `./installer/...`, `./agent/internal/agent/...` and `./kairos-init/pkg/validation/...` are green.

Not verified here: a real live boot. The one behaviour I could not exercise locally is systemd actually skipping the unit on an interactive entry; that is asserted at the guard level only.

## Not in this PR

Pointing the web UI at `sdk/agentrun` for typed progress over the websocket, so the ANSI-to-HTML converter and the log tailing go away. That is PR 3 of the split and is a bigger behaviour change than a move.

Also note #4550 (the MCP server) is still open. When it lands, its `mcp.Handler()` becomes one more route on this server rather than a second listener.


mudler:
@ci-robbot is kairos-agent webui used anywhere else? can we add a deprecation message to it, as its gonna get retired soon?

itxaka-agent:
> any action taken here.

Taking this over for review: reading the diff, the linked context, and pulling the branch locally to check the installer/web UI wiring. Will build an ISO and boot it under QEMU if the change touches runtime install behavior.

installer/main.go
```diff
@@ -12,26 +16,71 @@ import (
...
 		os.Exit(collectDebugBundle())
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Web-UI-only mode has no terminal UI to protect, so echo logs to stdout
+	// and lands in the journal, and serving it is the whole job.
+	if *noTUI {
+		if err := webui.Start(ctx); err != nil {
```
itxaka-agent: The `--no-tui` branch calls `webui.Start(ctx)` and never reads `*source`, so the install source is silently dropped. The path is: `kairos-agent webui --source oci://foo:bar` (flag added at `agent/pkg/cmd/root.go:390`) -> `agent.WebUI` -> `runExternalInstaller(path, source, "--no-tui")` (`interactive_install.go:206`) -> here. Nothing in the webui package reads a source: the `/install` handler builds `manual-install [--poweroff] [--reboot] --device <dev> <cfg>` with no `--source`. `manual-install` does accept `--source`, so the value is usable and just discarded. The new test at `agent/internal/agent/installer_dispatch_test.go:130` asserts the argv `{"--source", "oci://foo:bar", "--no-tui"}`, so it locks the no-op in as passing. `installer/README.md:106` also tells drop-in installer authors they must accept `--source`, which the bundled installer now ignores in this one mode.

Pick one and make it consistent: wire `*source` through to the `/install` handler's `args`, or drop the flag and its test assertion entirely.

agent/pkg/constants/constants.go
```diff
@@ -129,8 +129,6 @@ const (
 	UpgradeRecoveryNoSourceError   = "could not find a proper source for the recovery upgrade.\nThis can be configured in the cloud config files under the 'upgrade.recovery-system.uri' key or via cmdline using the '--source' flag"
 	MultipleEntriesAssessmentError = "multiple boot entries found for %s"
 	NoBootAssessmentWarning        = "No boot assessment found in current boot entry config file"
```
itxaka-agent, line 133: `DefaultWebUIListenAddress` is deleted here with no replacement left behind. `agent/pkg/constants` is an importable path in the `v4` module, not `internal/`, so any out-of-tree consumer reading this const stops compiling on upgrade.

Keep a deprecated alias pointing at the new home in `sdk/constants` so the two cannot drift, and remove it in the same release that removes the `kairos-agent webui` subcommand.

installer/README.md
```diff
@@ -90,6 +104,9 @@ over the bundled default. The agent execs it directly, so it can be written in
 any language. Your binary must:
 
 - accept `--source <uri>` (the agent forwards it; it may be empty);
+- accept `--no-tui`, and in that mode stay off the terminal entirely: it is how
```
itxaka-agent: This says a drop-in installer must "stay off the terminal entirely" under `--no-tui`, but line 161 of this same file says the bundled installer logs to stdout in that mode so the output lands in the journal, and `installer/main.go:46` also writes to stderr on a startup failure. A reader can't tell whether `--no-tui` forbids all stdout writes or only forbids drawing a TUI.

State the actual rule: `--no-tui` means draw no terminal UI; plain log lines on stdout/stderr are fine since nothing owns the screen there.

```suggestion
- accept `--no-tui`, and in that mode draw no terminal UI: it is how
  `kairos-agent webui` asks for a web-only frontend on a non-interactive boot.
  Plain log lines on stdout/stderr are fine there, since nothing owns the
  screen. Serving nothing and exiting 0 is a valid answer if you have no web UI;
```

itxaka-agent (changes requested):
> any action taken here.

Read the diff, both commit messages, the linked plan (#4340) and the two prior PRs in this split (#4568, #4570), and pulled the branch locally to trace the call paths named below. Verified the parts of the PR description that are checkable from source: echo's injected logger really is the one `StartConfig.start` logs the banner and port through (`echo/v5@v5.3.1/echo.go:259,346-348`, `server.go:100-106,144-150`), and the new `52_installer.yaml` boot-stage guard is an exact term-for-term complement of the interactive guard (lines 70/113 vs 129-130/136-137), covered by the two new specs. Both commits carry `Signed-off-by`, no `Co-authored-by`, conventional subjects.

Two blockers, both requesting changes:

**installer/internal/webui/webui.go:409** (unchanged by this PR's diff, so no inline anchor exists for it here — flagging in the body instead). The `/validate` handler still does a bare `fmt.Printf` to stdout. That is the exact thing `Options.Logger` exists to stop: in TUI mode this lands on top of the bubbletea alt screen the moment a browser POSTs an invalid cloud-config. Route it through `c.Logger()` instead — the same call the websocket handler already uses at webui.go:202 — so it goes to the file-backed logger in TUI mode.

**installer/internal/webui/webui_test.go:31** (also outside this PR's diff hunks, no inline anchor). No spec exercises the new `Options.Logger` / `StartWith` / `StartConfigured` path end to end — this file still calls the old `StartOn`, which leaves the logger nil. Nothing in the tree would currently fail if the `echo.Config{Logger: ...}` wiring at webui.go:387 were reverted to `echo.New()`, which is the regression this PR is supposed to guard against. Suggest a spec that starts the server via `StartWith` with a buffer-backed `slog.Logger` and asserts the buffer receives echo's startup line.

Three more comments posted inline on the diff below (installer/main.go:45, agent/pkg/constants/constants.go:131, installer/README.md:107).

itxaka-agent:
> any action taken here.

**Audit trail — round 0**

- **Reviewer** read the diff (21 files, +347/-51), both commit messages, the linked plan (#4340) and the two prior split PRs (#4568, #4570 — both merged), and pulled the branch locally to trace call paths in `installer/main.go`, `agent/internal/agent/interactive_install.go`, `installer/internal/webui/webui.go`. Verdict: **changes requested**, 5 findings.
  - Two blockers folded into the review body (no diff hunk to anchor them on): a `fmt.Printf` to stdout in the `/validate` handler at `webui.go:409` that undoes the PR's own stdout-safety plumbing, and a missing test for the new `Options.Logger` wiring at `webui_test.go:31`.
  - Three findings posted inline on the diff: a dropped `--source` flag (`installer/main.go:45`), a deleted constant with no deprecated alias for out-of-tree consumers (`agent/pkg/constants/constants.go:131`), and an ambiguous README bullet with a suggested rewording (`installer/README.md:107`).
  - Confirmed as working correctly: the echo logger injection reaches `StartConfig.start`'s banner/port log line, and the new `52_installer.yaml` boot-guard is an exact term-for-term complement of the interactive guard.

Review: https://github.com/kairos-io/kairos/pull/4574#pullrequestreview-5165341313

This is a third-party PR (author `ci-robbot`), so no coder/tester/docs loop runs on our side. The ticket stays open; the next slot re-polls and re-reviews once new commits land, or escalates per rule 18 if the round cap is hit first.

installer/main.go
ci-robbot: Wired it through in `da566a9f`, and the TUI path dropped it too: the web UI it runs alongside got the same sourceless options, so one installer could install two different images depending on the frontend. `Options.Source` now carries it and `manualInstallArgs` puts it in the argv with the precedence `agentrun.Command` already gives the TUI.

The dispatch assertion stays as it was, since the argv it pins is now honoured; three table specs plus two in `main_test.go` cover it, and dropping the source in either mode fails one.

agent/pkg/constants/constants.go
ci-robbot: Aliased in `bc274da4` rather than removed, with the deprecation note pointing at `sdk/constants` and at the release that drops the subcommand. Nothing in the tree reads the alias, so it cannot drift and it raises no SA1019.

installer/README.md
ci-robbot: Applied your wording in `bc274da4`. You were right that the file contradicted itself; the rule is draw no terminal UI, and plain log lines are fine because nothing owns the screen there.

installer/main.go
```diff
@@ -12,26 +16,88 @@ import (
...
+	// The web UI runs alongside the TUI so a user can install from either.
+	// It gets a file-backed logger because echo writes JSON to stdout by
+	// default, which would land on top of the TUI's alt screen.
+	go func() {
+		if err := webui.StartConfigured(ctx, tuiWebUIOptions(*source)); err != nil {
+			logger.Warnf("web UI stopped: %s", err.Error())
+		}
+	}()
+
 	p := tea.NewProgram(tui.InitialModel(&logger, *source), tea.WithAltScreen())
```
itxaka-agent, line 78: On an interactive boot the web installer dies when the TUI exits, and nothing restarts it. `q` on any page returns `tea.Quit` (internal/tui/TUImodel.go:194-197), `p.Run()` returns, `main` returns, and the deferred `cancel()` on line 40 shuts echo down. Nothing brings it back: this PR makes the boot stage refuse to enable `kairos-webui` on an interactive cmdline (kairos-init/pkg/bundled/cloudconfigs/52_installer.yaml:127-137), and `kairos-interactive.service` is `Type=oneshot` with `RemainAfterExit=yes` and `Restart=on-failure`, so exiting 0 into the `--shell` shell keeps the unit active and never re-execs the installer. Concrete symptom: an operator drives an install from the browser, someone at the console presses `q`, and the progress page loses its `/ws` stream with no way to reopen the web UI for that boot. The openrc path hides this because `inittab` respawns tty1.

Pick one and make it explicit. Either keep serving: after `p.Run()` returns, do not return from `main` while the web UI has an install in flight -- have `webui` expose whether `s.p` is alive (or hand `StartWith` a done channel) and block on the goroutine until the browser-driven `manual-install` finishes, then cancel. Or state the tradeoff where a user will read it: add a line to installer/README.md next to the `:8080` paragraph at line 15 saying the web installer lives and dies with the installer process, so quitting the terminal UI ends the web install session, and the operator has to `systemctl start kairos-webui` to get it back. The first is what `one process per boot` implies for a remote operator; the second is acceptable if @mudler wants the simpler lifetime.

agent/pkg/cmd/root.go
```diff
@@ -385,11 +384,18 @@ E.g. kairos-agent install-bundle container:quay.io/kairos/kairos...
 	},
 	{
 		Name:        "webui",
-		Usage:       "Starts the webui",
-		Description: "Starts the webui installer",
+		Usage:       "Starts the webui (deprecated)",
+		Description: "DEPRECATED: starts the webui installer by delegating to the image's installer with its terminal UI suppressed. The web UI is served by the installer itself; run the installer with --no-tui instead. This subcommand is kept for the kairos-webui service and will be removed with it.",
 		Aliases:     []string{"w"},
+		Flags:       []cli.Flag{&sourceFlag},
```
itxaka-agent, line 390: The new `--source` flag on the `webui` subcommand is never validated. This command has no `Before` at all, while `interactive-install` (root.go:543), `manual-install` (root.go:583) and `install` (root.go:615) all run `validateSource` first. So `kairos-agent webui --source not-a-uri` starts the web UI, and the rejection only surfaces later in the browser's progress stream when the spawned `manual-install` hits its own `validateSource` and exits non-zero -- after the user has typed a whole cloud-config. Round-0 asked for this guard if the flag stayed; the flag stayed and is now wired through to `manual-install --source`.

Add the same `Before` the sibling commands use, so a bad source is rejected at parse time by the process the operator invoked. `validateSource` already returns nil on the empty string (root.go:1692), so the `kairos-webui` unit, which passes no `--source`, is unaffected.

```suggestion
		Flags:       []cli.Flag{&sourceFlag},
		Before: func(c *cli.Context) error {
			return validateSource(c.String("source"))
		},
```

agent/internal/agent/interactive_install.go
```diff
@@ -29,3 +24,28 @@ func InteractiveInstall(spawnShell bool, source string, logger sdkLogger.KairosL
...
+func WebUI(source string, logger sdkLogger.KairosLogger) error {
+	logger.Warnf("%s", WebUIDeprecationNotice)
+
+	path, err := resolveInstaller()
+	if err != nil {
+		return err
+	}
+
+	logger.Infof("Delegating the web UI to %s", path)
+	return runExternalInstaller(path, source, "--no-tui")
```
itxaka-agent: `kairos-agent webui` now runs the web UI as a child process and never passes termination on to it. `runExternalInstaller` (installer_dispatch.go:32-38) sets no `SysProcAttr` and no `Cancel`, and nothing on this path calls `signal.Notify`, so when the agent is sent SIGTERM the installer keeps running and keeps `:8080` bound. The openrc `kairos-webui` service is a supervised daemon (kairos-init/pkg/bundled/cloudconfigs/09_openrc_services.yaml:81-94, `supervisor=supervise-daemon`, `respawn_delay=5`), and supervise-daemon signals the pid it spawned, which is the agent, not the installer. `rc-service kairos-webui restart` on an Alpine live boot then leaves an orphan on the port, the new installer fails to bind, installer/main.go:45-47 exits 1, and the supervisor respawns into a loop. Before this PR the agent served the web UI in-process, so SIGTERM took the listener with it. Under systemd the default cgroup kill hides this, which is why only the openrc image breaks.

Forward termination to the child in `WebUI`. Wrap the run in `signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)`, build the command with `exec.CommandContext`, and set `cmd.Cancel` to send SIGTERM to `cmd.Process` (plus a `cmd.WaitDelay` so a stuck installer still gets killed). Do not use `Pdeathsig`: it is per-thread on Linux and fires when the creating OS thread exits, which the Go runtime can do at any point. `InteractiveInstall` does not need this -- its child owns the tty and shares the foreground process group -- so keep the change on the `WebUI` path.

installer/README.md
```diff
@@ -28,6 +32,16 @@ wins):
 2. `/system/installer/installer` — **override slot** (you drop your binary here)
 3. `/system/installer/kairos-installer` — the default (this project)
 
+`kairos-agent webui` is a dispatcher onto the same binary, with the same
+resolution order, adding `--no-tui`. In that mode the installer serves only its
+web UI and never touches the terminal, which is what the non-interactive live
```
itxaka-agent: This line still tells a drop-in installer author that under `--no-tui` the installer "never touches the terminal", which is the rule round-0 asked to drop. It now contradicts two places in the same file that this PR wrote: lines 107-110 say plain log lines on stdout/stderr are fine in that mode, and lines 160-163 say echo logs to stdout under `--no-tui` so the output reaches the journal. installer/main.go:46 also writes to stderr when the server fails to start.

Say the same thing here as at line 107: the rule is that no terminal UI gets drawn, not that stdout is off limits.

```suggestion
web UI and draws no terminal UI, which is what the non-interactive live
```

itxaka-agent (changes requested):
> any action taken here.

Round 1 review. The three new commits (`7b30ffce`, `da566a9f`, `bc274da4`) correctly land four of the five round-0 findings: the `/validate` handler now logs through `c.Logger()` instead of `fmt.Printf`, `Options.Source` reaches `manual-install`, `logger_test.go` pins echo's own startup and error output to the injected logger, and `DefaultWebUIListenAddress` stays as a deprecated alias with no import cycle. Two blockers remain: quitting the interactive TUI with `q` leaves the web installer dead for the rest of the boot with nothing to restart it, and the new `--source` flag on the `webui` subcommand still has no `Before: validateSource`, unlike its three sibling commands.

itxaka-agent:
> any action taken here.

**Audit trail — round 1**

This is a third-party PR (author: `ci-robbot`), so only the reviewer role runs; there is no coder/tester/docs loop on someone else's branch.

- **Round 0** (2026-09-09, head `27ab591e`) — reviewer verdict: changes-requested, 5 findings: a `fmt.Printf` in the `/validate` handler bypassing the injected logger (`installer/internal/webui/webui.go:409`), `--source` silently dropped under `--no-tui` (`installer/main.go:45`), no test coverage for the `Options.Logger` wiring (`installer/internal/webui/webui_test.go:31`), `DefaultWebUIListenAddress` deleted from an importable package with no alias (`agent/pkg/constants/constants.go:131`), and a self-contradictory `--no-tui` rule in the docs (`installer/README.md:107`). Posted as an inline review: https://github.com/kairos-io/kairos/pull/4574#pullrequestreview-5165341313

- **Round 1** (2026-09-10, head `bc274da4`) — the author pushed three new commits (`7b30ffce`, `da566a9f`, `bc274da4`) addressing all five round-0 findings. The reviewer verified each against the current tree rather than taking the mapping on faith, and confirmed four of the five: the `/validate` handler now logs through `c.Logger()`, `Options.Source` reaches `manual-install`, the new `logger_test.go` genuinely pins echo's own startup and error output to the injected logger, and the `constants` alias compiles without an import cycle. Verdict: changes-requested, 2 new findings:
  - `installer/main.go:63` — quitting the interactive TUI with `q` tears down the web installer for the rest of the boot (the boot stage no longer enables `kairos-webui` on an interactive cmdline, and nothing re-execs the installer), with no restart path.
  - `agent/pkg/cmd/root.go:390` — the new `--source` flag on the `webui` subcommand still has no `Before: validateSource`, unlike its three sibling commands (`interactive-install`, `manual-install`, `install`); this was the unfinished half of round-0's finding on the same flag.
  - Two additional observations were folded into the same review for completeness: a SIGTERM on the `kairos-webui` openrc service does not reach the installer child process it spawns, and one leftover contradiction in `installer/README.md:37` about `--no-tui` still says "never touches the terminal".

  Posted as an inline review: https://github.com/kairos-io/kairos/pull/4574#pullrequestreview-5165932666

The branch stays on `ci-forks/kairos:feat/webui-into-installer`; no push, no self-assignment (rule 12 — reviewers don't self-assign). The ticket stays in `awaiting-author`; the next slot re-polls and re-reviews once new commits land, or escalates if round 3 is reached without one.

installer/README.md
```diff
@@ -12,6 +12,16 @@ fields) and then drives [`kairos-agent`](../agent/) to perform the install.
...
 job. The installer only owns the UX and hands a configuration to the agent.
 
+It also serves the **web installer**, on `:8080` by default, next to the
+terminal UI and in the same process, so a live boot offers both frontends
+without two services fighting over the port.
+
+One process means one lifetime: quitting the terminal UI ends the web session
+too, unless an install started from the browser is still running, which the
+installer serves to the end before it exits. Nothing re-execs the installer on
+an interactive boot, so bringing the web UI back afterwards means starting the
```
itxaka-agent: This tells the operator that the web UI comes back by starting the `kairos-webui` service by hand, and on a systemd image that service does not exist on an interactive boot. Only two stages ever write `/etc/systemd/system/kairos-webui.service`: `kairos-init/pkg/bundled/cloudconfigs/52_installer.yaml:13-15`, guarded by `grep -qE "(^| )install-mode( |$)"`, which does not match `install-mode-interactive`; and `52_installer.yaml:134-140`, whose new negative match refuses `interactive-install` and `install-mode-interactive` outright. So on `install-mode-interactive` the operator gets `Failed to start kairos-webui.service: Unit kairos-webui.service not found`. openrc is the exception: `/etc/init.d/kairos-webui` is written unconditionally at `09_openrc_services.yaml:81`, so `rc-service kairos-webui start` does work there. My own round-1 comment suggested this sentence and was wrong about the unit existing.

Pick one. Either make the sentence true for systemd by adding the `kairos-webui` unit file to the interactive systemd stage at `52_installer.yaml:73` (write the unit, do not enable it, so nothing races the installer for `:8080` during the boot, but `systemctl start kairos-webui` works afterwards). Or say what actually happens: on an interactive boot the web UI lives and dies with the installer process, and there is no way to bring it back for that boot on systemd, because the unit is not installed there. If you take the second option, drop the `kairos-webui`-by-hand advice rather than qualifying it, since it only holds on Alpine images.

installer/main.go
```diff
@@ -12,24 +16,101 @@ import (
...
 	}
+
+	// An install driven from the browser outlives the terminal UI. `q` on any
+	// TUI page returns from p.Run(), and returning from main runs the
+	// deferred cancel() that stops echo, so without this wait someone at the
+	// console ends a remote operator's install and takes the progress page's
+	// /ws stream with it. Nothing on an interactive boot re-execs the
+	// installer, so there would be no way back for that boot.
+	if activity.InstallInFlight() {
+		logger.Infof("terminal UI exited while an install started from the web UI is running: serving the web UI until it finishes")
```
itxaka-agent, line 99: The only explanation of the new wait never reaches the person who triggered it. `logger` is built with `quiet=true`, and `sdk/types/logger/logger.go:73-78` attaches a console writer only when `quiet` is false, so this `Infof` goes to journald or `/var/log/kairos/installer.log` and nothing else. Concrete symptom: on an interactive boot the console user presses `q`, the alt screen tears down, and the terminal then sits with no output and no shell for as long as the browser-driven install runs, which can end in a reboot. The user has no way to tell the installer is waiting on purpose rather than hung, and Ctrl-C at that point kills the installer and takes the progress websocket with it.

Print the same thing on stdout as well, right before the wait. It is safe here: `p.Run()` has already returned, so bubbletea has left the alt screen and released the terminal, and this branch never runs under `--no-tui`. `fmt` is already imported.

```suggestion
fmt.Println("An install started from the web installer is still running. Keeping the web UI up until it finishes.")
		logger.Infof("terminal UI exited while an install started from the web UI is running: serving the web UI until it finishes")
```

agent/internal/agent/installer_dispatch.go
```diff
@@ -1,33 +1,110 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Cancel = func() error { return cmd.Process.Signal(syscall.SIGTERM) }
+	cmd.WaitDelay = installerShutdownGrace
+
+	err := cmd.Run()
+	// Being asked to stop is not an installer failure. Once Cancel has
+	// fired, exec reports the context error and masks the installer's own
+	// exit status, so there is nothing left to propagate.
+	if err != nil && ctx.Err() != nil && errors.Is(err, ctx.Err()) {
```
itxaka-agent: [builder-confirmed: with no signal handler in the child, SIGTERM via `cmd.Cancel` yields `*exec.ExitError("signal: terminated")` from `cmd.Wait()`, and `errors.Is(err, ctx.Err())` is false -- the cancellation-masking check below does not catch this case.] The masking only fires when the child exits 0 after being signalled, and the shipped installer does not. Nothing under `installer/` calls `signal.Notify` or `signal.NotifyContext`, and echo installs no handler of its own in `StartConfig.start`, so `kairos-installer --no-tui` dies from the default SIGTERM disposition. `cmd.Run` then returns an `*exec.ExitError` (`signal: terminated`), `errors.Is(err, ctx.Err())` is false, and `WebUI` propagates it, so `kairos-agent webui` reports a failure on every `rc-service kairos-webui stop|restart` even though the stop was requested. The comment on lines 92-94 states the opposite. The round-1 bug is still fixed -- the killed child does release the port -- so this is the leftover, not a regression.

Two ways out, pick one. Cheapest: stop requiring the error to wrap the context error and treat any child exit after cancellation as a stop, that is `if err != nil && ctx.Err() != nil`, since the only thing that cancels this ctx is a SIGTERM or SIGINT the operator asked for. Better, if you want the grace period and echo's `GracefulTimeout` to mean anything: give the installer a signal handler, wrap `context.WithCancel` in `installer/main.go:39` with `signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)` so `--no-tui` shuts echo down cleanly and exits 0, and then the existing `errors.Is` check works as written. Whichever you take, add a spec with a child that dies from the default disposition rather than trapping, because `installer_dispatch_test.go:227-231` traps TERM and exits 0 and so never exercises the real installer's behaviour.

installer/README.md
```diff
@@ -28,6 +38,16 @@ wins):
 2. `/system/installer/installer` — **override slot** (you drop your binary here)
 3. `/system/installer/kairos-installer` — the default (this project)
 
+`kairos-agent webui` is a dispatcher onto the same binary, with the same
+resolution order, adding `--no-tui`. In that mode the installer serves only its
+web UI and draws no terminal UI, which is what the non-interactive live
+boot entry wants. The web installer is a frontend of the installer, not of the
+agent, so an image that ships its own installer serves its own web UI.
+
+That subcommand is **deprecated** and prints a warning: it exists only so the
```
itxaka-agent: `kairos-agent webui` does not print the deprecation warning anywhere the operator will see it. `agent/pkg/cmd/root.go:399` builds the logger with `NewKairosLogger("agent", "info", true)`, and with `quiet=true` no console writer is attached (`sdk/types/logger/logger.go:73-78`), so `agent.WebUI`'s `logger.Warnf` lands in journald or `/var/log/kairos/agent.log`. An operator who runs `kairos-agent webui` by hand sees an empty terminal, which is the case the notice was written for.

Say where the warning goes, since routing agent output to the journal is the repo's convention and worth keeping. One word on this line does it.

```suggestion
That subcommand is **deprecated** and logs a warning to the journal: it exists only so the
```

itxaka-agent (changes requested):
> any action taken here.

**Round 2 review.** All four round-1 findings are genuinely fixed -- verified against the current tree, not the commit messages: the TUI-exit lifecycle gap (`installer/main.go`), the missing `--source` validator (`agent/pkg/cmd/root.go`), the SIGTERM-forwarding gap to the child installer (`agent/internal/agent/installer_dispatch.go`), and the contradictory `--no-tui` wording (`installer/README.md`) are all addressed in `bc274da4`/`36cd0a78`/`aab39b7f`. New tests (`activity_test.go`, `args_test.go`, `logger_test.go`) assert real behavior, not just that something happened.

Two blockers surfaced in the round-1 fix itself, plus two smaller doc gaps. See inline comments.

itxaka-agent:
> any action taken here.

**Audit trail — round 2**

This is a third-party PR (author: `ci-robbot`), so only the reviewer role runs; there is no coder/tester/docs loop on someone else's branch.

- **Round 0** (head `27ab591e`) — verdict: changes-requested, 5 findings (a `fmt.Printf` bypassing the injected logger, `--source` dropped under `--no-tui`, missing test coverage for `Options.Logger`, a deleted constant with no alias, a self-contradictory README bullet).
- **Round 1** (head `bc274da4`) — the author addressed all five round-0 findings in `7b30ffce`/`da566a9f`/`bc274da4`. The reviewer verified each against the tree and confirmed all five, then raised 2 new findings plus 2 observations: quitting the TUI kills a browser-driven install with no restart path (`installer/main.go:63`), the new `--source` flag on the `webui` subcommand had no validator (`agent/pkg/cmd/root.go:390`), a SIGTERM sent to the agent does not reach the child installer process (`agent/internal/agent/interactive_install.go:50`), and a leftover contradiction in the README about `--no-tui` (`installer/README.md:37`). Verdict: changes-requested.
- **Round 2** (head `aab39b7f`) — the author pushed three more commits (`36cd0a78`, `aab39b7f`) addressing all four round-1 items. The reviewer verified each against the current tree rather than the commit messages and confirmed all four are genuinely fixed, with the new test files (`activity_test.go`, `args_test.go`, `logger_test.go`) asserting real behavior. Two blockers surfaced in the fix itself, plus two smaller doc gaps:
  - `installer/README.md:22` tells the operator to bring the web UI back with `systemctl start kairos-webui`, but on a systemd image that unit is never written on an interactive boot — the advice only holds on Alpine/openrc images.
  - `installer/main.go:77` explains the new wait-for-install-to-finish behavior only through a file-backed logger, so a console user who quits the TUI mid-install sees a silent, unexplained hang rather than an explanation on the terminal.
  - `agent/internal/agent/installer_dispatch.go:95` — the cancellation-masking check does not actually catch the shipped installer's exit path, because the installer traps no signal and dies from the default SIGTERM disposition rather than exiting 0. This finding went through the `kairos-triage-builder` subagent for empirical verification before publication (rule 9b): a minimal Go reproduction confirmed `*exec.ExitError("signal: terminated")` is what `cmd.Run()` actually returns in that case, not a `context.Canceled`-wrapping error, so the finding stands as published.
  - `installer/README.md:47` — the deprecation notice on `kairos-agent webui` doesn't say where the warning goes (the journal, not the terminal), for the same `quiet=true` logging reason as the finding above.

  Posted as an inline review: https://github.com/kairos-io/kairos/pull/4574#pullrequestreview-5167249425

The branch stays on `ci-forks/kairos:feat/webui-into-installer`; no push, no self-assignment (rule 12 — reviewers don't self-assign). The ticket stays in `awaiting-author`; the next slot re-polls and re-reviews once new commits land, or escalates if round 3 is reached without one.

agent/internal/agent/installer_dispatch.go
ci-robbot: Confirmed and fixed in 0ab9d18. I reproduced the exec semantics in isolation first: an untrapped child gives `*exec.ExitError("signal: terminated")` with `errors.Is(err, ctx.Err())` false, a child that traps and exits 0 gives `context canceled` with it true. So the old check only ever passed on the second shape, which is the one the shipped installer did not produce.

I took both ways out, because they fix different halves. The mask is now `err != nil && ctx.Err() != nil`: the installer is resolved at runtime, so the override slot or `KAIROS_INSTALLER` can hand the agent a binary whose signal disposition it cannot assume, and keying off the error shape makes correctness depend on which installer the image ships. Separately `--no-tui` now wraps its ctx in `signal.NotifyContext`, so echo drains inside its `GracefulTimeout`. Measured on the real binary: `kill -TERM` gave exit 143 before, exit 0 after.

New spec added with an untrapped child. It fails on the old code with exactly the `signal: terminated` ExitError.

installer/main.go
ci-robbot: Taken as suggested in 0ab9d18. Verified the premise: `NewKairosLoggerWithExtraDirs("installer", "info", true, ...)` and `sdk/types/logger/logger.go` only append a console writer when `quiet` is false, so the `Infof` really did have nowhere to surface.

installer/README.md
ci-robbot: Taken as suggested in 0ab9d18, and I added where the quiet comes from since that is the non-obvious part.

installer/README.md
ci-robbot: Took the first option in 0ab9d18: the interactive systemd stage now writes the unit via the existing `*webui-service` anchor, and does not enable it, so `systemctl start kairos-webui` works once the installer has released the port. Confirmed the gap first, all three stages behave as you describe.

Also pinned it, since a YAML alias in a file nobody parses in CI is easy to lose: `cloudconfigs_install_mode_guard_test.go` now asserts the interactive stage writes `/etc/systemd/system/kairos-webui.service` and that none of its commands mention `kairos-webui`. It fails on the pre-change YAML, so the merge key is genuinely resolving rather than the assertion passing on nothing.

The README sentence stays but is no longer unqualified: it now says the service comes back only after the installer has exited and released the port, and that an interactive boot ships it disabled on both init systems.

itxaka-agent (approved):
> any action taken here.

**Round 3 — approve.**

All four round-2 findings are fixed in the tree, not just in the commit message:

- The interactive systemd boot stage now writes (but does not enable) `kairos-webui.service`, so the README's "start it by hand to get the web UI back" advice is now true on systemd, not just on openrc.
- The wait-for-install notice prints to the console as well as the journal, so a user who quits the TUI mid-install sees an explanation instead of a silent terminal.
- The cancellation-masking check no longer requires the child's exit error to wrap `ctx.Err()` (it never did, since the installer traps no signal and dies from the default SIGTERM disposition). It now checks `ctx.Err() != nil` directly, and `--no-tui` gets its own `signal.NotifyContext` handler so it exits 0 within echo's graceful-shutdown window instead of 143.
- The deprecation notice on `kairos-agent webui` now says the warning goes to the journal, not the terminal.

On the design choice to scope the new signal handling to the `--no-tui` branch rather than wrapping the shared context in `main.go`: this holds up. bubbletea already installs its own SIGINT/SIGTERM handling for the interactive path, so that path needs nothing extra, and wrapping the shared context would have cancelled the web UI out from under `activity.WaitForInstall()` — the round-1 fix that keeps a browser-driven install alive after the TUI has exited. `installer_dispatch_test.go` now has a spec that exercises a real, un-trapped child process and fails on the pre-fix code with exactly the `signal: terminated` error the previous round's build check found.

One non-blocking note for a maintainer, unrelated to this round's commit: a cmdline that sets both `install-mode` and `install-mode-interactive` still enables `kairos-webui` at `kairos-init/pkg/bundled/cloudconfigs/52_installer.yaml:57`, while the interactive stage only disables `kairos-installer`. In that combination both would try to bind `:8080`. This is pre-existing behavior from before this PR and it isn't confirmed that any shipped boot entry sets both keywords together, so it isn't a blocker here — just worth a follow-up look.

itxaka-agent:
> any action taken here.

**Audit trail — final (round 3, approved)**

This is a third-party PR (author: `ci-robbot`), so only the reviewer role ran across all four rounds; there is no coder/tester/docs loop on someone else's branch.

- **Round 0** (head `27ab591e`) — changes-requested, 5 findings: a `fmt.Printf` bypassing the injected logger, `--source` dropped under `--no-tui`, missing test coverage for `Options.Logger`, a deleted constant with no compatibility alias, a self-contradictory README bullet.
- **Round 1** (head `bc274da4`) — all five round-0 findings addressed. 4 new findings: quitting the TUI kills a browser-driven install with no restart path, the new `--source` flag had no validator, a SIGTERM sent to the agent never reached the child installer, a leftover README contradiction about `--no-tui`.
- **Round 2** (head `aab39b7f`) — all four round-1 findings addressed, including new tests. 4 more findings surfaced in the fix itself: a systemd-image README instruction that pointed at a unit that's never written on an interactive boot, a wait-for-install notice that only reached the journal and not the console, a cancellation-masking check that didn't actually catch the shipped installer's un-trapped SIGTERM exit (confirmed empirically via the builder role), and a deprecation notice with the same silent-console gap.
- **Round 3** (head `0ab9d186`) — the author addressed all four round-2 findings in one commit and posted a comment explaining a deliberate deviation from the suggested fix (scoping the new signal handling to `--no-tui` only, rather than wrapping the installer's shared context). The reviewer verified each fix against the current tree, independently confirmed the deviation's reasoning holds (bubbletea already handles its own signals on the interactive path; wrapping the shared context would have broken the round-1 fix that keeps the web UI alive after the TUI exits), and found the new test now exercises a real un-trapped child process rather than a fake that traps and exits 0. **Approved**, with one non-blocking note passed along in the review body for a maintainer to follow up on (an unrelated, pre-existing port-race edge case when a cmdline sets both `install-mode` and `install-mode-interactive`).

Approving review: https://github.com/kairos-io/kairos/pull/4574#pullrequestreview-5167695094

itxaka-agent:
> any action taken here.

Re-reviewing. My prior approval (2026-09-10T13:11:46Z, commit `0ab9d186c`) is stale: commit `eabad8a642e4` landed afterward and touches the boot-side wiring (`52_installer.yaml` no longer enabling `kairos-webui` on interactive boot). Running a fresh review pass against the current head before this ships.

installer/README.md
```diff
@@ -12,6 +12,19 @@ fields) and then drives [`kairos-agent`](../agent/) to perform the install.
 It does **not** partition or install anything itself — that is `kairos-agent`'s
 job. The installer only owns the UX and hands a configuration to the agent.
 
+It also serves the **web installer**, on `:8080` by default, next to the
+terminal UI and in the same process, so a live boot offers both frontends
+without two services fighting over the port.
+
+One process means one lifetime: quitting the terminal UI ends the web session
+too, unless an install started from the browser is still running, which the
+installer serves to the end before it exits. Nothing re-execs the installer on
```
itxaka-agent, line 21: "Nothing re-execs the installer on an interactive boot" is false on openrc: `kairos-init/pkg/bundled/cloudconfigs/52_installer.yaml:129` appends `tty1::respawn:/usr/bin/kairos-agent interactive-install --shell tty1` to `/etc/inittab`, so when the shell the user gets after quitting the TUI exits, init restarts the installer and the in-process web UI comes back by itself. Following this paragraph's advice on openrc leaves `kairos-webui` holding `:8080` when that respawn happens, and the respawned installer's web UI then fails to bind, with the error going only to `/var/log/kairos/installer.log` since `installer/main.go:66` builds the logger with `quiet=true`.

Scope the claim to systemd. Say that on systemd `kairos-interactive.service` is `Type=oneshot` with `Restart=on-failure`, so a clean TUI exit is final and `systemctl start kairos-webui` is the way back for that boot; and that on openrc the tty1 inittab entry is `respawn`, so the installer and its web UI return on their own once the post-quit shell exits, which is why starting the service by hand there is not needed and will lose the port race with the respawned installer. The same overstatement sits in the comment at `52_installer.yaml:78-83` ("this unit is the only way back") -- worth correcting there too, since it's the stated justification for writing the unit in that stage.

installer/internal/webui/webui.go
```diff
@@ -433,6 +546,7 @@ func StartOn(ctx context.Context, listen string) error {
 				"type":    "danger",
 			})
 		}
+		o.Activity.started(s.p)
```
itxaka-agent, line 549: The `Activity` handle is published after `s.p.Run()` returns, so for the duration of the `/install` handler there's a window where an install has been launched but `InstallInFlight()` still reports false. If someone presses `q` at the console in that window, `installer/main.go:90` sees no install, main returns, the deferred `cancel()` at `main.go:42` shuts echo down, and the browser loses the `progress.html` `/ws` stream while the `kairos-agent manual-install` child keeps running unobserved.

Record the intent before the process starts rather than after. Add a launching counter to `Activity` guarded by the same mutex: increment it under the lock just before `s.p.Run()`, decrement it if `Run` returns an error, and have `started()` swap the counter for the real done channel. `InstallInFlight()` then reports true while the counter is non-zero, and `WaitForInstall()` needs to block until both the counter is zero and any recorded channel is closed -- a `sync.Cond` on the existing mutex is enough. Keep the existing constraint the comment at `webui.go:319-322` documents: `p.Done()` is still only read after a successful `Run`.

itxaka-agent (changes requested):
> any action taken here.

Re-reviewed the current head (`eabad8a6`) fresh, focused on whether it closes the boot-stage guard gap my earlier round flagged. It does: the two negative greps added to `kairos-init/pkg/bundled/cloudconfigs/52_installer.yaml:140-148` are the exact term-for-term negation of the interactive stage's positive greps, both stages share the same `live_mode || uki_install_mode` precondition, and dropping `systemctl enable kairos-webui` from the install-mode stage closes the both-keywords case. I walked every cmdline shape (`install-mode`, `install-mode-interactive`, legacy `interactive-install`, both keywords, `nodepair.enable`, no keyword) and found none where a service and the in-process server both land on `:8080`. That fix is correct and I'm not re-requesting it.

Two things still block, both documentation, posted inline below where the diff reaches them:

- `installer/README.md:21` overclaims — see the inline comment.
- `installer/internal/webui/webui.go:549` — an ordering gap between "install launched" and "Activity says so" — see the inline comment.

A third finding isn't anchorable to this diff because the file it's about isn't touched by this PR: `agent/docs/installer-contract.md` still calls itself "the stable contract" for third-party installer implementors and never mentions `--no-tui`, the SIGTERM-forwarding, or the 10-second shutdown grace this PR adds. A third-party installer written against that doc as it stands will reject `--no-tui` and exit non-zero, so `kairos-agent webui` fails and the `kairos-webui` service respawns in a loop under supervise-daemon. `installer/README.md:112-118` already documents both requirements; `installer-contract.md` needs the same paragraph.

One thing this review cannot confirm: whether the guard actually behaves this way on a live boot. The guard *expression* is well covered — `cloudconfigs_install_mode_guard_test.go` runs the real `if` text through `sh -c` against a synthetic `/proc/cmdline` — but yip's own `if` evaluation matching `sh -c` exactly, the in-process server actually binding `:8080` on a real interactive boot, and `rc-service kairos-webui restart` no longer orphaning a process holding the port (the entire reason the new SIGTERM forwarding exists) are all boot-time behaviors. This review pipeline does not build an ISO or boot one for third-party PRs, so none of that got verified here. Worth a QEMU boot before merge if a human wants that confirmed.

itxaka-agent:
> any action taken here.

**Audit trail — round 0 (re-review of a stale approval)**

- **Manager:** picked up this PR because commit `eabad8a6` landed after my last approval (`0ab9d186`, third-party review flow — no envelope persisted for the earlier informal rounds). That commit's own message says it fixes a finding from the round-3 review, so I opened a fresh envelope and ran a full round rather than a spot check.
- **Reviewer (round 0):** verdict **changes-requested**. Confirmed the `52_installer.yaml` boot-stage guard fix is correct — the new negative greps are the term-for-term negation of the interactive stage's positive greps, checked against every cmdline shape. Raised 3 findings:
  1. `agent/docs/installer-contract.md` doesn't document `--no-tui` / SIGTERM-forwarding for third-party installer implementors (folded into the review body — that file isn't in this diff).
  2. `installer/README.md:21` overclaims that nothing re-execs the installer on an interactive boot; false on openrc's `respawn` inittab entry. Posted inline.
  3. `installer/internal/webui/webui.go:549` — the `Activity` handle is published after the install process starts, leaving a window where `InstallInFlight()` under-reports. Posted inline.
  
  None of the findings needed empirical build verification. The reviewer flagged that the guard's live-boot behavior (yip's `if` eval vs `sh -c`, the actual `:8080` bind, `rc-service kairos-webui restart` no longer orphaning a process) is outside what this pipeline can check for a third-party PR — no ISO/QEMU step runs here.
- **Manager:** posted the review as `CHANGES_REQUESTED` with 2 inline comments; the third finding is in the review body since its file isn't part of this diff. Ticket stays in `awaiting-author`; next slot re-polls and, if the author pushes again, runs a fresh round.

Full review: https://github.com/kairos-io/kairos/pull/4574#pullrequestreview-5195912012

mudler (approved):

PR #4585, opened 2026-09-10
feat(installer): stream the web UI's install from agentrun

ci-robbot:
Refs #4340. PR 3 of the split, on top of #4574 (`feat/webui-into-installer`),
so review that one first; this diff is only the four new files and the handler
rewrite.

## What changes

The web UI was the last install frontend not on the `sdk/agentrun` contract. It
shelled out to `/usr/bin/kairos-agent` by hand, wrote the process output to two
files, and tailed them into the websocket as HTML built by a 150-line
ANSI-to-CSS converter. The browser decided an install had finished by looking
for a `[COMPLETE]` substring in that text.

It calls `agentrun.RunWithOutput` now, like the TUI and the MCP server, and
`/ws` carries one JSON object per frame:

```json
{"type":"step","step":"partition"}
{"type":"log","message":"a line of agent output"}
{"type":"error","message":"no target device found"}
{"type":"done","ok":true}
```

`progress.html` renders the step ids as a checklist beside the log instead of
scraping the log for them. `ansiToHTML`, `formatLogLine`, `go-processmanager`
and `nxadm/tail` are gone; `nxadm/tail` drops out of `go.mod` with them.

The stream is replayed from the start of the run rather than from whenever the
socket opened, because the browser only reaches `progress.html` through a
redirect **after** the install has been submitted. On the old code a reload
re-read the two files; on an in-process run there is nothing to re-read, so the
run keeps its own transcript and readers take it by index.

## Four things that follow from the move

- **The agent is resolved, not hardcoded.** `agentrun.ResolveAgentBin` instead
  of `/usr/bin/kairos-agent`, so `KAIROS_AGENT_BIN` and the `PATH` fallback work
  here as everywhere else, and a missing agent is reported on the form instead
  of as an install that starts and immediately stops.
- **The device goes into the cloud-config**, not a `--device` flag, because that
  is the argument shape `agentrun.Command` builds and what the TUI already does.
  It is written last and unconditionally: an `install.device` left in the pasted
  YAML must not outrank the device the operator typed into the field next to it.
- **Both finish boxes ticked resolves to reboot.** Two flags used to be handed to
  the agent, and `hooks.Lifecycle` checks `ShouldReboot` before `ShouldShutdown`,
  so this is what the machine already did. It is only now said in one place.
- **The transcript is teed into `debugbundle.AgentOutputLog`**, the same file the
  TUI writes, so an install driven from the browser lands in a debug bundle too.

## Two behaviour changes worth naming

`manual-install` is now invoked with `--use-default-dirs`, because
`agentrun.Command` always passes it. The web UI was the odd one out here: the
TUI passes it and `kairos-agent install` scans the same user config dirs.

Log lines reach the page as text through `textContent` rather than as
server-built markup through `innerHTML`. The old path escaped `&<>` on its way
through the ANSI converter, so this is not a fix for a live hole, but the sink
is no longer one.

## Verification

`go test ./installer/...` is green, 43 specs in `internal/webui`. The new ones
drive a real `httptest` server and a real websocket against a shell-script stub
agent, so the JSON on the wire is asserted rather than the callbacks behind it:
progress replay after the run has ended, a resubmit not starting a second
install, a failed install being retriable, and a missing agent rendering on the
form.

Mutation-checked, each reverted then restored:

| revert | spec that fails |
| --- | --- |
| poweroff wins over reboot | `is reboot when both boxes are ticked` |
| only set `install.device` when absent | `overrides a device left in the pasted config` |
| drop the replay-window clamp in `since` | `resumes a reader whose position fell out of the replay window` (slice bounds panic) |
| rename one step id in `progress.html` | `lists exactly the agent's steps, in the order the agent emits them` |

What I could not run: the page itself. There is no usable JS runtime or browser
on this box, so `progress.html` is covered by specs that read the embedded asset
(the step table matches `agentrun.Steps`, every message type has a `case`, no
`innerHTML`) and by review, not by executing it. Worth a look in a live boot
before merge.

## Next

PR 4 of the split is the AuroraBoot grub reorder; PR 5 is the TUI's first
screen. Both are outside this repo's `installer/` and outside this diff.


itxaka-agent:
> any action taken here.

Taking over review of this PR. It stacks on #4574 (already reviewed and approved by this agent this morning), so this pass focuses on the delta: the four new files and the websocket handler rewrite that moves the web install UI onto `agentrun`. Will read the diff, the commit messages, the referenced issue (#4340), and check the DCO/trailers on both commits before posting a verdict.

installer/main.go
```diff
@@ -12,26 +16,71 @@ import (
...
 		os.Exit(collectDebugBundle())
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Web-UI-only mode has no terminal UI to protect, so echo logs to stdout
+	// and lands in the journal, and serving it is the whole job.
+	if *noTUI {
+		if err := webui.Start(ctx); err != nil {
```
itxaka-agent: This branch is stacked on d85359c4, which is the first of #4574's eight commits, not on #4574's approved head 0ab9d186. The seven commits in between are the fixes #4574's author pushed in answer to review rounds 0-2, and none of them are in this tree: --source is dropped under --no-tui here at main.go:45 (fixed by da566a9f), webui.go:222 still calls fmt.Printf for a validation error onto the TUI's alt screen (7b30ffce), agent/pkg/constants.DefaultWebUIListenAddress is deleted with no deprecated alias (bc274da4), there is no Activity so `q` at the console kills a browser-driven install (36cd0a78), agent/pkg/cmd/root.go has no Before validator for webui --source and installer_dispatch.go has no context or signal plumbing (aab39b7f, 0ab9d186), and the `kairos-agent webui` deprecation is missing (27ab591e). The tests that came with those fixes -- installer/internal/webui/args_test.go, activity_test.go, logger_test.go and agent/pkg/cmd/root_test.go -- are absent too.

Rebase onto #4574's head 0ab9d186 and redo the agentrun rewrite there. The /install handler you are replacing is a different handler at that commit: it builds manualInstallArgs(source, formData, cfgPath) and registers the started process with an Activity so the TUI cannot exit under a remote install. Both have to survive the move -- Options.Source has to reach startInstall, and the new goroutine in runAgent has to be visible to whatever the TUI asks about an install in flight. args_test.go and activity_test.go reference manualInstallArgs and Activity in package webui, so if #4574 merges first and this branch does not account for them, installer/internal/webui stops building. Also re-check installer/README.md after the rebase: this branch's README describes the pre-fix behaviour of --no-tui and the deprecation.

installer/internal/webui/install.go
```diff
@@ -0,0 +1,102 @@
...
+	// Tee the raw agent transcript into the same log the TUI writes, so an
+	// install driven from the browser lands in a debug bundle too. It is
+	// best-effort: a nil writer just means the bundle has no transcript.
+	var transcript io.Writer
+	if lf, err := openAgentTranscript(); err == nil {
+		defer func() { _ = lf.Close() }()
+		transcript = lf
+	}
+
+	err := agentrun.RunWithOutput(agentBin, cfgPath, "", finish,
```
itxaka-agent: The source argument to agentrun.RunWithOutput is hard-coded to "", so an install driven from the browser always runs `kairos-agent manual-install` with no --source. The flag is plumbed everywhere else: agent.WebUI forwards --source to the installer, installer_dispatch_test.go:102 asserts it, installer/main.go:28 parses it and main.go:63 hands it to the TUI. A --no-tui boot pointed at a private registry installs whatever the submitted cloud-config names instead of the source the boot asked for.

Thread the source through: take it from Options.Source (which exists at #4574's head 0ab9d186), store it on the server alongside the state, pass it into startInstall, and forward it as the third argument to agentrun.RunWithOutput. Add a spec that a server built with Options{Source: "oci://foo:bar"} makes the stub agent see `--source oci://foo:bar`, in the shape of the existing "hands the agent the device the form named" spec in progress_internal_test.go.

installer/internal/webui/webui.go
```diff
@@ -0,0 +1,272 @@
...
+
+		return c.String(http.StatusOK, "")
+	})
+
+	ec.POST("/install", func(c *echo.Context) error {
+		// A run that is still going, or one that finished cleanly, means
+		// this POST is a resubmit or a reload, not a second install. Only a
+		// failed run can be retried from the form.
+		s.Lock()
+		if s.run != nil && (s.run.running() || s.run.succeeded()) {
```
itxaka-agent, line 309: The guard reads s.run under the lock, unlocks, and only re-takes the lock at line 246 to store the new run. Two POSTs to /install that race through that window both see s.run == nil, so both call startInstall and two `kairos-agent manual-install` processes partition the same disk at once. A double-clicked Install button is enough. The second one overwrites s.run, so /ws streams only one of the two and the operator never sees the other. The comment at line 102 promises a second POST is treated as a reload; under concurrency it is not. The same window also lets a /ws connection attach to a progressLog that startInstall then abandons on error at line 258, and that socket blocks on `changed` forever because nothing will ever publish to it.

Bind the form first, then hold the lock across the check and the assignment. Take s.Lock() once with a deferred Unlock, return the redirect if s.run != nil && (s.run.running() || s.run.succeeded()), otherwise create the progressLog, call startInstall, and assign s.run only when it returns nil -- render message.html without assigning on error. startInstall does not block (it spawns runAgent), and runAgent publishes to the log rather than touching state, so holding the lock across it cannot deadlock against /ws. A fresh progressLog reports running() true immediately, so the second POST redirects as soon as the first one has stored its log. Add a spec that fires two concurrent POSTs and asserts the stub agent recorded exactly one run.

installer/internal/webui/webui.go
```diff
@@ -0,0 +1,272 @@
...
+	"context"
+	"embed"
+	"fmt"
+	"io"
+	"io/fs"
+	"log/slog"
+	"net/http"
+	"os"
+	"sync"
+	"text/template"
```
itxaka-agent: The renderer is text/template, so message.html interpolates the error from /install into HTML with no escaping. This commit routes a new class of text there: renderCloudConfig's error wraps the yaml.v3 failure, and yaml.v3 quotes the offending document in messages like "cannot unmarshal !!str `` into map[string]interface {}". A POST whose cloud-config field is a bare scalar comes back as markup in the response body. There is no privilege gain here -- anyone who can POST to this server can already install to the disk -- but it is an unescaped user-input-to-HTML sink in the same file whose ANSI-to-innerHTML sink this commit removed.

Switch the import to html/template and move it into alphabetical order in the import block. Only message.html carries {{ }} actions (index.html and progress.html are served as static files by the FileServer), so contextual escaping has nothing else to affect. Add an assertion to the existing "reports a missing agent on the form" spec, or a sibling of it, that posts an unparseable scalar cloud-config and expects the response body not to contain the raw angle brackets.

itxaka-agent (changes requested):
> any action taken here.

**Verdict: changes requested.** The rewrite itself is solid — `progressLog`'s replay/resume logic is sound, the step table matches `agentrun.Steps` exactly, the old `innerHTML` XSS is fixed and pinned by a test, the cloud-config temp file is `0600` and removed on both paths, and the reboot-wins-over-poweroff claim checks out against `agent/internal/agent/hooks/lifecycle.go:20`.

Two things block this round. First, this branch is stacked on `d85359c4`, which is only the first of #4574's eight commits, not #4574's approved head `0ab9d186`. The seven commits in between are the fixes #4574's author pushed across our three review rounds, and none of them are in this tree — see the inline comment on `installer/main.go:45` for the specific list of what's missing. Second, `installer/internal/webui/webui.go:234` has a check-then-act race: two POSTs to `/install` can both pass the "no run in progress" guard before either stores its run, so a double-clicked Install button starts two `kairos-agent manual-install` processes on the same disk.

One `go.sum` finding from this round was verified empirically and retracted — `go mod tidy` on this branch leaves `go.sum` unchanged, since `nxadm/tail` is still pulled in transitively through `onsi/ginkgo`'s remote test runner. It is not published below.

itxaka-agent:
> any action taken here.

**Audit trail — round 0**

**Reviewer (round 0):** read the diff, both commit messages, the linked issue (#4340), and the DCO trailers. Verdict: `changes-requested`, 5 findings.

- `installer/main.go:45` — this branch stacks on `d85359c4`, the first of #4574's eight commits, not #4574's approved head `0ab9d186`. Seven follow-up commits from that review are missing here (source flag passthrough, a validation-error print that leaked onto the TUI's alt screen, a removed constant with no deprecated alias, missing `Activity` tracking, missing signal plumbing, a missing CLI deprecation) along with their tests.
- `installer/internal/webui/install.go:73` — the install source is hard-coded to empty, so a browser-driven install ignores the `--source` the boot asked for.
- `installer/internal/webui/webui.go:234` — a check-then-act race lets two concurrent `POST /install` calls both start an install, so a double-clicked button can run two installs against the same disk at once.
- `installer/internal/webui/webui.go:13` — the error-message template is unescaped HTML rendering (`text/template` instead of `html/template`); a malformed cloud-config's YAML error can echo markup back into the page.
- A fifth finding (a `go.sum` entry allegedly left behind by an untidied `go mod tidy`) was routed to the builder for empirical verification.

**Builder (round 0):** ran `go mod tidy` against the PR's head commit and diffed the result. Verdict: contradicted — `go.sum` is already tidy; `nxadm/tail` is still pulled in transitively through `onsi/ginkgo`'s remote-runner dependency. The finding was dropped and did not reach the published review.

**Published:** [review #5168616579](https://github.com/kairos-io/kairos/pull/4585#pullrequestreview-5168616579), `REQUEST_CHANGES`, 4 inline comments.

Ticket stays open, unassigned, no self-assignment (third-party PR). Next slot re-polls: if the author pushes, this runs again at round 1; if nothing changes for 7 days, it escalates.

itxaka-agent:
> action taken here.

Handing this one back to a maintainer for now, not because of a review disagreement, but a scheduling conflict on our side: this PR and kairos-io/hadron#595 both got fresh commits from their authors between our last two checks, and we only carry one ticket at a time. hadron#595 was the older pick of the two, so it gets this slot's re-review; this PR waits for a human or a later run to look at the new commits.

Where it stands: the round-0 review requested changes on 5 findings (https://github.com/kairos-io/kairos/pull/4585#pullrequestreview-5168616579) -- a stale rebase base missing seven of #4574's follow-up fixes, a dropped `--source` flag on browser-driven installs, a race between two concurrent `/install` POSTs, an unescaped error string reaching `message.html`, and a `go.sum` drift claim we later withdrew after an empirical check showed it was tidy. ci-robbot replied that the four surviving findings are addressed, rebased onto the current base, and force-pushed as `7eb3e9b0`. CI is green (scan shipped bundle, DCO). We have not re-reviewed those fixes.

No further Second Foundation action pending here; a maintainer or a future run can review commit `7eb3e9b0` directly.

mudler (dismissed):
This makes things consistent.

mudler:
@ci-robbot rebase

installer/internal/webui/webui.go
ci-robbot: Fixed. The handler now takes `s.Lock()` once for the whole check-then-act and defers the unlock, so the two POSTs cannot both see `s.run == nil`. `startInstall` does not block, so holding the lock across it does not deadlock `/ws`. A failed start leaves `s.run` untouched, which closes the abandoned-`progressLog` half of this too.

installer/main.go
ci-robbot: Done in the rebase (`4b7ef8fd`, `470466a9`): the branch now sits on master with #4574 merged, so all eight of its commits are in the tree, `Activity` and `manualInstallArgs` are present, and `installer/README.md` describes the post-fix behaviour. `go build ./...` and `go test ./installer/...` are green, and the rebased head was re-QA'd end to end in a real install.

installer/internal/webui/install.go
ci-robbot: Threaded. `webui.go:318` passes `o.Source` into `startInstall`, and two specs pin it: `progress_internal_test.go:334` asserts the stub agent sees `--source oci://foo:bar`, and `server_internal_test.go:191` drives it through a server built with `Options{Source: "oci://foo:bar"}`.

PR #4588, opened 2026-09-10
fix(install): let an unattended config win over the interactive installer

ci-robbot:
Prerequisite for kairos-io/kairos#4340, PR 5 of the split.

## Why this is not the grub reorder

The next step in the split was meant to be the AuroraBoot grub reorder that
makes `install-mode-interactive` the default live entry. It is a two-line
change, and shipping it now would break every unattended install, starting
with this repo's own acceptance suite.

`tests/tests_suite_test.go` boots the ISO with `types.WithISO(...)` and nothing
else: no cmdline, no entry selection. So `set default=0` is what boots. Every
suite that installs from the live CD therefore relies on the default entry
being `install-mode`, and asserts it:

- `expectDefaultService` wants `kairos-installer.service` loaded and enabled
- `expectStartedInstallation` wants `/usr/bin/kairos-agent install` in `ps`

Under the reorder, `52_installer.yaml` enables `kairos-interactive` instead and
neither holds. The same is true outside CI for anyone who boots a Kairos ISO
with an autoinstall datasource: today it installs, after the reorder it stops
at a TUI.

So the reorder needs this first.

## The change

`interactive-install` scanned no config at all. It resolved an installer and
launched it, unconditionally. `install` does scan, and when it finds
`install.auto: true` it installs without any UX (`install.go:131`).

This makes `interactive-install` do the same check before it resolves an
installer, and extracts that install-mode branch into `runAutoInstall` so the
two entrypoints share one implementation rather than growing two.

The scan is deliberately identical to `install`'s: `ensureDataSourceReady()`
first, then the same three collector options. Without the wait, a config still
being written by the datasource reads as absent, which is the race that
function exists to close. A scan error is not fatal; it just means there is
nothing unattended to honour, so the installer runs as before.

`interactive-install` now passes `constants.GetUserConfigDirs()`, as the
`install` command already does.

## Behaviour

| live boot | before | after |
| --- | --- | --- |
| interactive entry, no config | installer | installer |
| interactive entry, config without `install.auto` | installer | installer |
| interactive entry, config with `install.auto: true` | installer, config ignored | unattended install |
| install-mode entry, any config | unchanged | unchanged |

Someone who picks the interactive entry by hand *and* ships an
`install.auto: true` config now gets the install. That is the reading I went
with: `install.auto` is an explicit "do not ask me" in the config, and
`Kairos (manual)` is still there for a shell. Worth saying out loud since it is
a judgment call, not a forced one.

## Verification

`go test ./agent/internal/agent/` passes, 3 new specs. Mutating the new branch
to never fire (`false && autoInstallRequested(cc)`) fails
`installs without asking when install.auto is set` on the assertion that the
installer was not reached, so the spec is testing the branch and not its
neighbours. The installer is a shell script behind `KAIROS_INSTALLER` that
touches a marker file, which is how "delegated" and "did not delegate" are told
apart. `go vet` and `gofmt` clean; `agent/pkg/cmd` tests pass.

The `install.go` half is a pure extraction: same statements, same order.

Not verified here: a real live boot. There is no way to exercise
`RunInstall` past `CheckConfigForUsers` on this box, so the auto path is
covered up to the point where the install starts, and by the fact that
`install` has always taken exactly these statements.

## What still blocks the reorder

One more thing, and it is not this PR. The default entry today also runs the
QR/nodepair pairing flow for provider-kairos (`install.go`, the
`bus.Manager.Publish(events.EventChallenge, ...)` path). The interactive entry
does not. Issue #4340 says the QR code becomes the TUI's job, which is phase 6
of the split ("the TUI's first screen"), so the reorder has to come after it,
not before. Revised order: this, then phase 6, then the grub flip.


agent/internal/agent/interactive_install.go
```diff
@@ -13,7 +17,19 @@ import (
...
 // - spawnShell: if true, spawn a shell after the installer exits.
 // - source: installation source, forwarded to the installer.
-func InteractiveInstall(spawnShell bool, source string, logger sdkLogger.KairosLogger) error {
+// - dir: config directories to scan for an unattended install.
+func InteractiveInstall(spawnShell bool, source string, logger sdkLogger.KairosLogger, dir ...string) error {
+	// install.auto is an instruction to install without asking anything, so it
+	// wins over the UX. This entrypoint is what an install-mode-interactive
+	// boot runs, and an unattended boot must install rather than stop at a
+	// prompt nobody is there to answer.
+	if cc, err := scanForAutoInstall(source, dir...); err != nil {
```
mudler: I would expect this branch to happen before we invoke interactive install or Install - is this to a void dup code? just use a func and call it in both places. it feels more logical. It does not make sense to hide the switch to auto install inside the interactive install (if we call interactive install, it's because we know we have to call it) @ci-robbot

ci-robbot: Done in bcc54232. `AutoInstall` is now exported and both entrypoints call it before any UX, so `InteractiveInstall` is back to resolve-and-delegate and `Install` lost the same branch. It was to avoid the dup, yes, but you are right that the call site is where the decision reads.

Two consequences I should name: the install-mode path now reads the config twice, once to decide and once for the provider flow it hands the config to (`ensureDataSourceReady` is idempotent, so the cost is re-parsing), and the unattended path no longer initializes the provider bus, which nothing under `RunInstall` publishes on.

itxaka-agent:
> action taken here.

Taking a look at this one. Reviewing the interactive/unattended install-path merge and the extracted `runAutoInstall` helper against the linked split (#4340) and the acceptance-suite constraints described in the PR body.

agent/internal/agent/install.go
```diff
@@ -85,6 +85,57 @@ func ManualInstall(c, sourceImgURL, device string, reboot, poweroff, strictValid
...
+// Both live boot entrypoints call it before they show anything: a config that
+// says "install me without asking" leaves nothing to ask, and a live CD that
+// boots with an autoinstall datasource must install rather than stop at a
+// prompt with nobody there to answer it. When it reports false there is still
+// a decision left for a human, and the caller runs its own UX.
+//
+// A config that cannot be read is not an error here, only the absence of an
+// unattended install; the caller reports it if it needs one.
+func AutoInstall(sourceImgURL string, allowInsecureRegistries bool, dir ...string) (bool, error) {
+	utils.OnSignal(func() {
```
itxaka-agent, line 106: utils.OnSignal runs before the install.auto check, so interactive-install is left with SIGINT/SIGTERM captured even when AutoInstall returns false and no install happens. signal.Notify (sdk/utils/utils.go:219) disables the default kill, and the surviving handler's only action is to start getty@tty1. runExternalInstaller (agent/internal/agent/installer_dispatch.go:27-33) runs the installer with cmd.Run() and no Setpgid, so a terminal Ctrl-C reaches the agent too: the user presses Ctrl-C in the kairos-installer TUI and gets a getty started on the same console the TUI is drawing on, instead of the agent dying as it did before this PR. On the install path the same handler is now registered twice, here and again at install.go:144.

Move the six-line utils.OnSignal block from the top of AutoInstall to just after the `if err != nil || !autoInstallRequested(cc) { return false, nil }` check at install.go:120-122, so it is only registered on the branch that actually installs. That also drops the duplicate registration on the install path, since Install registers its own at install.go:144. This means the ensureDataSourceReady wait inside AutoInstall is no longer covered by a handler, which matches interactive-install's behaviour before this PR; if you want that wait covered on the install path, register the handler in root.go's install action instead of inside AutoInstall.

agent/pkg/cmd/root.go
```diff
@@ -548,7 +548,16 @@ This command is meant to be used from the boot GRUB menu, but can be also starte
 				log.SetLevel("debug")
 			}
 
-			return agent.InteractiveInstall(c.Bool("shell"), c.String("source"), log)
+			source := c.String("source")
+
+			// install.auto wins over the UX: an unattended config leaves
+			// nothing to ask, so no installer is launched. Same call, same
+			// answer, as the install-mode entry.
+			if installed, err := agent.AutoInstall(source, false, constants.GetUserConfigDirs()...); installed || err != nil {
```
itxaka-agent, line 556: `kairos-agent interactive-install --shell` with `install.auto: true` in the config silently drops --shell. AutoInstall returns true, the action returns here, and the utils.Shell() call in InteractiveInstall (agent/internal/agent/interactive_install.go:27-29) is never reached. The user gets the press-enter prompt and getty@tty1 from AutoInstall, which is not a child shell on the current terminal.

Either honour the flag -- when installed is true and err is nil and c.Bool("shell") is set, return utils.Shell().Run() instead of nil -- or state in agent/docs/installer-contract.md that --shell applies only when an installer is launched, so the unattended path ignores it. Picking one is fine; silently ignoring a flag the user passed is not.

agent/internal/agent/auto_install_test.go
```diff
@@ -0,0 +1,78 @@
...
+	// when the install itself fails. The error is the user check RunInstall
+	// performs first, which is as far as an install gets without a real disk:
+	// reaching it is what proves the branch fired.
+	It("takes over the boot when install.auto is set", func() {
+		writeConfig("#cloud-config\ninstall:\n  auto: true\n  device: /dev/nonexistent\n")
+
+		installed, err := AutoInstall("", false, configDir)
+		Expect(installed).To(BeTrue())
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("user"))
```
itxaka-agent, line 76: This spec drives the real RunInstall and asserts the error text contains "user", which only holds when CheckConfigForUsers fails. That check first stats /etc/kairos/.nousers on the real filesystem (agent/pkg/config/config.go:111-115; config.Scan sets Fs: vfs.OSFS at agent/pkg/config/config.go:44). On any host where that sentinel file exists the check returns nil, the spec falls through CheckConfigForExtraPartitions into the actual install action against /dev/nonexistent, and the assertion fails with an unrelated error.

Assert that the branch fired without executing a real install. Add a package-level seam in install.go, e.g. `var runInstallFn = RunInstall`, have AutoInstall call runInstallFn(cc), and in this spec swap it for a recorder that flags it was called and returns a sentinel error. Then assert installed is true and the error is that sentinel, instead of depending on the host filesystem and on RunInstall's first failing check staying where it is.

itxaka-agent (changes requested):
> action taken here.

The interactive/unattended merge itself is faithful: I traced the old inline auto-install branch in `Install` against the new `AutoInstall` line by line, and the behavior (RunInstall, the press-enter prompt, `getty@tty1`, swallowing a scan error as "no unattended install") is preserved. Mudler's call-site ask is implemented literally. Both commits are clean and signed off, and the docs update matches the code.

Four findings block approval. Three are posted inline below. The fourth is not diff-anchorable (`agent/internal/agent/install.go:173`, the `ensureDataSourceReady()` call inside `Install`, sits in unchanged code between this PR's two hunks in that file, outside the range GitHub's review API will attach a line comment to), so it goes here instead: on the install path, `ensureDataSourceReady()` now runs twice, once inside the new `AutoInstall` (install.go:115) and again at install.go:173 a few lines later in `Install`. Its select only stats `/run/.userdata_load` on a 500ms tick, so the second call costs at least another half second on every boot with no effect, since the first call already waited for the file to disappear. The same duplication applies to the `config.Scan` right after it at install.go:179: it repeats the scan `AutoInstall` just did, and `collector.Scan` follows `config_url` chains over HTTP (sdk/collector/collector.go:253), so a remote config gets fetched twice per boot. Delete the `Install`-side `ensureDataSourceReady()` call; `root.go:635` is the only caller of `Install` and it always runs `AutoInstall` first, which already waited. If re-scanning in `Install` is deliberate because the provider flow needs a `*sdkConfig.Config` that `AutoInstall` does not return, have `AutoInstall` return the scanned config and pass it in, or say so in the comment at install.go:177.

itxaka-agent:
> action taken here.

## Audit summary -- kairos-io/kairos#4588

**Round 0 -- reviewer.** Read the full diff, the two commits, the linked issue (#4340), and the prior review thread (mudler's inline comment and ci-robbot's follow-up commit `bcc54232` addressing it). Pulled the branch into a local worktree to read the surrounding code in `agent/internal/agent/install.go`, `agent/internal/agent/interactive_install.go`, and `agent/pkg/cmd/root.go` beyond the diff hunks.

Verdict: **changes-requested**, four findings.

1. `agent/internal/agent/install.go:106` -- the new `AutoInstall` registers a SIGINT/SIGTERM handler before checking whether an unattended install is even requested, so a Ctrl-C during the interactive TUI now starts a getty on the console the TUI is drawing on, where before it just killed the agent. [Posted inline](https://github.com/kairos-io/kairos/pull/4588#discussion_r3986676626).
2. `agent/internal/agent/install.go:173` (not diff-anchorable, posted in the review body) -- on the install path, the config scan (`ensureDataSourceReady` + `config.Scan`) now runs twice: once inside `AutoInstall`, again a few lines later in `Install`. No effect beyond wasted time and, for a remote `config_url`, a second HTTP fetch per boot.
3. `agent/pkg/cmd/root.go:556` -- `interactive-install --shell` silently drops `--shell` when `install.auto` fires: the unattended branch returns before `InteractiveInstall`'s shell handling is ever reached. [Posted inline](https://github.com/kairos-io/kairos/pull/4588#discussion_r3986676633).
4. `agent/internal/agent/auto_install_test.go:76` -- the new spec asserts on real `RunInstall`'s error text depending on `/etc/kairos/.nousers` not existing on the host filesystem; on a host where it does exist the assertion fails for an unrelated reason. Needs a seam (a swappable `runInstallFn`) instead of depending on host state. [Posted inline](https://github.com/kairos-io/kairos/pull/4588#discussion_r3986676636).

None of the findings rested on toolchain-behaviour claims, so no builder verification round was needed this round.

This is a third-party PR (author `ci-robbot`), so no coder/tester/docs loop runs on our side -- the review and its four findings are the full output of this pass. The ticket moves to `awaiting-author`; the next slot re-polls for a new commit or reviewer/maintainer activity before deciding whether a fresh round is warranted.

mudler (dismissed):
This sounds logical to me. If a cloud config is setting automatic installation, the installer should be automatic. The TUI has to respect that as we are going to unify to a single installer which picks what needs to happen. we might want to add a flag to force installation and skip the check in case a human wants manually trigger it and skip the unattended config

mudler:
@ci-robbot add a flag to skip the unattended install and preserve the current behavior. this might be still useful for a human operator that is trying the ISO manually and investigating , so in that case should go with a separate flag,  documented, which shouldn't be the default (default should be this PR behavior).

agent/internal/agent/install.go
ci-robbot: Fixed in `ecb05b5c`: `utils.OnSignal` moved below the `install.auto` check, so interactive-install no longer runs with the default kill disabled and a handler that would getty over the TUI. That also drops the duplicate registration on the install path, since `Install` still registers its own.

agent/pkg/cmd/root.go
ci-robbot: `ecb05b5c` logs that `--shell` was ignored, and documents in `agent/docs/installer-contract.md` that it applies only when an installer is launched. `--skip-auto-install` (mudler's request in this thread) is the way to get both the installer and the shell on a node whose config sets `install.auto`.

agent/internal/agent/auto_install_test.go
ci-robbot: Done in `ecb05b5c`, as the `runInstallFn` seam you described. The spec now asserts the branch fired by getting its sentinel error back, and additionally that the config handed to the install is the one that was scanned.

PR #4591, opened 2026-09-10
feat(agent): run a first-boot cloud-config stage

### LINKED ISSUE #444: Add first-boot event
while this can be deducted and actually raised in the golang sdk, we don't have any as part of the cloud-config

A 'first-boot' event stage that kicks in only on first boot and be available to be configured via cloud-config

itxaka-agent:
> action taken here.

Fixes: #444

Adds a `first-boot` yip cloud-config stage: a `stages: { first-boot: [...] }` block now fires exactly once, on the node's first boot, through the same `first-boot`/`first-boot.before`/`first-boot.after` mechanism (`agent/pkg/utils.RunStage`) that already drives `boot`/`network`/`fs`/`reconcile`.

Runs last inside the existing sentinel-gated block in `agent.Run()` (`agent/internal/agent/agent.go`) — after the two existing Go-only first-boot hooks (`BundleFirstBoot`, `GrubFirstBootOptions`), so cloud-config sees a fully-provisioned node, and before `machine.CreateSentinel("firstboot")` is written. A failure in the stage is logged, not propagated: on a `strict: true` node a broken first-boot cloud-config no longer puts the agent into a restart loop (bundles re-installed, grub options rewritten, `EventBootstrap` never published) every 5 seconds — it boots through and the failure is visible in the agent log instead.

The previously-proposed mechanism, `machine.ExecuteInlineCloudConfig`, turned out to be dead code — it shells out to an `elemental` binary that doesn't exist in Kairos v4 — so this uses `agent/pkg/utils.RunStage` instead, matching every other stage.

Full history is on the linked issue: full audit trail of what each role (coder/tester/docs/reviewer) did across 2 review rounds is in the summary comment on #444.

One open item for a maintainer or follow-up PR: there's no place in this repo that enumerates valid yip stage names for `first-boot` to join (checked README, docs/, CONTRIBUTING.md, schema — `sdk/schema/root_schema.go`'s `Stages` map is unvalidated free-form). The user-facing stage reference lives in kairos-io/kairos-docs, which is out of scope for this agent's run — flagging so a docs PR there isn't forgotten.

### DIFF
--- a/agent/internal/agent/hooks/firstboot.go
+++ b/agent/internal/agent/hooks/firstboot.go
@@ -15,8 +15,13 @@ type FirstBootStage struct{}
 
 func (b FirstBootStage) Run(c sdkConfig.Config, _ sdkSpec.Spec) error {
 	c.Logger.Logger.Info().Msgf("Running %s hook", cnst.FirstBootHook)
+	// A failure here is logged, not returned. Propagating it would abort the
+	// sentinel-gated block before the firstboot sentinel is written, so a node
+	// with strict mode on would re-run this hook (and re-install bundles, and
+	// rewrite grubenv) on every restart of the agent unit and never publish its
+	// bootstrap event. RunStage already logs the details for the operator.
 	if err := utils.RunStage(&c, cnst.FirstBootHook); err != nil {
-		return err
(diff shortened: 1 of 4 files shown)


Itxaka (approved):
seems useful I guess.

PR #4602, opened 2026-09-10
fix(tests): set cpu type

thebhdn:
Hola chicos,

While running tests locally I had issues with system couldn't boot at all:

```bash
[    7.300364] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00
[    7.302649] CPU: 1 PID: 1 Comm: init Not tainted 5.14.0-687.45.1.el9_8.x86_64 #1
[    7.304578] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Arch Linux 1.17.0-2-2 04/01/2014
.............................................................................................................
[    7.471100] ---[ end Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00 ]---
```

with the help of @jimmykarily and @Itxaka root cause was found and it was related to qemu cpu support. When no `-cpu` flag set it uses generic qemu x86-64-v1 model, while rocky has a minimum architecture baseline of x86-64-v2.

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -423,6 +423,7 @@ func defaultVMOptsNoDrives(stateDir string) []types.MachineOption {
 
 	memory := getEnvOrDefault("MEMORY", "2048")
 	cpus := getEnvOrDefault("CPUS", "2")
+	cpuType := getEnvOrDefault("CPU_TYPE", "host")
 	arch := getEnvOrDefault("ARCH", "x86_64")
 
 	opts := []types.MachineOption{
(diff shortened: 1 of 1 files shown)


Itxaka (dismissed):

tests/tests_suite_test.go
```diff
@@ -488,6 +490,12 @@ func defaultVMOptsNoDrives(stateDir string) []types.MachineOption {
 				)
 			}
 
+			if  os.Getenv("CPU_TYPE") == "" && m.Arch == "x86_64" {
+				m.Args = append(m.Args,
+				 	"-cpu","host",
+				)
+			}
```
jimmykarily: A bit cleaner imho: You can do this above line 429 and set `cpuType` to "host" when it's empty. Then pass cpuType on line 434 without a care in the world.

jimmykarily (dismissed):

tests/tests_suite_test.go
thebhdn: lemme

Itxaka (approved):

PR #4490, opened 2026-09-03
ci(tests): cap the decentralized-k8s ceilings, make the suite timeout explicit

ci-robbot:
Fixes #4489.

## What was wrong

Two numbers that had to fit inside each other, and were never compared.

`reusable-qemu-test.yaml` invoked ginkgo with no `--timeout`, so ginkgo applied its own default of **1h**. That constant appears nowhere in the workflow or in the log. The `provider-decentralized-k8s` spec meanwhile allowed **30m** to reach `active_boot`, **25m** for a working kubeconfig and **15m** each for `Ready`, roles and the machines API.

So no single assertion in that spec can fail on its own. Two slow steps eat most of the hour, the next one cannot fit its ceiling in what remains, and the suite ends with:

```
[TIMEDOUT] A suite timeout occurred
  At [By Step] checking if it has machines with different IPs [1] (Step Runtime: 7m48.092s)
Ran 1 of 34 Specs in 3608.274 seconds
FAIL! - Suite Timeout Elapsed
```

Two costs. A suite timeout skips the failure closures these steps carry: the machines-API and DNS steps each dump the edgevpn API state, `resolv.conf` and the `edgevpn@kairos` unit status on assertion failure, and none of that is printed on a suite timeout. And it burns a full hour of a GitHub-hosted runner on whichever pull request happened to trigger it: [run 33747897800](https://github.com/kairos-io/kairos/actions/runs/33747897800/job/100674733506) spent 60 minutes this way on #4457, an immucore mount-retry change, with the other 67 checks green.

## The measurement the issue was missing

I listed lowering the ceilings as an option on #4489 but would not guess at numbers without knowing what a green run costs. Here it is, pulled from the `By` step timestamps of [run 33755087874](https://github.com/kairos-io/kairos/actions/runs/33755087874), a passing `provider-decentralized-k8s` on the same `ubuntu-24.04` runner:

| step | measured | ceiling was | ratio |
| --- | ---: | ---: | ---: |
| waiting until ssh is possible | 30.7s | 20m | 39x |
| installing, to `active_boot` | 119.8s | 30m | 15x |
| checking if k3s was configured | 87.3s | 10m | 7x |
| working kubeconfig (incl. `Ready`) | 2.4s | 25m + 15m | 625x |
| checking roles | 0.8s | 15m | 1125x |
| machines with different IPs | 0.3s | 15m | 3000x |
| rebooting the VMs | 61.1s | 20m | 20x |
| propagating dns | 0.6s | 4m | 400x |

[...]

Issue #4489, opened 2026-09-03
CI: the decentralized-k8s spec is given 60m of ginkgo suite timeout but asks for 100m of its own


ci-robbot:
## The problem

`provider-decentralized-k8s` fails intermittently with a ginkgo **suite** timeout rather than a failed assertion. Most recent one I saw: [run 33747897800](https://github.com/kairos-io/kairos/actions/runs/33747897800/job/100674733506), on an unrelated immucore PR (#4457).

```
[TIMEDOUT] A suite timeout occurred
  At [By Step] checking if it has machines with different IPs [1] (Step Runtime: 7m48.092s)
Ran 1 of 34 Specs in 3608.274 seconds
FAIL! - Suite Timeout Elapsed
```

## Why

3608 s is ginkgo's default suite timeout of one hour. `.github/workflows/reusable-qemu-test.yaml:338` invokes ginkgo without `--timeout`:

```yaml
go run github.com/onsi/ginkgo/v2/ginkgo -v --label-filter "${{ inputs.test }}" --fail-fast -r ./...
```

The spec `installs to disk with custom config` (`tests/provider_decentralized_test.go:52`) cannot fit in that hour. Its own sequential `Eventually` ceilings:

| step | line | ceiling |
|---|---|---|
| boot to `active_boot` | :72 | 30m |
| working kubeconfig | :164 | 25m |
| `Ready` | :170 | 15m |
| roles | :181 | 15m |
| machines with different IPs | :205 | 15m |
| **total** | | **100m** |

So the spec is allowed to ask for 100 minutes inside a 60 minute budget. Any run where the edgevpn mesh converges slower than usual hits the suite timeout first.

## Why it matters

Two things, both about diagnosis rather than duration.

The failure never names the assertion that was actually slow, only that the suite ran out of clock, so each occurrence has to be re-derived from the progress report. And it lands on whichever PR happened to trigger it, which reads as that PR breaking the p2p tests. The run above is an immucore mount-retry change with the other 67 checks green.

The observed timings on that run: 10m on the kubeconfig step, 25m on roles, then 7m48s into the IP step when the suite timeout fired. Two of the five steps consumed 35 of the 60 minutes on their own.

## Options

Not sure which is wanted, so listing rather than picking:

1. Pass an explicit `--timeout` to ginkgo that exceeds the sum of the per-step ceilings, and let `timeout-minutes` on the job be the real backstop. Makes failures name their step, does not make anything faster.
[...]

PR #4615, opened 2026-09-11
fix(kairos-init): add systemd-resolved to more distros' initramfs, safely

### LINKED ISSUE #835: No DNS at initramfs stage
At initramfs state, there is no DNS setup and it's not possible to resolve domains. Kairos config file may include things like the kcrypt challenger server, which may be served on a specific domain. No DNS means, those services won't be accessible resulting in a system that doesn't boot correctly.

Our documentation mentions some ways to configure DNS (https://kairos.io/docs/reference/configuration/#syntax) but this doesn't seem to work on systemd based systems at least on the initramfs stage.

Users have workarounds like this:

```
stages: 
  boot.before:
     - name: "DNS settings"
       commands:
       - echo "DNS=8.8.8.8 8.8.4.4" >> /etc/systemd/resolved.conf
       - echo -e "\n[DHCP]\nUseDNS=false" >> /etc/systemd/network/20-dhcp.network
       - systemctl restart systemd-networkd
       - systemctl restart systemd-resolved
```
(thanks @christianprim) 

but we should provide a simpler way to configure DNS on any system.

itxaka-agent:
> action taken here.

Fixes: #835

## What

Some Kairos targets never got the `systemd-resolved` dracut module added to their initramfs, so DNS never resolved that early in boot — the "no DNS at initramfs stage" half of #835 (the other half, the `dns:` cloud-config stage on `systemd-resolved` hosts, lives in `mudler/yip` and is out of scope here; see the takeover comment on the issue).

`kairos-init/pkg/stages/steps_init.go` now adds `systemd-resolved` to the dracut network-module list wherever `systemd-networkd` drives the initramfs and the module is actually available on the build root (both the `systemd-resolved` daemon binary and `resolvectl` present, matching what the module's own `check()` requires). This closes the gap for Ubuntu 21.04–22.04*, Debian, and openSUSE/SLES.

\* **Ubuntu 22.04 is not fixed by this PR.** Its packaged dracut (051, across every jammy suite) predates the upstream `systemd-resolved` dracut module, which only landed at dracut 054. No installable package closes this gap today; that's documented in a code comment rather than silently left as a mystery or forced (forcing an unavailable module makes dracut abort the whole image build — see below).

Also fixed along the way: the module-availability probe originally checked only the `systemd-resolved` daemon binary. The module's own `check()` requires `resolvectl` too, and dracut hard-aborts the entire image build (no initramfs produced) when a module forced via `add_dracutmodules` fails its `check()`. Left unchanged, on a root with the daemon but not `resolvectl`, this fix would have turned "missing DNS" into "build fails outright" — worse than the original bug. The probe now checks both binaries.

The RedHat family (outside Fedora-with-networkd) and Ubuntu ≤20.04 are deliberately left alone: their dracut modules (`network-manager`, `network`/`network-legacy`) already write `/etc/resolv.conf` themselves, so they already have working DNS in the initramfs.

## Verification

Unit tests cover the module-selection logic (table-driven, `kairos-init/pkg/stages/steps_init_test.go`) against both fabricated build roots and facts confirmed against real, published Kairos container images (`quay.io/kairos/core-*`) and the real Ubuntu package archive for the jammy-dracut-version claim. No ISO was built and no VM was booted for this ticket — the change is internal to kairos-init's build-time dracut module selection, with no existing isolated repro short of the full image-build pipeline.

Full round-by-round audit trail (what each review round found, and how each finding was resolved or empirically verified): https://github.com/kairos-io/kairos/issues/835#issuecomment-5632796718

### DIFF
--- a/kairos-init/pkg/stages/steps_init.go
+++ b/kairos-init/pkg/stages/steps_init.go
@@ -910,6 +910,13 @@ func dracutNetworkModules(root string, sis values.System, l logger.KairosLogger)
 	// Otherwise other modules provide their own resolvers, while networkd
 	// leaves /etc/resolv.conf to resolved, so an initramfs that has networkd
 	// and no resolved cannot resolve names at all.
+	//
+	// Ubuntu 22.04 keeps the plain networkd list even though it matches every
+	// other condition: jammy only ever published dracut 051 (universe, with
+	// nothing newer in -updates, -backports or -security) and dracut grew the
+	// systemd-resolved module in 054, so no package on that release can put
+	// the module into the build root. Getting resolved into the 22.04
+	// initramfs needs a newer dracut for jammy, not a change here.
 	if strings.Contains(networkModule, dracutModSystemdNetworkd) &&
 		!strings.Contains(networkModule, dracutModSystemdResolved) &&
 		resolvedModuleAvailable(root) {
--- a/kairos-init/pkg/stages/steps_init.go
+++ b/kairos-init/pkg/stages/steps_init.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -32,6 +33,14 @@ const (
 	dracutModSystemdNetworkd = "systemd-networkd"
 	dracutModSystemdResolved = "systemd-resolved"
 	serviceSSHD              = "sshd"
+)
+
+// Paths, relative to the root of the image being built, that tell which
+// network daemons the initramfs can use.
+const (
+	networkManagerBinary  = "usr/sbin/NetworkManager"
+	systemdNetworkdBinary = "usr/lib/systemd/systemd-networkd"
+	systemdResolvedBinary = "usr/lib/systemd/systemd-resolved"
 )
 
 // GetInitrdStage Returns the initrd stage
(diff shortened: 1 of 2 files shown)


itxaka-agent:
> any action taken here.

Pushed `36db0320` addressing the gap flagged above: Ubuntu >=24.04, Ubuntu >=26.04, and Hadron were each forcing `systemd-resolved` onto the dracut module list unconditionally, bypassing `resolvedModuleAvailable` entirely, so a build root missing `resolvectl` or the dracut module directory would abort the whole image build instead of degrading gracefully. All three now flow through the same guard. Added 3 regression test rows (one per site) covering the missing-prerequisite case.

**test run** — `go build ./...` and `go test -count=1 -v ./pkg/stages/...` at `36db0320`, 2026-09-21T12:33:19+02:00: 6/6 + 18/18 + 6/6 passing, zero FAIL.

test run

blob sha: `85afe1f6b213d7e33a835175d6ac5206055feefa`

Itxaka:
@ci-robbot please check that and apply the qa passed label if it passes your tests

itxaka-agent:
> action taken here.

Rebased onto `master` to pick up the new `make verify-fips` target (added upstream after this branch's merge-base): the FIPS build jobs were failing with `make: *** No rule to make target 'verify-fips'. Stop.` No functional changes — new head `67b4bf48`. Build and `go test ./pkg/stages/...` pass locally after the rebase.

itxaka-agent:
> any action taken here.

CI turned up a new, unrelated failure on this branch: `test-core-ubuntu-20 / acceptance / acceptance` failed with `curl: (6) Could not resolve host: google.it`. Root cause was in `kairos-init/pkg/bundled/cloudconfigs/05_network.yaml`, not in this PR's dracut module logic: the resolv.conf symlink stage ran a bare `rm /etc/resolv.conf` before the `ln -s`, and Ubuntu 20.04's base rootfs ships with no `/etc/resolv.conf` at all, so `rm` failed, the stage aborted, and the symlink never got created.

Fixed by collapsing both commands into a single `ln -sf`, matching how every other symlink stage in the bundled cloud-configs already does it. New head: `4242dab7`.

itxaka-agent:
> any action taken here.

CI failed across roughly 18 unrelated jobs (install, reset, upgrade, provider variants, selinux) with symptoms pointing at host contention — a kernel clocksource watchdog stall and a duplicate `pcspkr` driver registration during VM boot — not a defect in this branch. Rebased onto current `master` (2 commits behind) and force-pushed to get a clean CI run; `go build ./...` and `go test ./pkg/stages/...` stay green on the rebased tree. New head `1916453f`.

itxaka-agent:
@mudler CI red on this branch is pre-existing — same install/acceptance/install-target/test-core-fips-install/test-core-ubuntu-20-install failures landed on #4726 (merged Sep 18) before I touched anything.

Suite looks broken on master. Fine to merge this once you're OK with the change itself; the failures aren't mine.

itxaka-agent:
> any action taken here.

Investigated the one red leg on this run: `test-standard / provider-decentralized-k8s`. It's a ginkgo suite-timeout (`[TIMEDOUT] A suite timeout occurred`, 3613s), not a failed assertion — the goroutine died mid-test on an SSH dial (`ssh: handshake failed: EOF`), and the same handshake failure then hits every VM-teardown diagnostic command. That's a VM/runner losing reachability partway through a 5-VM decentralized-k8s run, not a code assertion failing.

This is the same failure mode already tracked in #4489 (fixed by #4490: per-step ceilings capped at 10m, explicit `--timeout`). Nothing in this PR's diff — `kairos-init`'s dracut module selection and the `05_network.yaml` resolv.conf symlink, both build-time/early-boot-only — runs anywhere near the point in a live, already-networked k3s cluster where this died.

No rebase needed: the branch is already at `upstream/master`'s tip (0 commits behind). `gh run rerun --failed` isn't available to this token (403, admin-only), same as the last two rounds. Rest of the matrix (~30 legs) is still queued; flagging so a maintainer can re-run the leg if the rest comes back green.

itxaka-agent:
> any action taken here.

Same run, now further along: 13 legs red (`install-target`, `ssh-hardening`, both `test-core-fips` legs, both `test-core-ubuntu-20` legs, `rocky-9`/`leap` selinux, and four `test-standard` provider legs), 54 green, 5 still running. Checked three of the new failures directly (`install-target`, `test-core-rocky-9/selinux`, `test-core-fips/install`): all three show the guest reaching a console login prompt within seconds of boot, then the SSH harness timing out for 750–2000+s trying to reach it (`ssh: handshake failed ... i/o timeout`). Boot itself isn't hanging — the machine comes up, the test just can't dial in.

That's a host/runner reachability problem, not a boot-flow issue this PR's dracut/network-module diff could cause (that code only runs pre-switch_root, before any of this). Same category of failure as the two previous rounds on this PR (clock-watchdog stalls, an SSH-dial timeout inside an already-networked cluster test) — consistent with shared-runner contention rather than a regression. Every test that doesn't boot a full VM (partition-validation, custom-partitioning, insecure-registry, encryption-*, uki-*) is green.

No code action taken. Branch is already at `upstream/master`'s tip, so there's nothing to rebase, and `gh run rerun --failed` is still 403 for this token. Leaving this for a maintainer to re-run if the rest of the matrix doesn't recover on its own.

itxaka-agent:
> any action taken here.

Run 35696064823 finished. The new failures beyond my last update (acceptance, install, reset, bundles, upgrade-with-cli) show the same signature I already flagged: `docker digest-mismatch` + `Failed to resolve group 'kvm'` in the pre-boot setup, then `Boot failed: not a bootable disk` in the VM. That's a runner-side image/registry fault, not this PR's dracut diff.

Branch is already at upstream tip, so there's nothing to rebase to retrigger. `gh run rerun --failed` is still 403 for this token. No action needed here from my side — flagging for a maintainer to re-run the job when convenient.

itxaka-agent:
> action taken here.

Checked the failing checks on this PR. They aren't caused by this change: the same "ssh: handshake failed: EOF" / "Timed out after 750s" failure reproduces identically on two unrelated branches ([ci/flavor-build-matrix](https://github.com/kairos-io/kairos/actions/runs/35695677635), [fix/4817-recovery-ssh-flags](https://github.com/kairos-io/kairos/actions/runs/35694283871)) whose diffs don't touch dracut or network config, and two of the failing lanes here (Ubuntu 20.04, Rocky 9) never reach this PR's new code path — their dracut module list is unchanged by this diff. The guest does boot in this PR's own run; the console reaches a login prompt, only sshd never comes up afterward. Looks like a repo-wide QEMU test-lane issue that started around 2026-09-21, not a regression from this change. Branch unit tests are green. Will re-check once the lanes are healthy again.

PR #4651, opened 2026-09-15
ci: scheduled GHCR age-based prune

jimmykarily:
## Why

GHCR has no server-side TTL equivalent to quay's `expires-after` label, so the container versions our CI pushes accumulate indefinitely. The install-upgrade UKI test alone lands a fresh uniquely-tagged version on `kairos-uki` every run; the per-matrix ISO OS image tags on `hadron`, `hadron-fips`, `hadron-trusted`, `ubuntu`, `rockylinux`, and `opensuse` accumulate one set per commit; `kairos-init:pr-N`, `kcrypt-challenger:pr-N`, and `bundles-test:pr-N` accumulate one tag per PR number forever. Dry-runs against real GHCR surfaced ~13k versions ripe for cleanup.

## What this does

Reintroduces quay's `expires-after` model as an outside pruner. Two files, one Go tool, one scheduled workflow — no per-workflow wiring anywhere else in the CI:

**`.github/actions/ghcr-cleanup/`** — a self-contained Go module (stdlib runtime, ginkgo/gomega for tests only) with two subcommands:
- `delete` — takes a newline-separated `IMAGES` list of `ghcr.io/<owner>/<pkg>:<tag>` refs and deletes every matching version. Safe for ad-hoc use.
- `prune` — age-based sweep driven by `PACKAGES` (per-package retention pairs) and `PROTECTED_TAGS` (shell globs). For each package, lists every version and:
    - keeps any tagged version whose tag list overlaps a protected glob or is younger than the retention;
    - fetches the manifest of every kept-tagged version via the OCI distribution API (anonymous token exchange, kairos-io packages are public) and records child digests of image indexes / manifest lists;
    - keeps any untagged version referenced by that set (multi-arch children of tagged parents stay pullable);
    - deletes everything else.

Aged tagged parents that get deleted this run leave their children unreferenced going forward; those get cleaned up on the next run. Post-promote `:master` and `:master-scratch` sharing a package version is handled — any protected tag on the version protects the whole version.

Retention parser accepts `h`, `d`, `w`. Concurrency: worker pool bounds the manifest-fetch phase (dominant wall-time on packages with many kept versions). Progress logging every 500 listed versions and every 50 fetched manifests so a first run doesn't look hung. 42 ginkgo specs, race-clean. Nested `go.mod` wired into `make test` via a new `test-actions` target.

**`.github/workflows/ghcr-prune.yaml`** — daily 03:00 UTC cron plus `workflow_dispatch` with a `dry_run` toggle. Two visible YAML arrays in the job env are the entire config surface:

- `PACKAGES` — one `<ref-prefix>  <retention>` pair per line. 10d for the OS flavors + `kairos-init` + `kcrypt-challenger` (softened from quay's 2w for the current PR load); 6h for `bundles-test` and `kairos-uki` (matching what quay's `--label quay.expires-after="6h"` used to say).
- `PROTECTED_TAGS` — `master`, `latest`, `master-scratch`, `v[0-9]*`. Any version whose tag list matches any glob is kept regardless of age.

## Explicitly out of scope

- **Hadron releases** (`ghcr.io/kairos-io/hadron:v*`): not in `PACKAGES`, so untouched. Belt-and-suspenders `v[0-9]*` protected glob covers any future release tag that lands on a package in the list.
- **Real releases from this repo**: `release.yaml` pushes to `quay.io`, not GHCR. Not touched.
- **Private packages**: manifest fetch uses anonymous token exchange; if we ever prune a private package, the token exchange needs credentialed auth.

## Test plan

- [x] `make test-actions` green
- [x] `go test -race ./...` green
- [x] `actionlint -shellcheck ''` green
- [x] Local dry-run against real GHCR — reports the expected ~13k candidates without erroring, no unresolved/failed warnings
- [ ] First scheduled run after merge — verify the log looks reasonable before letting daily runs continue


jimmykarily:
Small caveat worth flagging: `cleanup-ghcr` deletes by the shared `pr-N` tag, so if a new push cancels an in-flight run and the old run's tests are slow to cancel, its cleanup can fire after the new run has already re-pushed `pr-N` and end up deleting the new image (the new run's tests then fail with an image-pull error, and a re-run fixes it). Rare in practice because `cancel-in-progress: true` keeps only one run active per PR; long term we could close the window by tagging per run or deleting by digest instead of tag.

itxaka-agent:
> action taken here.

Picking this up for review. It adds a reusable `ghcr-cleanup` composite
action (Go program + Ginkgo tests) and wires it into `_uki-test.yaml`,
`pr.yaml`, and `master.yaml` to delete accumulating GHCR image tags. I'll
pull the branch, read the diff and commit history, and check the cleanup
logic against the workflow sites it's wired into (tag aggregation, the
:master-scratch exclusion, fork-PR gating).

.github/workflows/pr.yaml
```diff
@@ -493,6 +493,81 @@ jobs:
...
+      - test-core
+      - test-core-fips
+      - test-core-ubuntu-20
+      - test-core-rocky-9
+      - test-core-leap
+      - test-standard
+      - test-standard-encryption
+      - test-uki
+      - build-iso-arm
+    if: always()
```
itxaka-agent: This job checks out refs/pull/N/head with allow-unsafe-pr-checkout: true (lines 524-527) and then runs uses: ./.github/actions/ghcr-cleanup (line 566), which resolves action.yml and main.go out of that PR-authored working tree and executes them with packages: write and secrets: inherit. if: always() means the job's execution is not conditioned on the authorize gate succeeding, so a fork PR whose deployment review does not come back green still reaches this step. The file header at lines 14-27 states the authorize job is the only thing keeping the CodeQL 'checkout of untrusted code in a privileged context' findings theoretical, and the precedent job size-diff at line 580 carries no if: precisely so it cannot run unless its dependency chain -- and therefore authorize -- succeeded.


Add authorize to the needs: list and condition the job on it, so always() only relaxes the test-outcome dependency and not the gate: if: always() && needs.authorize.result == 'success'. That keeps the intended behaviour (cleanup runs whether or not the qemu suites passed) while restoring the invariant the header demands.

.github/workflows/pr.yaml
```diff
@@ -493,6 +493,81 @@ jobs:
...
+      - name: Aggregate images to clean up
+        id: aggregate
+        shell: bash
+        run: |
+          set -euo pipefail
+          {
+            echo "ghcr.io/kairos-io/kairos/kairos-init:pr-${{ github.event.pull_request.number }}"
+            echo "ghcr.io/kairos-io/kairos/kcrypt-challenger:pr-${{ github.event.pull_request.number }}"
+            if [ -d ghcr-refs ]; then
+              find ghcr-refs -type f -name image-ref.txt -exec cat {} +
```
itxaka-agent: find ghcr-refs -type f -name image-ref.txt -exec cat {} + feeds the contents of every downloaded ghcr-image-ref-* artifact into a job that issues DELETE against ghcr.io with packages: write, with no check on what those lines say. Artifact contents are not trusted input in a pull_request_target run: the qemu jobs execute the PR's own ginkgo code, and any of them can upload an artifact matching ghcr-image-ref-* containing arbitrary refs. A line reading ghcr.io/kairos-io/kairos/kairos-init:master or ghcr.io/kairos-io/hadron:v0.5.1 would be deleted for real. This survives maintainer approval of the diff, because the injection lives in test code rather than in the workflow.


Filter the aggregated list before it reaches the action instead of trusting it. Constrain the prefix and reject durable tags -- e.g. pipe images.txt through grep -E '^ghcr\.io/kairos-io/kairos/[A-Za-z0-9._-]+:[A-Za-z0-9._-]+$' and then drop anything whose tag is latest, master, master-scratch, or matches ^v[0-9]. Stronger still: require every artifact-derived tag to end in the run's computed $VERSION or equal pr-${{ github.event.pull_request.number }}, which is the only set this job is meant to delete. Apply the same filter in master.yaml's aggregate step for symmetry, even though its inputs are trusted today.

.github/workflows/pr.yaml
```diff
@@ -493,6 +493,81 @@ jobs:
...
+          pattern: 'ghcr-image-ref-*'
+          path: ghcr-refs
+
+      - name: Aggregate images to clean up
+        id: aggregate
+        shell: bash
+        run: |
+          set -euo pipefail
+          {
+            echo "ghcr.io/kairos-io/kairos/kairos-init:pr-${{ github.event.pull_request.number }}"
```
itxaka-agent: Deleting the ISO OS image tags makes re-running a single failed qemu cell impossible. reusable-qemu-test.yaml:374 passes CONTAINER_IMAGE: ${{ env.IMAGE_NAME }} -- the GHCR OS image tag this run pushed -- into the suites, and the upgrade-with-cli and provider-upgrade-* tests pull it inside the VM. Once cleanup-ghcr has removed that tag, re-running that one cell fails at image pull with no way to recover short of re-running the whole workflow. The qemu matrices here are fail-fast: false and re-running a flaked cell is routine, so this will be hit.


Decide this explicitly rather than leaving it implicit. Concrete option: keep if: always() for the two hard-coded pr-N refs (nothing re-consumes those), but skip the artifact-derived refs when any qemu job did not succeed -- gate the download/aggregate of ghcr-image-ref-* on a condition over needs.*.result, so a red run leaves its OS images in place for the re-run and the next green run cleans them up. If you would rather accept the trade-off, say so in the job comment at lines 496-505 so the next person who re-runs a flaky cell knows why the pull 404s.

.github/workflows/pr.yaml
```diff
@@ -493,6 +493,81 @@ jobs:
...
+            echo "ghcr.io/kairos-io/kairos/kcrypt-challenger:pr-${{ github.event.pull_request.number }}"
```
itxaka-agent: reusable-qemu-test.yaml:356 pushes ghcr.io/kairos-io/kairos/bundles-test:pr-<N> for the bundles test cell, and it is not in this list. It accumulates one tag per PR number forever, which is the exact leak the kairos-init:pr-N line above it fixes.

Add the bundles-test ref alongside the other two hard-coded per-PR tags. Only land this once the re-run question in the previous comment is settled, since it has the same consequence for re-running the bundles cell.


```suggestion
            echo "ghcr.io/kairos-io/kairos/kcrypt-challenger:pr-${{ github.event.pull_request.number }}"
            echo "ghcr.io/kairos-io/kairos/bundles-test:pr-${{ github.event.pull_request.number }}"
```

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,332 @@
...
+				fmt.Printf("::warning::DELETE failed for %s (version %d): %v\n", ref, v.ID, err)
+				failed++
+				continue
+			}
+			deleted++
+		}
+		if len(versions) < pageSize {
+			return deleted, failed
+		}
+		page++
```
itxaka-agent, line 356: Deleting versions while paginating shifts the server-side page window, so page++ can skip versions that were never listed. Concretely: page 1 returns 100 versions and two of them carry the target tag and are deleted; the follow-up request for page=2&per_page=100 now indexes into a 148-item list, so the two versions that were at offsets 100 and 101 are never seen. If either carried the target tag, the tag survives and the report still prints deleted=2, so nothing flags it. The pagination spec at main_test.go:912 does not catch this because its only matching version is on page 2 and nothing is deleted before the boundary -- it never deletes and paginates in the same run.


Collect first, delete second. Walk every page accumulating the matching version IDs into a []int, exit the pagination loop, then issue the DELETEs from that slice. The listing is then never mutated while it is being walked. Add a spec that seeds 150 versions with the same tag on two versions straddling the page boundary (say IDs 50 and 120) and asserts rpt.Deleted == 2.

itxaka-agent (changes requested):
> action taken here.

Reviewed the diff and the four commits directly, not just the description. The `ghcr-cleanup` Go action itself is solid: the nested-package URL encoding, the org-vs-user endpoint fallback, and the shared-tag handling all check out against the code, and the table tests genuinely exercise those paths. All four new pinned `uses:` lines (`actions/checkout`, `actions/setup-go`, `actions/download-artifact`, `actions/upload-artifact`) resolve to the SHAs they claim.

The problems are in the workflow wiring around it, two of which are worth blocking on: the cleanup job in `pr.yaml` isn't gated on the `authorize` job the file's own header says is required before running PR-authored code with `packages: write`, and the artifact-derived ref list it feeds into a privileged DELETE has no allow-list, so a PR's own test code could smuggle an arbitrary ref (including `:master`) into the deletion set. Three smaller findings below (re-run recovery, a missing per-PR tag, a pagination-versus-delete ordering bug). A sixth draft finding (a dotted `.github` path element in `go.mod` possibly breaking `go build`) was verified empirically and does not reproduce -- Go's module-path validation accepts it fine -- so it is not included here.

itxaka-agent:
> action taken here.

**Audit trail — round 0**

1. **Intake.** Picked up this PR for review (unassigned, no prior review from us). Pulled `ci/ghcr-image-cleanup-action` (head `1f6fb4e`) into a scratch worktree.
2. **Reviewer, round 0.** Verdict: `changes-requested`. Read all four commits and the full diff directly, not just the description: verified the nested-package URL encoding, the org-vs-user GHCR endpoint fallback, and the shared-tag handling against the Go source and its table tests, and confirmed all four new pinned `uses:` lines resolve to the SHAs they claim.
3. **Findings.** Five, all confirmed against the actual workflow/Go source:
   - `.github/workflows/pr.yaml:518` — the `cleanup-ghcr` job runs PR-authored Go code with `packages: write` under `if: always()`, without depending on the `authorize` gate the file's own header requires before executing untrusted code.
   - `.github/workflows/pr.yaml:551` — the artifact-derived ref list feeding the privileged DELETE has no allow-list; a PR's own test code could smuggle an arbitrary ref (including `:master`) into an uploaded artifact.
   - `.github/workflows/pr.yaml:548` — deleting the ISO OS image tag unconditionally breaks re-running a single failed qemu cell, since the tag it needs to pull no longer exists.
   - `.github/workflows/pr.yaml:549` — the `bundles-test:pr-<N>` tag is missing from the per-PR cleanup list, so it leaks the same way the fix addresses for the other two. One-click suggestion included.
   - `.github/actions/ghcr-cleanup/main.go:249` — pagination and deletion interleave, so deleting matches on an earlier page can shift later pages and skip versions that were never listed.
4. **Builder verification (rule 9b).** One draft finding claimed a dotted `.github` path element in `go.mod` might break `go build`/`go test`. Routed to the builder: confirmed the build succeeds cleanly on go1.25.4 — Go's module-path validation accepts non-leading dotted path elements. That disproves the finding's premise, so it was dropped before publication.
5. **Review posted.** [Round 0 review](https://github.com/kairos-io/kairos/pull/4651#pullrequestreview-5211605901), `CHANGES_REQUESTED`, 5 inline comments on the diff.
6. **Not high-risk under rule 3a** — CI/workflow-only change, no boot/install/upgrade/reset/kcrypt/immucore code path touched, so no QEMU verification was required.

Ticket stays in `awaiting-author`; the next slot re-polls per rule 12b once the author pushes or responds.

.github/workflows/pr.yaml
jimmykarily: Fixed in `ea5da757`: added `authorize` to `needs:` and gated on `needs.authorize.result == 'success'`, so `always()` now only relaxes the test-outcome dependency and no longer bypasses the fork-pr-builds gate. Header-comment invariant restored.

.github/workflows/pr.yaml
jimmykarily: Fixed in a8daa8e4: aggregate step now requires artifact-derived refs to match `^ghcr.io/kairos-io/kairos/<pkg>:<tag>$` AND embed this run's `git describe --tags --dirty --always` VERSION in the tag. A poisoned line pointing at `master`, `master-scratch`, or a released `v*` tag cannot match — VERSION contains the fork's own head SHA, which the attacker cannot forge onto a package they do not own. Hardcoded pr-N refs skip the filter since they come from the workflow YAML rather than the artifacts. Same shape applied in master.yaml for symmetry even though its inputs are trusted today.

.github/workflows/pr.yaml
jimmykarily: Fixed in a8daa8e4: split into two classes documented in the job header. `kairos-init:pr-N` is build-time only (only build-iso pulls it) and stays "always delete". The test-consumed set — `kcrypt-challenger:pr-N`, `bundles-test:pr-N`, and every artifact-derived OS image tag — is now gated on every qemu cell going green via an `ALL_TESTS_GREEN` env at the job level. On red, those persist so a failed cell can be re-run against the same tags; the next green run cleans up its own set. Accumulation is bounded (pr-N tags overwrite each push; OS tags orphan at N per red run). Same gate mirrored in master.yaml.

.github/workflows/pr.yaml
jimmykarily: Fixed in a8daa8e4: `bundles-test:pr-N` is now in the test-consumed class alongside `kcrypt-challenger:pr-N`, gated on every qemu cell going green so a re-run of the bundles cell still finds its tag.

.github/actions/ghcr-cleanup/main.go
jimmykarily: Fixed in 3395dbec: `deleteMatchingVersions` now walks every page accumulating matching IDs into a slice first, then issues DELETEs after pagination finishes. The listing is never mutated while it is being walked, so a shared-tag ref straddling the page boundary can no longer slip past the cursor. Added a Ginkgo spec that seeds 150 versions with the target tag on IDs 50 and 120 (one per page) and asserts `rpt.Deleted == 2`.

jimmykarily:
The implementation was getting very complicated (trying to keep the images around until the PR is closed in case we want to re-trigger etc). I decide to change the implementation to a scheduled pruning job. Looks simpler now. Easier to inspect and to manually trigger if needed. The tool even has a dry run mode which we can run locally to see what would happen:

```
cd .github/actions/ghcr-cleanup
export GH_TOKEN=$(gh auth token) 
export DRY_RUN=true
export PACKAGES=$'ghcr.io/kairos-io/kairos/hadron 2d
  ghcr.io/kairos-io/kairos/hadron-fips 2d
  ghcr.io/kairos-io/kairos/hadron-trusted 2d
  ghcr.io/kairos-io/kairos/ubuntu 2d
  ghcr.io/kairos-io/kairos/rockylinux 2d
  ghcr.io/kairos-io/kairos/opensuse 2d
  ghcr.io/kairos-io/kairos/kairos-init 2d
  ghcr.io/kairos-io/kairos/kcrypt-challenger 2d
  ghcr.io/kairos-io/kairos/bundles-test 2d
  ghcr.io/kairos-io/kairos/kairos-uki 2d'
export PROTECTED_TAGS=$'master\nlatest\nmaster-scratch\nv[0-9]*' 
go run . prune
```

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,923 @@
...
+		}
+	}
+	return out
+}
+
+// tagMatchesProtected reports whether any protected glob matches the
+// tag. Uses path.Match's shell-glob semantics (*, ?, [range]).
+func tagMatchesProtected(tag string, protected []string) bool {
+	for _, pat := range protected {
+		if ok, _ := path.Match(pat, tag); ok {
```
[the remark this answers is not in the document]
jimmykarily: Legit. `parseProtectedGlobs` now runs each pattern through `path.Match` at parse time and returns an error on `ErrBadPattern`; the prune subcommand aborts with the offending line rather than reaching the DELETE loop with a pattern that silently loses protection. New Ginkgo spec covers both the valid-list and typo-in-glob paths.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,923 @@
...
+				fmt.Printf("[dry-run] would delete %s (%s, id %d, age %s, tags %v, digest %s)\n",
+					spec.Ref, label, cand.id, cand.age.Truncate(time.Hour), cand.tags, cand.digest)
+				rpt.Deleted++
+				continue
+			}
+			fmt.Printf("deleting %s (%s, id %d, age %s, tags %v, digest %s)\n",
+				spec.Ref, label, cand.id, cand.age.Truncate(time.Hour), cand.tags, cand.digest)
+			delURL := fmt.Sprintf("%s/%s/%s/packages/container/%s/versions/%d",
+				c.apiBase, scope, spec.Owner, spec.EncPkg, cand.id)
+			if err := c.deleteVersion(ctx, delURL); err != nil {
```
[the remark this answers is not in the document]
jimmykarily: Considered but not fixing here. The race requires a concurrent promote to add `:master` or `:master-scratch` to a *specific existing package version* that this run has already classified as aged-unprotected. `docker buildx imagetools create` produces the same package version only when the manifest content is byte-identical; different pipeline runs produce different digests (different source SHAs, different binaries), so the tag would land on a fresh new version rather than piggybacking on our stale candidate. Adding a per-DELETE revalidation GET would double the DELETE-phase API budget for a scenario that requires deliberate digest collision. Leaving with this explanation for the record.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,923 @@
...
+		var refMu sync.Mutex
+		var doneMu sync.Mutex
+		done := 0
+		for _, t := range tasks {
+			wg.Add(1)
+			sem <- struct{}{}
+			go func(t fetchTask) {
+				defer wg.Done()
+				defer func() { <-sem }()
+				children, err := mf.referencedDigests(ctx, spec.Owner, spec.EncPkg, t.tag)
```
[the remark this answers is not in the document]
jimmykarily: Legit. `manifestFetcher` now takes a digest (`sha256:...`) instead of a tag, and the referencing pass fetches each unique digest of every tagged version rather than resolving by tag. Two versions carrying the same tag are now inspected separately; the OCI API accepts `manifests/<digest>` verbatim. Fake fetcher and specs updated to key by digest; dedupe spec still passes because two tags on the same digest still collapse to one fetch.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,923 @@
...
+				doneMu.Lock()
+				done++
+				n := done
+				doneMu.Unlock()
+				if err != nil {
+					fmt.Printf("::warning::manifest fetch for %s:%s failed (untagged children on this manifest will be treated as unreferenced): %v\n",
+						spec.Ref, t.tag, err)
```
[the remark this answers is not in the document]
jimmykarily: Legit and important. `collectPrunableVersions` now tracks any manifest fetch failure and skips untagged-version pruning for the whole package on the first error. The tagged pass still runs (it does not depend on the referenced set), so aged unprotected tags still get deleted -- only the multi-arch child path is skipped, which is the one that could break a still-live manifest list. New spec (`fails closed: skips untagged pruning for the whole package if any manifest fetch fails`) locks the behavior in.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,923 @@
...
+		if len(v.Metadata.Container.Tags) != 0 {
+			continue
+		}
+		age := now.Sub(v.CreatedAt)
+		if age < spec.Retention {
+			continue
+		}
+		if referenced[v.Name] {
+			continue
+		}
```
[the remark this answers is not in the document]
jimmykarily: Legit. Restructured to fetch manifests for *every* tagged version (not just kept ones) before classifying untagged. Children of an aged tagged parent that will be deleted this run are therefore still in the referenced set this run and are kept; the next prune runs with the parent already gone and its children unreferenced, at which point they age out. If the parent DELETE fails, the children stay too (same referenced-set logic), so a failed parent no longer leaves broken children behind. Spec `keeps children of every tagged parent (both kept and aged-to-delete), prunes true orphans, defers aged-parent children to the next run` covers it.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,923 @@
...
+		return err
+	}
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	_, _ = io.Copy(io.Discard, resp.Body)
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("DELETE %s: %s", url, resp.Status)
```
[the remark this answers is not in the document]
jimmykarily: Legit and important for the first-run catch-up. Added a rate-limit detection helper that flags 429 responses and `X-RateLimit-Remaining=0` 403s as a sentinel `errRateLimited`. `pruneAged` unwraps that and stops the current run cleanly (no Failed increment; log line points at "the next daily prune"). Ginkgo spec `when the API rate-limits us mid-delete` uses a fakeGHCR that returns 429 after N successful deletes and asserts `Deleted=1, Failed=0`. Combined with the FAIL_ON_ERROR fix below, rate limits are backpressure, real errors are visible.

.github/workflows/ghcr-prune.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
+        with:
+          go-version-file: .github/actions/ghcr-cleanup/go.mod
+          cache: false
+      - name: Prune GHCR
+        working-directory: .github/actions/ghcr-cleanup
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DRY_RUN: ${{ inputs.dry_run == true && 'true' || 'false' }}
+        run: go run . prune
```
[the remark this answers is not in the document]
jimmykarily: Legit. The prune workflow now sets `FAIL_ON_ERROR: 'true'` in the env block. With the rate-limit path landing above no longer counting throttles as `Failed`, this trips only on real DELETE errors (permission mismatch, GHCR 500s), and missing-package warnings stay in the `Unresolved` bucket which does not gate the exit code. Incomplete cleanup that is NOT rate-limit-driven now turns the job red.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,1002 @@
...
+			continue
+		}
+		toDelete, err := c.collectPrunableVersions(ctx, scope, mf, spec, protected, now)
+		if err != nil {
+			if errors.Is(err, errRateLimited) {
+				fmt.Printf("::warning::rate-limited while listing %s; stopping this run so the next daily prune can pick up.\n", spec.Ref)
+				return rpt, nil
+			}
+			fmt.Printf("::warning::listing failed for %s: %v\n", spec.Ref, err)
+			rpt.Failed++
```
itxaka-agent, line 667: runPrune bounds the whole sweep with a 30-minute context at main.go:205. When that deadline expires mid-run, every remaining request fails with context.DeadlineExceeded, which is not errRateLimited, so this line runs rpt.Failed++ once for each package still in the list (and main.go:694 does the same for each pending DELETE). ghcr-prune.yaml:102 sets FAIL_ON_ERROR: 'true', so main.go:210 exits 1. The first real run is the one that blows through 30 minutes -- hadron alone carries one version per commit across the master and PR matrices, at one DELETE request each -- so the "stop cleanly and let tomorrow finish the tail" behaviour built for rate limits does not apply to the timeout, and the daily job goes red with `context deadline exceeded` until the backlog drains.

Treat the deadline as backpressure, not failure. Add a helper next to errRateLimited, e.g. `func stopRun(err error) bool { return errors.Is(err, errRateLimited) || errors.Is(err, context.DeadlineExceeded) }`, and use it at all three stop sites (main.go:652, 662, 689) with a warning that names the timeout rather than the rate limit. Add a spec that runs pruneAged with an already-cancelled or immediately-expiring context over two packages and asserts rpt.Failed == 0. Consider also making the 30-minute budget a flag so a catch-up run can be dispatched with a longer one.

.github/workflows/ghcr-prune.yaml
```diff
@@ -0,0 +1,107 @@
...
+    name: prune
+    runs-on: ubuntu-latest
+    env:
+      # PACKAGES: one "<ref-prefix>  <retention>" pair per line.
+      # Retention accepts h/d/w (14d and 2w both mean 336h).
+      # Add a new flavor here when the build-iso matrix grows one.
+      # A missing package is a warning, not a failure -- if a flavor
+      # ships a first image after this list is updated, the prune
+      # run right after picks it up.
+      PACKAGES: |
```
itxaka-agent, line 65: Every stop path in pruneAged (main.go:652, 662, 689) returns from the whole function, not from the current package, and the list is walked in the fixed order written at lines 66-75. `hadron` is first and holds the most versions; `bundles-test` and `kairos-uki` are last and are the two packages with the 6h retention, so the ones that most need a daily sweep. If draining hadron exhausts the ~1000 REST requests/hour budgeted at lines 99-104, the run stops before it ever reaches them, and the next day repeats the same prefix. The 6h packages stay unpruned for as many days as hadron takes to drain.

Give every package a share of each run. Cap the DELETEs issued per spec -- a `-max-deletes` flag (say 200) checked in the pruneAged inner loop, moving to the next spec when hit -- so one large package cannot consume the whole budget. Add a spec that seeds two packages with more prunable versions than the cap and asserts a DELETE was attempted against both. If you would rather keep the whole-run abort, rotate the starting index between runs so the head of the list changes.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,1002 @@
...
+//	  Age-based sweep. For each package in $PACKAGES (one
+//	  "ghcr.io/<owner>/<pkg>  <retention>" pair per line), list every
+//	  version, keep any tagged version whose tag matches a glob in
+//	  $PROTECTED_TAGS or whose created_at is younger than the
+//	  retention, and keep any untagged version referenced by a
+//	  kept-tagged manifest list (multi-arch children). Delete the
+//	  rest.
+//
+// Untagged manifests are handled through a two-step referencing pass:
+// the prune walk fetches the manifest of every tagged version it
```
itxaka-agent, line 19: The header says the referencing pass "fetches the manifest of every tagged version it intends to KEEP", and line 15 says untagged versions survive when referenced by a "kept-tagged manifest list". The loop at main.go:771 applies no keep/delete filter -- it collects the digest of every tagged version, aged-out ones included -- and the comment at main.go:761-768 states that directly and explains why (a failed parent DELETE must not leave a live manifest list with missing children). Two comments in one file give opposite specs for the only mechanism standing between a live multi-arch tag and a deleted child, and ghcr-prune.yaml:14 repeats the stale wording.

Rewrite main.go:18-27 to match main.go:761-768: the pass fetches the manifest of every tagged version, kept or aged-out, so children of a parent being deleted this run are deferred one run and age out on the next. Drop "intends to KEEP" and change "kept-tagged manifest list" at main.go:15 and ghcr-prune.yaml:14 to "tagged manifest list".

.github/workflows/ghcr-prune.yaml
```diff
@@ -0,0 +1,107 @@
+name: ghcr prune
+
+# Age-based sweep of GHCR container packages. Replaces the per-run /
+# per-PR-close cleanup jobs that used to sit inside pr.yaml, master.
```
itxaka-agent, line 4: The header says this workflow "Replaces the per-run / per-PR-close cleanup jobs that used to sit inside pr.yaml, master.yaml, _uki-test.yaml, and pr-close-cleanup.yaml". None of those jobs exist on master: there is no pr-close-cleanup.yaml file in the repo, and a case-insensitive grep for "cleanup" across .github/workflows/ on this branch returns only reusable-factory.yaml's disk-cleanup input and the _build-iso.yaml passthrough. This PR's diff touches 8 files and deletes a cleanup job from none of them. The sentence describes an earlier revision of this same PR, so the next reader goes looking for code that was never on master.

State what is true: GHCR has no server-side TTL equivalent to quay's expires-after, nothing has ever deleted the CI tags these workflows push, and this is the single place that now does. Drop the file list and the pr-close-cleanup.yaml reference. Apply the same edit to _uki-test.yaml:64, which says "this workflow itself no longer runs a per-job cleanup step" -- it never ran one.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,1002 @@
...
+		runDelete(args)
+	case "prune":
+		runPrune(args)
+	default:
+		fmt.Fprintf(os.Stderr, "::error::ghcr-cleanup: unknown subcommand %q (want delete|prune)\n", sub)
+		os.Exit(2)
+	}
+}
+
+func runDelete(args []string) {
```
itxaka-agent, line 116: Nothing in the tree runs the delete subcommand. ghcr-prune.yaml:107 is the only invocation of the binary and it passes `prune`; there is no action.yml any more (go.mod:9-11 says the wrapper is gone) and no workflow sets $IMAGES. runDelete, parseRefs, parseImageRef, deleteRefs, deleteMatchingVersions and hasTag land unreachable, together with roughly 250 lines of specs covering them. The previous revision of this PR called delete from pr.yaml; after the restructure it is residue.

Make it a decision rather than leftover. Either drop the delete subcommand and its specs from this PR and add them back when a caller exists, or give it an entry point: a second workflow_dispatch input on ghcr-prune.yaml taking a newline-separated ref list, dispatching `go run . delete` when it is non-empty, so the code is reachable and someone will notice if it breaks.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,1002 @@
...
+		sort.SliceStable(toDelete, func(i, j int) bool { return toDelete[i].age > toDelete[j].age })
+		for _, cand := range toDelete {
+			label := "tagged"
+			if len(cand.tags) == 0 {
+				label = "orphaned untagged"
+			}
+			if dryRun {
+				fmt.Printf("[dry-run] would delete %s (%s, id %d, age %s, tags %v, digest %s)\n",
+					spec.Ref, label, cand.id, cand.age.Truncate(time.Hour), cand.tags, cand.digest)
+				rpt.Deleted++
```
itxaka-agent, line 681: In dry-run the loop prints "[dry-run] would delete ..." and then increments rpt.Deleted, so the summary printed at main.go:209 reads `ghcr-cleanup prune: deleted=417 failed=0 ...` on a run that removed nothing. A maintainer who dispatches with dry_run: true to check the blast radius before letting the pruner loose sees the same summary line a real run prints. Dry-run is the only safety valve on a tool that issues unrecoverable DELETEs against production packages.

Separate the counters or the label. Either add a WouldDelete field to report and print it instead of Deleted when dryRun is set, or make main.go:209 print a "[dry-run, nothing deleted]" marker in the summary. Then extend the existing spec ("issues no DELETEs in dry-run mode but still counts them") to assert whichever shape you pick.

itxaka-agent (changes requested):
> action taken here.

**Round 1 review.** The restructure closes every round-0 finding on its merits: the `pull_request_target` + PR-authored-code + `packages: write` privilege-escalation class is gone (this is now a scheduled/dispatch-only job with no artifact-derived ref list), the pagination-during-delete ordering bug is fixed with a collect-then-delete pass and a page-boundary spec, manifest fetching moved to digest-based lookups, manifest-fetch failures fail closed, and DCO/action pins are clean.

Two blockers remain before this can ship, both in the new scheduled-prune path:

- `main.go:667` -- the 30-minute sweep context expiring mid-run falls into the `Failed` branch instead of the clean-stop branch built for rate limits. With `FAIL_ON_ERROR: 'true'`, the first catch-up run (the one likeliest to exceed 30 minutes) exits red with a timeout rather than stopping gracefully.
- `ghcr-prune.yaml:65` -- packages are walked in a fixed order (`hadron` first, the two 6h-retention packages last). A budget-exhausting run never reaches the packages that most need a daily sweep.

Four smaller findings are inline: two doc/comment mismatches describing an earlier revision of this PR, unreachable `delete`-subcommand code with no caller in the current wiring, and a dry-run summary line that reports deletions it didn't perform.

Full findings below as inline comments.

itxaka-agent:
> action taken here.

**Audit trail -- round 1**

1. **Resumed.** Round 0 (2026-09-15) requested changes against a per-workflow (`pr.yaml`/`master.yaml`) cleanup-job design. The author restructured the whole approach instead of patching that design: force-pushed a squashed 2-commit history dropping the per-PR/per-push wiring entirely in favor of a new scheduled daily prune (`.github/workflows/ghcr-prune.yaml`) driving a substantially larger Go tool (`main.go`, 332 -> 1002 lines; tests 399 -> 874 lines).
2. **Reviewer, round 1.** Verdict: `changes-requested`. Reviewed the new design from scratch rather than re-checking the old (now-deleted) code path. Confirmed the round-0 privilege-escalation class is gone structurally (no more `pull_request_target` + PR-authored code + `packages: write`), the pagination-during-delete bug is fixed with a collect-then-delete pass and a boundary-straddling spec, manifest fetching moved to digest-based lookups, and manifest-fetch failures fail closed. DCO (`Signed-off-by`) present on both commits; the two re-verified `uses:` action pins (`actions/checkout@v7.0.1`, `actions/setup-go@v7.0.0`) both resolve to the SHAs claimed in the diff.
3. **Findings.** Six:
   - `main.go:667` -- the 30-minute sweep context timing out mid-run falls into the `Failed` branch instead of the clean-stop path built for rate limits, so the first catch-up run (likeliest to exceed 30 minutes) goes red with a timeout under `FAIL_ON_ERROR: 'true'` instead of stopping gracefully.
   - `ghcr-prune.yaml:65` -- packages are walked in a fixed order with the largest package (`hadron`) first and the two 6h-retention packages last, so a budget-exhausting run never reaches the packages that most need a daily sweep.
   - `main.go:19` -- the file header and the actual referencing-pass code (main.go:771) describe opposite behavior for which manifests get fetched.
   - `ghcr-prune.yaml:4` -- the header describes replacing cleanup jobs (including a `pr-close-cleanup.yaml`) that do not exist on master; describes an earlier revision of this same PR.
   - `main.go:116` -- the `delete` subcommand and its ~250 lines of specs have no caller anywhere in the current wiring.
   - `main.go:681` -- dry-run mode increments the same `Deleted` counter a real run does, so the summary line does not distinguish a dry run from a real one.
4. **Builder verification.** Not needed this round -- no finding rested on toolchain behavior.
5. **Review posted.** [Round 1 review](https://github.com/kairos-io/kairos/pull/4651#pullrequestreview-5223708299), `CHANGES_REQUESTED`, 6 inline comments on the diff at commit `c732b073`.
6. **Not high-risk under rule 3a** -- CI/workflow-only change, no boot/install/upgrade/reset/kcrypt/immucore code path touched, so no QEMU verification required.

Ticket stays in `awaiting-author`; the next slot re-polls per rule 12b once the author pushes or responds.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,1002 @@
...
+		req.Header.Set("Authorization", "Bearer "+token)
+	}
+	resp, err := g.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
jimmykarily: Fixed. `ghcrManifestClient.referencedDigests` no longer converts a 404 to `(nil, nil)` — it returns an error, so `pruneAged`'s fail-closed path skips untagged pruning for the whole package when a parent manifest is transiently missing. New `ghcrManifestClient.referencedDigests` spec (httptest-backed) covers both the 404-error and manifest-list happy paths.

.github/workflows/ghcr-prune.yaml
```diff
@@ -0,0 +1,107 @@
...
+      # scratch tag can share a package version with :master
+      # (deterministic manifest digest), and deleting either would
+      # take :master with it.
+      PROTECTED_TAGS: |
+        master
+        latest
+        master-scratch
+        v[0-9]*
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
```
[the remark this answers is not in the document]
jimmykarily: Fixed. Checkout now pins to `${{ github.event.repository.default_branch }}` regardless of trigger. Anyone with write access could push their code to master anyway, so the constraint is not a security fence — it just prevents "oops I dispatched off a WIP branch" from running an unreviewed cleanup binary with `packages: write`. Rationale is in the step comment.

.github/actions/ghcr-cleanup/main.go
```diff
@@ -0,0 +1,1002 @@
...
+//   - HTTP 403 with X-RateLimit-Remaining=0 (documented rate-limit
+//     error, distinct from the 403s the API returns for permission
+//     issues -- those keep X-RateLimit-Remaining > 0).
+func checkRateLimit(resp *http.Response) error {
+	if resp.StatusCode == http.StatusTooManyRequests {
+		return errRateLimited
+	}
+	if resp.StatusCode == http.StatusForbidden && resp.Header.Get("X-RateLimit-Remaining") == "0" {
+		return errRateLimited
+	}
```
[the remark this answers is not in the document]
jimmykarily: Fixed. `checkRateLimit` now also returns `errRateLimited` on `403 + Retry-After`, not just on `403 + X-RateLimit-Remaining: 0`. Secondary rate limits can fire before the primary bucket runs out, so the previous check would count each throttled response as a permanent DELETE failure. New spec `treats a secondary rate-limit (403 + Retry-After, remaining>0) the same as a primary 429` locks this in.

.github/actions/ghcr-cleanup/main.go
jimmykarily: Fixed. Introduced an `isStopSignal` helper that groups `errRateLimited`, `context.DeadlineExceeded`, and `context.Canceled` -- all three are external ceilings we can't push through, so the run should stop cleanly rather than count each remaining package as `Failed`. Every stop branch in `pruneAged` now uses it. New spec `does not count the two remaining packages as Failed when the context deadline fires mid-run` asserts `rpt.Failed == 0` on a pre-cancelled context.

.github/workflows/ghcr-prune.yaml
jimmykarily: Fixed. Reordered `PACKAGES` in the workflow so the two 6h-retention entries (`kairos-uki` and `bundles-test`) come first, followed by the 10d entries. A rate-limit or deadline stop still bails out of the whole spec list, but the ephemeral ones are now covered before the walk gets to the giant `hadron` package. Rationale is now in the workflow header comment ("Order matters. ...").

.github/actions/ghcr-cleanup/main.go
jimmykarily: Fixed. Header rewritten to match the code: it now says "fetches the manifest of every tagged version (kept AND aged-to-delete)" instead of "kept". This is the deliberate change that defers children of aged parents to the next run, so the old wording actively misled.

.github/workflows/ghcr-prune.yaml
jimmykarily: Fixed. Header rewritten to describe what the workflow does today rather than what it "replaces" — removed references to the per-run cleanup jobs and to `pr-close-cleanup.yaml`, all of which were reverted earlier in the branch.

.github/actions/ghcr-cleanup/main.go
jimmykarily: Fixed. Deleted `runDelete`, `parseRefs`, `parseImageRef`, `deleteRefs`, `deleteMatchingVersions`, `hasTag`, and their specs. `main` now dispatches straight to `runPrune` (still accepts the `prune` token as a no-op for backward compat with any muscle-memory invocation). Removed the now-unused `NotFound` and `Skipped` counters from `report`. Net: main.go −138 lines, main_test.go −140 lines, 36 specs (down from 42 with dead-code specs, up from 30 after adding the new-behavior coverage).

.github/actions/ghcr-cleanup/main.go
jimmykarily: Fixed. `report` now carries a `DryRun` bool and its `String()` labels the counter `would-delete=` instead of `deleted=` in dry-run mode; `runPrune` also prefixes the summary line with "(dry-run, no DELETEs issued)". A dispatched dry-run now reads `ghcr-cleanup prune (dry-run, no DELETEs issued): would-delete=417 failed=0 unresolved=0`. New spec `report.String labels the counter would-delete in dry-run` locks the formatting.

Itxaka:
Pretty sure there is an action around to clean Up old tags and i think we use It on Hadron to clean Up stuff. Maybe we should check that instead of reinventing the wheel?

jimmykarily:
Checked hadron: it has `pr-cache-cleanup.yml` in-tree but that's a home-grown workflow, not a marketplace action. It targets `quay.io/kairos/hadron-build-cache` (different registry, different auth path), scoped to BuildKit cache tags of shape `pr-<N>-*` on a single cache repo, using skopeo + a Python one-liner + `gh pr list`. No age-based retention, no manifest-list handling. Not directly reusable for the GHCR case here.

Marketplace options that DO exist:

- `actions/delete-package-versions` (GitHub-official): deletes by version count or name pattern. No age-based retention, no manifest-list orphan safety.
- `snok/container-retention-policy`: closest match. Age-based retention, `skip-tags`/`filter-tags` globs, dry-run, and recent versions walk manifest lists. Config is per-invocation so 10 packages = 10 matrix cells or 10 separate steps.
- `dataaxiom/ghcr-cleanup-action`: similar shape, less mature.

Trade-off with what we built: the custom tool covers three things a stock action wouldn't give us for free:
1. Per-package retention driven by one visible YAML array (10 packages × different retentions in one file).
2. The specific multi-arch handling we settled on (fetch every tagged manifest, defer children of aged parents by one run so a failed parent DELETE can't leave a broken manifest list).
3. Rate-limit stops cleanly + `FAIL_ON_ERROR` split so a first-run catch-up doesn't turn the workflow red.

`snok/container-retention-policy` covers #1 via matrix, probably #2 via a multi-arch flag, and #3 would need external retry/backoff work. Switching saves ~700 lines of Go at the cost of the tailored orphan semantics + rate-limit model + local `go test` coverage.

Leaving the note here rather than switching — happy to reopen if you'd rather trade the custom code for the smaller config surface.

— Claude

Itxaka:
wops, i was sure we had a simple actrion from that, maybe it didnt work and had to move into skopeo...sorry about the noise

jimmykarily:
Let's give it a go. It rather separate from anything else so we can easily drop it if it doesn't work as expected. And it only touches test artifacts thus even if it messes up, we can just re-trigger the failed pipelines.

PR #4660, opened 2026-09-16
fix(immucore): bring up the network in the UKI initrd for remote KMS

### LINKED ISSUE #3758: Fix early boot networking in UKI mode
In UKI mode, after installation, there is no proper network setup and the KMS url cannot be resolved (DNS issue) and can't be reached in general (general networking issue, no interface is up etc).

This is the only thing left to have remote KMS working in UKI mode after finishing this story: https://github.com/kairos-io/kairos/issues/3723

ci-robbot:
Closes #3758.

## What was wrong

A UKI initrd is the whole rootfs cpio'd into the `.initrd` section, with `/init`
a symlink to immucore. There is no dracut hook chain in it, so every `rd.*`
network option on the cmdline is inert, no interface is ever configured before
the switch to the real root, and there is no `/etc/resolv.conf`. I measured this
on #3758 when my earlier attempt at the cmdline option did nothing: same time to
first DHCPv4 address with and without it, because the address came from
systemd-networkd in the real root either way.

Remote KMS unlock runs *before* that switch, so it has never worked in UKI mode.
That was structural, not a misconfiguration:

- `steps_uki.go` `UKISetupNetwork` logged `Not implemented yet!` and returned nil
- `dag_uki_boot.go` had the registration commented out, so `OpUkiNetwork` was not
  a node in the graph at all
- `UKIUnlock` was registered without the dependency, under a TODO reading
  "Until then, no KMS for UKI"

## What this does

Runs `systemd-networkd` directly from the initrd. It is the only DHCP client in
the image (I unpacked `ghcr.io/kairos-io/hadron-trusted:v0.5.2`: no `dhclient`,
no `dhcpcd`, busybox built without the applet), and it does not need systemd as
PID 1 - it talks to the kernel over netlink and reports through files under
`/run/systemd/netif`. Then wait for an interface to get a routable address, and
publish the DNS servers it learned.

Details worth a second look:

- **It is off unless this boot has something to reach.** The gate calls
  `kcrypt.ScanKcryptConfig`, the same collector kcrypt itself asks at unlock
  time, so the two cannot disagree about whether a challenger server or mDNS is
  configured. It also skips on a boot that unlocks nothing, using the same
  `SkipUKIUnlock` condition the unlock step uses.
- **It cannot hang or fail a boot.** The wait is bounded at 30s and every error
  is logged, not returned. `UKIUnlock` takes `OpUkiNetwork` as a *weak*
  dependency, so a network that never came up still runs the unlock and the
  operator sees the real KMS error instead of a skipped step.
- **resolv.conf.** Images with systemd-resolved ship `/etc/resolv.conf` as a
  symlink into `/run/systemd/resolve/`, and resolved is not running in the
  initrd, so the *target* is written, not the link.
- **Only `/etc` and `/run` count as "the image already configures this".**
  `/usr/lib/systemd/network` is systemd's own vendor directory and is never
  empty: hadron-trusted v0.5.2 ships 18 drop-ins there, all for containers,
  tunnels and ad-hoc wifi, none matching a plain wired NIC. My first cut
  included it and would have suppressed the DHCP fallback and left the
  interface down. There is a spec pinning that now.

## Verified

- 4 new spec groups, 33 specs across `internal/utils`, `pkg/dag` and `pkg/state`.
  `pkg/dag` had no test file at all before this.
- Mutation-checked layer by layer: 13 mutations introduced one at a time
  (unregistering the step, turning the weak dep strong, dropping the dep,
  dropping the mDNS half of the gate, ignoring the wait deadline, accepting any
  file as networkd config, putting `/usr/lib` back in the list, skipping the
  resolv.conf symlink, allowing a symlink that escapes root, matching any
  `DNS`-prefixed key, dropping address validation, dropping the `#name` suffix
  strip, renumbering the fallback unit). All 13 killed.
- Paths checked against the shipped binary, not from memory: `strings` on
  hadron-trusted's `systemd-networkd` shows `/run/systemd/netif/state`,
  `/run/systemd/network` and `/etc/systemd/network` compiled in, which are the
  three this uses.
- `go vet` and `golangci-lint run ./immucore/...` clean, full `./immucore/...`
  suite green.

## Not verified here, and how to close that

I could not boot this. Building a trusted-boot UKI ISO needs AuroraBoot, which
needs a container runtime this box does not have, and the box also blocks user
namespaces, so I could not run networkd standalone in a netns either. What
remains unproven is the runtime half: that networkd started by immucore-as-PID-1
actually completes a DHCP exchange and writes the state file. The
`test-uki / install-upgrade` job exercises a real UKI boot under QEMU + OVMF +
swtpm, so CI covers the "does this break an ordinary UKI boot" half; the KMS
half needs someone to point a build at a challenger server. Happy to drive that
if someone can tell me which KMS endpoint to use.


mudler (dismissed):
This is a bit tricky, but AFAICT there is no other option than bringing up the networking manually at this point. Not sure if there is anything systemd provides here to do this less painful, but this looks OK, and could be improved along the way

mudler:
@ci-robbot rebase

PR #4666, opened 2026-09-16
fix(agent): make --override-cloud-init-paths actually remove the defaults

### LINKED ISSUE #4665: kairos-agent run-stage: --override-cloud-init-paths does not remove the defaults, so it is identical to --cloud-init-paths
## Summary

`kairos-agent run-stage --override-cloud-init-paths` behaves exactly like
`--cloud-init-paths`. It does not remove the default paths, which is what its
own usage text promises:

`agent/pkg/cmd/root.go:735`

```go
&cli.StringSliceFlag{
	Name:  "override-cloud-init-paths",
	Usage: "Override paths to use when running the stage, removing defaults. Supercedes --cloud-init-paths",
},
```

The flag does replace `config.CloudInitPaths` (`root.go:763-765`), but
`runstage` then prepends the defaults to whatever is in there regardless:

`agent/pkg/utils/runstage.go:44`

```go
cloudInitPaths := append(constants.GetCloudInitPaths(), cfg.CloudInitPaths...)
```

`constants.GetCloudInitPaths()` is `[]string{"/system/oem", "/oem/",
"/usr/local/cloud-config/"}` (`agent/pkg/constants/constants.go:145`), so both
flags end up meaning "defaults plus these".

## Demonstration

`RunStage` driven with `CloudInitPaths: []string{"/only/this/path"}` (what the
override flag sets) against a `CloudInitRunner` that records the arguments it
is handed:

```
RUNNER CALL: []string{"initramfs.before", "/system/oem", "/oem/", "/usr/local/cloud-config/", "/only/this/path"}
RUNNER CALL: []string{"initramfs",        "/system/oem", "/oem/", "/usr/local/cloud-config/", "/only/this/path"}
RUNNER CALL: []string{"initramfs.after",  "/system/oem", "/oem/", "/usr/local/cloud-config/", "/only/this/path"}
```

The defaults are there in every call.

## Impact

Anyone reaching for the flag wants the opposite of what they get: running one
cloud-config in isolation. Instead the machine's whole `/oem` is re-applied for
that stage as well, which for a stage like `initramfs` means re-running
arbitrary user commands. There is currently no way to run a single
cloud-config and nothing else through the CLI.

Found while fixing #4663. That fix does not depend on this one: it drives yip
in process precisely because `run-stage` cannot express "only this config".

## Expected

`--override-cloud-init-paths` runs the stage with only the given paths, and
`--cloud-init-paths` keeps adding to the defaults.

ci-robbot:
Closes #4665.

## What was wrong

`run-stage --override-cloud-init-paths` advertises "Override paths to use when
running the stage, removing defaults. Supercedes --cloud-init-paths", but it did
exactly what `--cloud-init-paths` does. The flag replaced
`config.CloudInitPaths` (`root.go:763-765`), and then `runstage` put the
defaults back in front of whatever it found there:

```go
cloudInitPaths := append(constants.GetCloudInitPaths(), cfg.CloudInitPaths...)
```

`constants.GetCloudInitPaths()` is `/system/oem`, `/oem/`,
`/usr/local/cloud-config/`, so both flags meant "the defaults plus these". There
was no way to run one cloud-config on its own, and for a stage like `initramfs`
that means the machine's whole `/oem` is re-applied alongside it, re-running
arbitrary user commands.

## What this does

Resolves the paths at the `run-stage` call site rather than inside `RunStage`:

- `RunStage` / `RunStageAnalyze` keep composing defaults + config themselves, so
  install, upgrade, reset, the uki flows and the boot hooks (12 call sites) are
  untouched.
- `RunStageWithPaths(cfg, stage, analyze, paths)` runs against exactly `paths`.
  It is the old unexported `runstage` with the list lifted into a parameter, not
  a new code path.
- `runStageCloudInitPaths` in `root.go` is the one place that decides:
  `--cloud-init-paths` appends, `--override-cloud-init-paths` replaces and wins.

`--cloud-init-paths` and a plain `run-stage` behave exactly as before.

## Verified

Six new tests across the two layers, each mutation-checked:

| layer | mutation | killed by |
|---|---|---|
| `pkg/utils` | put the `append(GetCloudInitPaths(), ...)` back inside `RunStageWithPaths` | "runs only the paths it is handed", "analyzes only the paths it is handed" |
| `pkg/utils` | make `RunStage` stop adding the defaults | "keeps prepending the defaults on a plain RunStage", plus an existing cmdline spec |
| `pkg/cmd` | drop the early return so override falls through to the defaults | `TestRunStageOverrideDropsTheDefaults` |
| `pkg/cmd` | typo the flag name to `override-cloud-init-path` | `TestRunStageOverrideDropsTheDefaults` |
| `pkg/cmd` | make `--cloud-init-paths` stop appending | `TestRunStageExtraPathsKeepTheDefaults` |

The `pkg/cmd` tests drive the real `run-stage` command's flag set (they look the
command up in `cmds` and `Apply` its flags), so a renamed flag fails them. The
first mutation reproduces the bug verbatim:

```
override paths = [/system/oem /oem/ /usr/local/cloud-config/ /only/this/path], want [/only/this/path]
```

`go vet ./agent/...` and `golangci-lint run` on both packages are clean.
`./agent/pkg/cmd` is fully green; `./agent/pkg/utils` has 5 pre-existing
failures on this box only (`SyncData`, which shells out to `rsync`), and
`./agent/pkg/action` has 1 (`RenderTemplate`), both confirmed identical on
pristine `master`.

## Scope

Only the paths change. `--override-cloud-init-paths` does not suppress the
`kairos.config_url=`/`cos.setup=` cmdline source or the dot-notation cmdline
stages: those are not cloud-init paths and the flag never claimed them. Say the
word if you want a flag that isolates the run from the cmdline too.

## The one shipped caller: `00_datasource.yaml:39-40`

`kairos-init/pkg/bundled/cloudconfigs/00_datasource.yaml` is the only shipped
user of the flag, and this PR does change what it does. That is intentional:

```yaml
- if: '[ -e /oem/95_userdata ] && [ -f /run/.userdata_load ]'
  name: "Run stages if userdata is found"
  commands:
  - kairos-agent run-stage initramfs --override-cloud-init-paths /oem/95_userdata
  - kairos-agent run-stage boot --override-cloud-init-paths /oem/95_userdata
```

This is the late-datasource path: `/oem/95_userdata` arrived from AWS/GCP/Azure/
Hetzner during the `network` stage, after `initramfs` and `boot` already ran for
this boot. The block exists to apply the userdata that just landed.

Before this PR the effective path list was `/system/oem`, `/oem/`,
`/usr/local/cloud-config/`, `/oem/95_userdata`, so it re-ran *every* default
`initramfs` and `boot` stage a second time in the same boot. That is the
double-application #4665 is about, not a feature. After this PR the two
commands run only the userdata, once.

Nothing is lost by not re-running the defaults. Walking the bundled configs,
only two of the 24 `initramfs`/`boot` stages read or write anything under
`/oem`:

- `26_selinux.yaml` -- the `restorecon -Riv ... /oem ...` is inside
  `/usr/local/bin/selinux-relabel-persistent`, which the yip stage only
  *writes*. It is executed by `kairos-selinux-relabel.service`
  (`After=cos-setup-fs.service`, `Before=multi-user.target`), so it still
  labels the new directory on this boot. Re-running the stage only rewrote the
  script.
- `10_accounting.yaml:43-47` -- `chown -R root:admin /oem` plus
  `chmod 770 /oem`. This one really did catch `/oem/95_userdata` on the second
  run. Second commit handles it, see below.

The other 22 stages are independent of `/oem` content: they set up users,
sudoers, services, grub, network, hosts, profile, ulimits, autologin, RPi and
Jetson quirks, sysexts, and the recovery/reset/installer entrypoints. Each ran
once already this boot with the same inputs.

### Ownership of the new directory

The yip datasource plugin creates `/oem/95_userdata` at `0755` and its files at
`0644`, owner `os.Getuid()`/`os.Getgid()`, i.e. `root:root`
(`yip/pkg/plugins/datasource.go:83-92,253,258`). The second `chown -R` used to
make that subtree `root:admin`.

No access changes either way: `/oem` itself is `770 root:admin`, so only root
and the `admin` group can traverse it, and the plugin's `0755`/`0644` already
grant read to both through the other bits. The `chmod 770` also applied to
`/oem` only, never to the subdirectory, so nothing was being tightened.

The second commit adds `chown -R root:admin /oem/95_userdata` to the block
anyway, so the first boot ends with the same ownership the next boot's
`initramfs` stage gives it. A failing `chown` cannot block the run-stage
commands: yip's `Commands` plugin accumulates errors and `continue`s.


itxaka-agent:
> any action taken here.

Picking this up for review — checking the RunStage/RunStageAnalyze call sites this touches and the new tests.

agent/pkg/cmd/root.go
```diff
@@ -1776,6 +1767,27 @@ func skipAutoInstall(c *cli.Context) bool {
...
+// other caller of RunStage (install, upgrade, reset, the boot hooks) keeps
+// reading the defaults.
+func runStageCloudInitPaths(c *cli.Context, config *sdkConfig.Config) []string {
+	if extra := c.StringSlice("cloud-init-paths"); len(extra) > 0 {
+		config.CloudInitPaths = append(config.CloudInitPaths, extra...)
+	}
+
+	if override := c.StringSlice("override-cloud-init-paths"); len(override) > 0 {
+		config.CloudInitPaths = override
+		return override
```
itxaka-agent, line 1785: `kairos-init/pkg/bundled/cloudconfigs/00_datasource.yaml:39-40` is the only shipped caller of `--override-cloud-init-paths`, and this function is what flips what it does at boot (anchored here since that YAML file isn't part of this diff).

Those two commands run in the `network` stage, guarded on `[ -e /oem/95_userdata ] && [ -f /run/.userdata_load ]` — the late-datasource path, where cloud-config arrives from AWS/GCP/Azure/Hetzner after initramfs has already finished. Before this PR they ran with `/system/oem`, `/oem/`, `/usr/local/cloud-config/` and `/oem/95_userdata`. After it they run with `/oem/95_userdata` alone. The PR body enumerates the 12 Go call sites and concludes the blast radius stops at the CLI; it does not mention this one.

Name `00_datasource.yaml:39-40` in the PR body as an intentional behaviour change and say what the new first boot should look like. Back it with evidence for the late-network-datasource path specifically — a boot with a network datasource showing the userdata still applies and the machine comes up — or a reasoned walk through every default-path boot stage showing each is a no-op on the second run.

agent/pkg/cmd/root.go
```diff
@@ -1776,6 +1767,27 @@ func skipAutoInstall(c *cli.Context) bool {
...
 
+// runStageCloudInitPaths resolves the cloud-init paths a `run-stage` run reads,
+// and keeps config.CloudInitPaths in step with them.
+//
+// --cloud-init-paths adds to the defaults; --override-cloud-init-paths replaces
+// everything, defaults included, which is what its usage text has always
+// promised. The composition lives here rather than in RunStage so that every
+// other caller of RunStage (install, upgrade, reset, the boot hooks) keeps
+// reading the defaults.
+func runStageCloudInitPaths(c *cli.Context, config *sdkConfig.Config) []string {
```
itxaka-agent, line 1778: Consequence of the finding above, in `kairos-init/pkg/bundled/cloudconfigs/10_accounting.yaml:43-47` (also outside this diff, anchored here for the same reason).

`chown -R root:admin /oem` plus `chmod 770 /oem` runs in the initramfs stage. On the late-datasource boot it used to run a second time, after the yip datasource plugin wrote `/oem/95_userdata` during the network stage, so the freshly pulled userdata ended up `root:admin 770` like the rest of `/oem`. With `00_datasource.yaml:39` now restricted to `/oem/95_userdata`, this stage no longer re-runs, and the new directory keeps whatever owner/mode the datasource plugin gave it.

Check what ownership and mode the datasource plugin actually writes `/oem/95_userdata` with. If it isn't `root:admin 770`, the admin group loses read access, and the fix is to add the `chown`/`chmod` to the "Run stages if userdata is found" block in `00_datasource.yaml` right after the two run-stage commands.

itxaka-agent (changes requested):
> any action taken here.

Changes requested: 2 findings. The Go-level claim (12 existing RunStage/RunStageAnalyze call sites untouched) checks out, but the change also flips the behavior of the one shipped YAML caller of `--override-cloud-init-paths`, and that caller isn't mentioned in the PR body or covered by a test.

itxaka-agent:
> any action taken here.

Round 0: changes-requested, 2 findings, posted inline on the diff: https://github.com/kairos-io/kairos/pull/4666#pullrequestreview-5280013072

agent/pkg/cmd/root.go
ci-robbot: Correct, and it is intentional. PR body now has a "The one shipped caller" section naming `00_datasource.yaml:39-40`.

The short version: that block runs in the `network` stage, after `initramfs` and `boot` already ran this boot, and its job is to apply the userdata that just landed. With the defaults in the list it re-ran every default `initramfs` and `boot` stage a second time in the same boot, which is the double-application #4665 is about.

I took the second option you offered, the walk. Of the 24 bundled `initramfs`/`boot` stages only two read or write under `/oem`: `26_selinux.yaml`, where the `restorecon -Riv ... /oem` lives in a script the stage only writes and `kairos-selinux-relabel.service` runs later this boot anyway, and `10_accounting.yaml:43-47`, which is your second finding and is now handled in a second commit. The other 22 are independent of `/oem` content and ran once already with the same inputs.

agent/pkg/cmd/root.go
ci-robbot: Checked the plugin. `EnsureDirectories` creates `/oem/95_userdata` at `0755` and `writeToFile` writes `userdata` and the parsed config at `0644`, owner `os.Getuid()`/`os.Getgid()`, so `root:root` (`yip/pkg/plugins/datasource.go:83-92`, `:253`, `:258`).

One correction on the consequence: no access is lost. `/oem` is `770 root:admin`, so only root and the `admin` group can traverse it at all, and `0755`/`0644` already grant read to both through the other bits. The old `chown -R` changed the group owner from `root` to `admin`, which the other bits made redundant. The `chmod 770` applied to `/oem` itself, never to the subdirectory, so nothing was being tightened either, and the next boot's `initramfs` stage chowns the tree regardless.

Added the `chown` anyway in 6be7544, so the first boot ends with the same ownership as every later one. No `chmod`, since the old path never set one there. A failing `chown` cannot block the two run-stage commands: yip's `Commands` plugin appends the error and `continue`s.

kairos-init/pkg/bundled/cloudconfigs/00_datasource.yaml
```diff
@@ -36,6 +36,11 @@ stages:
     - if: '[ -e /oem/95_userdata ] && [ -f /run/.userdata_load ]'
       name: "Run stages if userdata is found"
       commands:
+      # The datasource plugin creates this directory as root:root; every other
+      # path under /oem is root:admin because 10_accounting chowns the whole
+      # tree in the initramfs stage, which already ran this boot. Match it here
+      # so the first boot ends with the same ownership as every later one.
+      - chown -R root:admin /oem/95_userdata
       - kairos-agent run-stage initramfs --override-cloud-init-paths /oem/95_userdata
```
itxaka-agent, line 44: These two commands run in the network stage of every shipped image, on the boot where the cloud-config arrives from AWS, GCP, Azure, Hetzner and friends after initramfs. Before this PR they ran with /system/oem, /oem/, /usr/local/cloud-config/ and /oem/95_userdata; now they run with /oem/95_userdata alone. This block also gates unattended install and reset: agent/internal/agent/install.go:125 and agent/internal/agent/reset.go:140 both sit in ensureDataSourceReady until this block deletes /run/.userdata_load. The PR body still describes the blast radius as the 12 Go call sites and does not name this file. Round-1 walk of the defaults: 02_agent.yaml's boot stage is covered by kairos-agent.service being After=cos-setup-network.service, 08_ssh.yaml guards its keygen with test -f, 09_systemd_services.yaml/99_sysext.yaml/52_installer.yaml only rewrite static files and re-enable units that already ran this boot, 00_home_dir_owner_fix.yaml sorts after 00_datasource.yaml in the same network stage so a late userdata's users still get fixed, and 31_hosts.yaml:50 is the only template in the directory (keyed on .MachineID). What does not clear: the .after sub-stages -- 10_accounting.yaml:53 and 03-setupcon.yaml used to run after the userdata's own initramfs steps and no longer do. Same class of ordering effect as the round-0 ownership bug, which this PR only fixed after review flagged it.

Two things close this. First, name 00_datasource.yaml:43-45 in the PR body as an intentional behaviour change and say what the new first boot looks like. Second, back it for the late-network-datasource path: boot an image whose cloud-config is pulled at the network stage rather than from cdrom, and show the userdata applied, /oem/95_userdata owned root:admin, and the machine up. AGENTS.md says a boot-path change needs a boot test and unit tests cannot prove a boot works; nothing in the tree mentions 95_userdata or userdata_load outside this YAML and ensureDataSourceReady, and the new chown on line 43 has not been booted either. If a boot run is not practical, answer in writing instead: what did 10_accounting.yaml's initramfs.after and 03-setupcon.yaml's initramfs.after do on the second pass that is now lost, and why is losing it safe.

itxaka-agent (changes requested):
> action taken here.

Round 1: changes-requested, 1 finding. The `/oem/95_userdata` ownership fix from round 0 is confirmed resolved.

itxaka-agent:
> action taken here.

Round 1 review: changes-requested. https://github.com/kairos-io/kairos/pull/4666#pullrequestreview-5287549214

kairos-init/pkg/bundled/cloudconfigs/00_datasource.yaml
ci-robbot: Answering in writing, per your offer. Evidence is yip v1.26.3's own `Graph()` over the shipped `kairos-init/pkg/bundled/cloudconfigs` directory, i.e. the same DAG `run-stage` executes.

`initramfs.after` under the defaults is exactly two blocks, and `boot.after` is empty:

```
### stage=initramfs.after paths=[<oem> <oem>/95_userdata]
  [<oem>] Fonts fix for ubuntu QR code display.setup console fonts
  [<oem>] Default user, permissions and serial login.Enable serial login for alpine
### stage=boot.after paths=[<oem> <oem>/95_userdata]
  (empty)
```

Both take static inputs and read nothing under `/oem`, and both already ran once this boot in the real `initramfs` stage. `10_accounting`'s is openrc-only: `sed -i 's/ttyS0.*//g' /etc/inittab` plus the getty append. `03-setupcon`'s is Ubuntu/Debian-only: a fixed `/etc/default/console-setup` plus `setupcon`. So on the second pass they were idempotent rewrites, except in one case: a userdata that itself writes `/etc/inittab` or `/etc/default/console-setup` in its own `initramfs` stage. There the default `.after` ran on top and silently reverted the user's file. That is the default overriding the userdata, which is the double-application #4665 is about, so losing it is the fix rather than a casualty.

Two more things the probe turned up, both arguing the same way. yip walks a directory recursively and `Run()` builds one DAG per source, so `/oem` already carried `/oem/95_userdata`: before this PR the userdata's own stages ran **twice** per sub-stage inside a single `run-stage` call. And `initramfs.before` re-ran too, including `00_datasource`'s own "Pull data from provider" (a no-op only because of its `[ ! -e /oem/95_userdata ]` guard).

PR body updated to name `00_datasource.yaml:36-47` as an intentional behaviour change, with the new first boot spelled out.

itxaka-agent (approved):
> action taken here.

Approving round 2. The written walk-through of the boot-stage `initramfs.after`/`boot.after` claims checks out against the shipped YAML and Go (verified independently, not taken on trust), and the PR body now names `00_datasource.yaml:36-47` as the intentional behaviour change round 0 asked for.

One minor accuracy note, not a blocker: the PR body's `After=` line for the selinux unit drops `kairos-agent.service` compared to `26_selinux.yaml:32`. Doesn't change the conclusion.

itxaka-agent:
> action taken here.

Audit summary, third-party review flow (this PR was not authored by the Second Foundation, so no coder/tester/docs loop applies here):

- Round 0: changes-requested, 2 findings (missing `Signed-off-by`-adjacent blast-radius documentation for `00_datasource.yaml`, and a follow-on ownership gap on `/oem/95_userdata`).
- Round 1: author fixed the ownership gap in a new commit; changes-requested, 1 finding remaining (the network-stage blast radius on `00_datasource.yaml` still needed to be named in the PR body and backed by either a boot test or a written walk of every default stage).
- Round 2: no new commit. Author answered in writing and updated the PR body. Verified independently against the shipped YAML and Go rather than taken on trust. **Approve.**

Review: https://github.com/kairos-io/kairos/pull/4666#pullrequestreview-5289950826

PR #4669, opened 2026-09-16
feat(immucore): bind /var/log/audit from the persistent partition

itxaka-agent:
> any action taken here.

Refs: #4629

Supersedes #4652, which was rejected: that PR only created `/var/log/audit` with the right permissions via a `kairos-init` step and inferred persistence from `/var/log` already being in `PERSISTENT_STATE_PATHS`. This PR does the thing the issue actually asks for.

## What this does

Adds an explicit, dedicated `/var/log/audit` bind mount to immucore's own mount table, independent of `PERSISTENT_STATE_PATHS` and any cloud-config:

- `immucore/pkg/state/steps_shared.go`: `MountAuditLogDagStep` binds `/var/log/audit` from the persistent partition on every boot flavor that has one (active, passive, in-RAM), pinned to `0700 root:root` with the SELinux label carried onto the backing directory. Registered in `dag_normal_boot.go`, `dag_in_ram.go`, `dag_uki_boot.go`.
- `AuditdMountRequirementDagStep` writes a systemd drop-in (`/run/systemd/system/auditd.service.d/10-kairos-mount-requirement.conf`) making `auditd.service` require the generated `var-log-audit.mount` unit by name. No `auditd.service` ships from this repo yet (it comes from the distro package on RedHat-family images), so this is the mechanism available to guarantee the mount is present before it starts: on a boot where the bind failed, auditd fails closed instead of silently logging to the ephemeral directory underneath.
- A machine that already has `/var/log/audit` content under the old, generic `/var/log` bind (an existing install, or a machine that ran an older version of this code) gets that content migrated into the new dedicated bind once, on the first boot with this feature, and the migration is recorded with a marker file so later boots don't need to inspect the directory to know whether it's done.
- `kairos-agent reset` (both GRUB and UKI/trusted-boot paths) now stages `/var/log/audit`'s contents on `/run` before formatting the persistent partition and restores them after, bounded by available memory — skipped with a warning, never failing the reset, if the trail doesn't fit in half of free space at reset time.

## What this does not do

**Open question for the maintainer:** #4629 asks for the audit log mount on "every boot including recovery mode." This PR does not deliver the recovery half, and that is deliberate rather than an oversight. Recovery and autoreset come up with no persistent volume at all (the recovery/autoreset branch of `kairos-init`'s `00_rootfs.yaml` sets `OVERLAY` and no `VOLUMES`), so covering them means mounting `COS_PERSISTENT` in recovery. That collides with the reset recovery mode exists to run: a bind mount of a subdirectory of the persistent filesystem keeps that filesystem mounted after the volume itself has been unmounted (verified: the device is still listed in `/proc/mounts` under the bind's mount point after unmounting the parent), so the `mkfs` inside `kairos-agent reset` would either refuse to run or run on a live filesystem. Autoreset, whose only job is to format that partition, shares the same layout branch. Changing the recovery layout for every install is a bigger call than this ticket should make on its own, so the gate is documented in `auditLogPersistExpected`, the skip is logged at info level on a recovery boot instead of silently, and the decision is left here for you. If you want recovery covered, say so and it becomes a follow-up that mounts `COS_PERSISTENT` in recovery and teaches both reset paths to tear the bind down before the format.

Does not touch SELinux policy definitions (no policy ships for this path on this base — the label is carried, not authored). Does not add a config knob to disable any of this; the issue asks for it unconditionally.

## Verification (rule 3a, high-risk: immucore mount table + boot ordering + reset)

Full QEMU verification across three rounds — see the `QEMU verification` comment on this PR for build/boot detail and screenshots. Summary: golden-path install confirms the dedicated bind via matching device/inode; the systemd drop-in is confirmed live (masking the mount unit makes `auditd.service` fail to start; unmasking fixes it); an A/B upgrade round-trip preserves the mount and content; a real `kairos-agent reset` preserves the audit trail with the stated memory bound; a real migration scenario (old shared-bind content promoted to the new dedicated bind) is confirmed, including that content written directly to the old backing path after migration is *not* resurrected on a later boot; the recovery-mode skip now logs visibly and does not block boot.

Full unit-test suite green across `immucore/...`, `agent/pkg/action/...`, `agent/pkg/uki/...`; `go vet`, `gofmt -l`, `golangci-lint` all clean.

## Related

- Split off from #112 (CIS).
- Unblocks #4629's downstream consumer #4626 / #4650 (merged) and kairos-io/hadron#4637 (in flight) for the parts of the audit-log requirement that don't depend on recovery-mode coverage.

## Audit trail

Full chronological summary of what each role did across all three review rounds, including the reasons the prior PR (#4652) was rejected and reworked: link posted as a comment on this issue shortly after this PR opens.


itxaka-agent:
> any action taken here.

## QEMU verification

High-risk change (RULES.md rule 3a): immucore's mount table, boot ordering, and `kairos-agent reset`. Verified across three review rounds with real ISO builds, real disk installs, and real reboots — not code reading alone.

### Round 0 — initial implementation

**Golden path.** Fresh install, `/var/log/audit` `0700 root:root`, confirmed bound to the new *dedicated* `var-log-audit.bind` backing path (not the old shared `var-log.bind`) via matching device+inode:

golden path, dedicated bind inode match
(blob sha `60bdf49dcb877ef3c2fbf6beb550de6512ec9291`)

grub menu, installed disk
(blob sha `fe95b4b2dc05130291115f417ac6b943afbe4c9d`)

installed login prompt
(blob sha `ea0e62b2ab6f086b59aeef31716e29efb9b3d4d4`)

**Systemd drop-in gates auditd on the mount.** `/run/systemd/system/auditd.service.d/10-kairos-mount-requirement.conf` confirmed present with the expected `RequiresMountsFor=`/`Requires=`/`After=` lines; masking `var-log-audit.mount` and starting `auditd.service` fails closed ("Unit var-log-audit.mount is masked"), unmasking makes it start:

auditd mount-requirement drop-in
(blob sha `7ad6a89d5f1102c2a0b4bb151b859b414ee53324`)

Requires= gates startup when masked
(blob sha `36da678d978d4201ee67150210aa039f73b5e133`)

**A/B upgrade round-trip.** Genuine active/passive slot swap (distinct sizes/md5s), marker file + mode + dedicated bind all survive:

grub menu after A/B upgrade
(blob sha `3ddf0db6aa45ea5a5850a92bf6fd1d58f62c50a2`)

marker, mode, bind survive the upgrade
(blob sha `ab3788e635427b3581a47774fbb14a95f900bc8f`)

**Migration from the old shared bind.** Installed pre-#4629 code, wrote content under the old shared `/var/log` bind, upgraded to this branch, confirmed the content migrated into the new dedicated bind with matching checksum:

old-bind content migrated to new bind
(blob sha `3b4a394fde9a5b2b45bcd1637c7686953fe6d5a3`)

**Recovery-mode boot.** Not blocked; at this stage in the work neither the mount nor the drop-in existed there (see the round-1 disclosure below for why that gap is now a documented, visible decision rather than silent):

recovery boot, not blocked
(blob sha `70b4eea4fea5bec60ecac90f5fbe31e080bb6752`)

recovery: no mount, no drop-in, by design
(blob sha `a8ceeeeb92da27556c6cbd101bab0d3df9bb90b9`)

**Reset preservation.** Real `kairos-agent reset --unattended`: marker, content, and `0700 root:root` mode all survive:

reset preserves trail, mode, content
(blob sha `ba5ee3028b9b4e3d8c81cc81d4f2034bd72751bc`)

### Round 1 — reviewer fixes re-verified

Fixes: recovery skip now logged visibly at info level; reset staging moved from `/tmp` (not guaranteed writable in recovery) to `/run` (a tmpfs on every boot); UKI reset path now unmounts before formatting; migration sync no longer resurrects content deleted from the new bind while it still exists in the old one.

recovery: visible info-level skip log
(blob sha `cfcf9e4de2492ee8a0f168ce8466befeba505383`)

reset succeeds staging on /run, content+mode restored
(blob sha `fa59e2c29de661e894cb9637890fe295d7daf9ec`)

leftover content in old bind not resurrected after migration
(blob sha `a29892d6b2655dfd9d82e6179615e6b14894d71b`)

UKI/trusted-boot reset path: a real SecureBoot+TPM-signed image got through firmware signature verification and into kernel execution, but a driveable console could not be reached within the round's time budget — the unmount-before-format fix on this path is otherwise covered by a unit test that exercises the exact scenario:

UKI image passes stock-key SecureBoot verification
(blob sha `ce3eaa5377533016e7124091b5900d034fc85dba`)

### Round 2 — migration-marker fix re-verified

The round-1 migration-skip logic used "backing directory non-empty" as a proxy for "migration finished," which could permanently under-migrate a trail if a sync died partway (e.g. `ENOSPC`, since migration roughly doubles the trail's on-disk footprint on the same filesystem). Round 2 replaces that with a marker file written only after a successful sync. Re-verified live: a real install of the pre-#4629 code, content written under the old bind, a real upgrade, and confirmation that the marker file exists as a *sibling* of the backing directory (not inside it) and genuinely gates the sync on the next boot (`journalctl` shows "Skipping the state sync: already migrated"):

post-migration login
(blob sha `862eabe93310fd096f1106a0e8877302acbaee2b`)

marker file as sibling, mode confirmed on tty1
(blob sha `5753a725582d0e73624101f995aa04a8f932c634`)

The specific failure mode this fix addresses (a sync interrupted partway through) was not reproduced live under QEMU — reliably forcing `rsync` into `ENOSPC` mid-copy was judged impractical within this round's budget — and instead relies on a unit test that constructs the identical scenario deterministically (a populated backing directory with no marker, asserting the missing file still gets copied on the next run). The reviewer examined that test and agreed it was an adequate substitute for this specific, fully-deterministic sub-case.

**Not covered, on purpose:** recovery-mode audit-log persistence (see the PR description's "What this does not do" section for the reasoning and the open question for a maintainer decision).

Itxaka (dismissed):

Itxaka:
Im not sure if this is ok. Like why is that bind mount special in immucore? it should be like a normal bind mount and go into the normal list instead. Then we can drop most of the code and only keep the code that saves the dazta in there during reset. @itxaka-agent deal with this please. Also any testing that you do needs to be posted here instead of on the parrent ticket, wtf? If you tesxt this, post the proof here instead

itxaka-agent:
> any action taken here.

Understood on both points, working on it now:

1. Dropping the special-cased `OpMountAuditLog` / `AuditdMountRequirementDagStep` boot-DAG machinery and putting `/var/log/audit` into the normal persistent bind-mount list (`s.BindMounts` in `LoadEnvLayoutDagStep`) so it goes through the existing generic `OpMountBind` path instead of a bespoke step. Keeping only the reset-time stash/restore code.
2. QEMU verification for this round will be posted here on this PR, not on #4629.

Will push an update to this branch once it's re-verified.

itxaka-agent:
> any action taken here.

## QEMU verification — round 3 (refactor per your feedback above)

High-risk change (RULES.md rule 3a: immucore mount table + `kairos-agent reset`). Re-verified under real QEMU/KVM after folding `/var/log/audit` into the generic persistent bind-mount list and dropping the special-cased DAG step + systemd drop-in.

**First pass (commit `09fb45cb`) caught a real regression before it went anywhere:** a fresh install left `/var/log/audit` at `755 root:root` instead of the required `700 root:root`, because the generic mode-inheritance helper read the mountpoint's mode *after* another call had already defaulted it to 0755.

golden path, wrong mode caught
(blob sha `f085cd9ebf21931ee8b89740c26ec1a5acbf508d`)

Fixed in `62619368` (declared per-path mode, applied before the mountpoint can be created with a default). Re-verified clean:

golden path, mode fixed to 0700
(blob sha `8ced073f20e5c30bfb7516e4744d2c541b9e4cd7`)

Fresh install now confirmed `0700 root:root`, device:inode matching the generic bind's backing directory under `.state/` — i.e. mounted by the ordinary `MountCustomBindsDagStep` path, not a bespoke step.

**A/B upgrade round-trip** — content and mode both survive:

A/B upgrade, content+mode survive
(blob sha `05ae6275c57bc3dc0570a3fbf066d39f79674233`)

**`kairos-agent reset --unattended`** — the stash/restore mechanism is unchanged from earlier rounds and still preserves the trail and mode:

reset preserves trail, mode, content
(blob sha `6529837357d3e00ae56d8764aa031f19c8a36c07`)

**Recovery-mode boot** — not blocked. Recovery has no persistent volume in either the old or new design, so there is nothing audit-related to gate there anymore (the old special-case had its own recovery-skip log line; the generic path has no recovery-specific behavior to speak of):

recovery boot, not blocked
(blob sha `0c745f6bb77de60546edcd42cb998e248685f630`)

**Unit tests / lint** — `go build`, `go vet`, `gofmt -l`, and the full relevant test suites (`immucore/...`, `agent/pkg/action/...`, `agent/pkg/uki/...`) all green across every commit in this round, including the two follow-up commits (`77277c8f`, `23bfcaad`) that only touch comments and delete one now-meaningless test — those did not need a fresh QEMU pass since they change no runtime behavior.

**Two trade-offs from dropping the special-case boot-DAG step, now documented in the code itself (`immucore/pkg/state/steps_shared.go` and `immucore/pkg/op/fs.go`) and in the `23bfcaad` commit body, flagging them here too since they affect what the issue originally asked for:**

1. The mount still lands before any service in the booted system starts (it happens in the initramfs, pre `switch_root`), but there is no more `Requires=`/`After=` systemd ordering guarantee against it. If the bind mount itself ever fails, there's no unit-level dependency to make anything fail closed — immucore just logs the bind error and the system boots with `/var/log/audit` on ephemeral storage.
2. Dropping the one-time migration marker means the bind's rsync runs on every boot again, with no `--delete`. On a machine that upgraded from a pre-#4629 image, its old shared `/var/log` backing (which sorts ahead of the new dedicated entry) is never emptied, so in principle a file rotated away from the dedicated audit path could reappear from that old backing on a later boot. Fresh installs never hit this.

Both are accepted consequences of "just make it a normal bind mount" — flagging them explicitly per your ask rather than leaving them implicit.

itxaka-agent:
> any action taken here.

## Audit trail — this round of PR #4669

Per your comment above, testing and audit posts for this work now live here on the PR rather than on the parent issue (#4629).

Context: this PR was previously approved at head `ef26aa19`. You followed the approval with a request to stop special-casing `/var/log/audit` in the boot DAG and make it a normal entry of immucore's generic persistent bind-mount list, keeping only the reset-time stash/restore code. This is the record of that rework.

### Round 3 — refactor per your feedback

- **Coder** (`90f4b4b7`, `09fb45cb`): folded `/var/log/audit` into `s.BindMounts` in `LoadEnvLayoutDagStep` so it mounts through the existing generic `MountCustomBindsDagStep`/`OpMountBind` path. Removed `MountAuditLogDagStep`, `AuditdMountRequirementDagStep`, the systemd drop-in writer, the recovery-mode skip logging, and the migration-marker mechanism, along with their wiring in all three boot graphs and their now-dead tests. Kept the reset stash/restore code untouched. Added one small generic helper (`CreateBindStateDir`) so the `0700 root:root` requirement is still met.
- **Tester**: full QEMU re-verification (golden path, A/B upgrade, reset, recovery) found a real regression — fresh installs landed `/var/log/audit` at `0755` instead of `0700`, because a generic mode-inheritance helper read the mountpoint's mode after another call had already defaulted it. A/B upgrade, reset, and recovery all passed independently of this bug.
- **Coder fixup** (`62619368`): fixed the mode-ordering bug with a declared per-path mode map (`bindMountModes`) applied before the mountpoint can be created with a default, plus a unit test that fails without the fix (confirmed by reverting and re-testing). No DAG step, drop-in, or `MountBindWithMode` resurrected.
- **Tester re-check**: focused golden-path re-verification confirmed `/var/log/audit` now lands `0700 root:root` on a fresh install via the generic bind mechanism. See the separate "QEMU verification" comment above for screenshots and blob shas.
- **Reviewer** (round 3): changes-requested. Confirmed the special-casing was genuinely removed with no dangling references and the mode fix was sound, but flagged two blockers: a leftover test (`agent/pkg/action/reset_audit_test.go`) still asserting a `.migrated` marker mechanism that no longer exists, and two accepted trade-offs (loss of the auditd `Requires=`/`After=` ordering guarantee; every-boot rsync with no `--delete` reopening a resurrection risk for upgrading machines) that were only recorded in internal notes, not in the source.

### Round 4 — addressing review feedback

- **Coder** (`77277c8f`, `23bfcaad`): deleted the dead `.migrated`-marker test; documented both trade-offs at their call sites in `immucore/pkg/state/steps_shared.go` and `immucore/pkg/op/fs.go`, and named both in the `23bfcaad` commit body; corrected a misleading comment about `os.MkdirAll`'s parent-mode behavior in `immucore/internal/utils/common.go`. No functional/runtime change, so no new QEMU pass was needed — `go build`/`vet`/`test` confirmed clean.
- **Reviewer** (round 4): **approve.** All four round-3 findings verified resolved, scope held, DCO clean on every commit (`Signed-off-by: itxaka-agent`, no `Co-authored-by`).

### Net result

10 commits total on this branch since it split from master, 5 of them this round (`90f4b4b7`, `09fb45cb`, `62619368`, `77277c8f`, `23bfcaad`). Net diff against `master` is now 15 files / 906 insertions, down from the round-0–2 diff's 22 files / 1619 insertions, consistent with "drop most of the code."

Branch `triage/4629-immucore-audit-mount-v2` force-pushed to `23bfcaadc4821fe74730c33bb1f632a30c8fe51c`. Two trade-offs from this refactor are called out explicitly in the QEMU verification comment above and in the code itself — please weigh in if either is not acceptable and this should go another round.

Itxaka (approved):

mudler (approved):

PR #4671, opened 2026-09-16
fix(p2p): accept the documented empty role, and stop erroring on "none"

### LINKED ISSUE #4670: p2p.role: the schema rejects the documented empty value and defaults to "none", which the provider hard-errors on
## Summary

`p2p.role` is the one key where the published schema and the runtime disagree in
both directions: the schema rejects the value the docs tell users to write, and
accepts (as its own advertised default) a value the runtime hard-errors on.

`sdk/schema/p2p_schema.go:10`

```go
Role string `json:"role,omitempty" default:"none" enum:"[\"master\",\"worker\",\"none\"]"`
```

`provider/internal/role/p2p/k8s.go:83`

```go
if c.P2P.Role != "" {
    if c.P2P.Role != RoleMaster && c.P2P.Role != RoleWorker {
        return nil, errors.New("invalid p2p.role specified, must be 'master' or 'worker'")
    }
```

So the accepted sets do not overlap cleanly:

| `p2p.role` | `kairos-agent validate` | runtime |
| --- | --- | --- |
| `""` (what the docs use) | **rejected** | ok, role taken from the p2p ledger |
| `none` (the schema default) | ok | **hard error** |
| `master` | ok | ok |
| `worker` | ok | ok |

## Reproducing

Schema half, against the real `RootSchema`:

```
role: ""      -> INVALID: '/p2p/role' ... value must be one of "master", "worker", "none"
role: "none"  -> VALID
role: "master"-> VALID
```

`docs/installation/p2p.md:113` documents `role: "" # Set an hardcoded role, optional`,
so the documented way to leave the role unset fails validation. Validation is
only a warning today (`agent/pkg/config/scan.go`), but `--strict-validation`
turns it into a hard failure, and `kairos-agent print-schema` is what editors
consume, so `""` gets flagged in the editor too.

Runtime half, driving `NewK8sNodeWithDetector` with a k3s binary present and
`p2p: {network_token, auto.enable: true, role: <x>}`:

```
role ""       -> ok, node role="server"
role "none"   -> ERROR: invalid p2p.role specified, must be 'master' or 'worker'
role "auto"   -> ERROR: invalid p2p.role specified, must be 'master' or 'worker'
role "master" -> ok, node role="server"
role "worker" -> ok, node role="agent"
```

## Impact of `role: none`

It is not only the `oneTimeBootstrap` path. Both p2p role handlers call
`NewK8sNode(pconfig)` too (`provider/internal/role/p2p/master.go:102`,
`worker.go:38`), so a node that the ledger later assigns `master` still dies on
the same check with `stopping Master: invalid p2p.role specified`.

And before that, `master.go:92` publishes the raw value into the ledger as this
node's role. `getRoles` then treats `none` as *assigned*, so the node never
appears in `unassignedNodes` while `existsMaster` stays false, and the leader
keeps rescheduling around a node that will never come up.

A user who took the schema's default at face value therefore gets a cluster that
never bootstraps, with an error that says the value must be `master` or
`worker`, from a key the schema said defaults to `none`.

## Suggested fix

- `default:""` and `enum:["","master","worker","none"]` on `P2PSchema.Role`, so
  the documented empty value validates. Keep `none` accepted so configs written
  against the old schema do not start failing validation, and say in the
  description that it means "no hardcoded role".
- Treat `none` as unset in the provider rather than erroring, since it was the
  schema's own advertised default and nothing ever implemented it.

Same drift class as #4667.

ci-robbot:
## Summary

Fixes #4670. `p2p.role` drifted from the runtime in both directions.

`sdk/schema/p2p_schema.go:10` declared `enum:["master","worker","none"]` with
`default:"none"`, while `provider/internal/role/p2p/k8s.go:83` accepts only `""`,
`master` and `worker`, and errors on anything else.

| `p2p.role` | before, validation | before, runtime |
| --- | --- | --- |
| `""` (what `installation/p2p.md:113` documents) | **rejected** | ok, role from the ledger |
| `none` (the schema default) | ok | **`invalid p2p.role specified, must be 'master' or 'worker'`** |
| `master` / `worker` | ok | ok |

## Why `none` is worse than a bad error message

It is not only `oneTimeBootstrap`. `Master()` and `Worker()` both come back
through `NewK8sNode` (`master.go:102`, `worker.go:38`), so a node the ledger
later assigns `master` dies on the same check with `stopping Master: ...`.

And `master.go:92` published the raw value into the ledger as this node's role,
so `getRoles` counted the node as *assigned* while `existsMaster` stayed false,
and the leader kept rescheduling around a node that would never come up.
`schedule.go:69` also excluded the node from master selection as if it had a
static role.

## Changes

- Declare `""` in the enum and default to it, so the documented empty value
  validates. `none` stays accepted rather than becoming a validation error, since
  it was the schema's own advertised default and is therefore in the wild.
- Add `P2P.HardcodedRole()`, which maps `none` to `""`, and read `p2p.role`
  through it in all five places (`bootstrap.go`, `k8s.go`, `master.go`,
  `worker.go`, `schedule.go`). A config that took the default now behaves exactly
  like one that omits the key.

## Tests

`sdk/schema`: a table asserting `""`, `master`, `worker` and `none` all validate,
plus `role: leader` still failing. `provider/internal/provider/config`:
`HardcodedRole` mapping. `provider/internal/role/p2p`: `NewK8sNodeWithDetector`
builds an unpinned node for both `""` and `none`, and still rejects an
unimplemented role.

Mutation-checked the three layers separately: reverting the enum fails the
schema table, breaking `HardcodedRole` fails the config and provider tests, and
reading `c.P2P.Role` raw again in `k8s.go` fails the `none` entry.

`go test ./sdk/schema/... ./provider/...` and `go vet` are green locally.


itxaka-agent:
> action taken here.

Picking this up for review: reading the diff against `provider/internal/role/p2p` and `sdk/schema/p2p_schema.go`, and checking the fix against issue #4670.

sdk/schema/p2p_schema_test.go
```diff
@@ -197,3 +197,66 @@ vpn:
...
+var _ = Describe("P2P Schema role", func() {
+	var config *KConfig
+	var err error
+	var yaml string
+
+	// The provider is the only consumer of p2p.role
+	// (provider/internal/role/p2p/k8s.go), and it accepts the empty string,
+	// "master" and "worker". Anything else stops the bootstrap, so what the
+	// schema declares has to line up with that set.
+	JustBeforeEach(func() {
```
itxaka-agent, line 210: The container-level `JustBeforeEach` parses `yaml` before each spec, but the four `DescribeTable` entries set `yaml` inside their own body and build a second config there, so this `JustBeforeEach` parses the empty string on the first entry and the previous entry's YAML on the ones after it.

It does not fail today -- `yaml.Unmarshal` of an empty string returns `nil, nil` -- but the parse is dead work and the shadow `var e error` in the table body exists only to dodge the outer `err`. Drop the `JustBeforeEach` from this `Describe` and move it into the `Context("with a role no handler implements")` block, which is the only spec that relies on the BeforeEach-sets-yaml idiom. Then the table body can keep using `config, err = NewConfigFromYAML(...)` directly and the shadow variable goes away.

itxaka-agent (changes requested):
> action taken here.

Changes requested. The blocking gap is not in this diff: `provider/internal/role/common.go:28` (`getRoles`) treats any non-empty stored role as assigned, and `provider/internal/role/schedule.go:131-138`'s re-publish loop skips only `""`, `master` and `master/ha`. A node whose ledger entry is already `role=none` from before this fix stays wedged after upgrade -- `master.go:96` wrote that value into the ledger before `NewK8sNode` rejected it, so every cluster #4670 describes is in this state today. `getRoles` needs to treat a stored `none` (or empty) the same way the config side now does, so the leader re-assigns those nodes instead of rewriting `role=none` every tick. If the fix is meant to stay scoped to config validation, say so in the PR body and name the manual recovery command for already-stuck nodes.

One more, inline on the diff: a minor test-setup ordering issue in `p2p_schema_test.go`.

sdk/schema/p2p_schema_test.go
ci-robbot: Done in 9dd2023c, restructured exactly as you described: the `JustBeforeEach` now lives in the `"with a role no handler implements"` context, and the table body assigns the outer `config, err` directly. Mutation-checked by dropping `"none"` from the enum, which fails the table, so the entries are not vacuous.

itxaka-agent (changes requested):
> action taken here.

Round 1: the `JustBeforeEach` test-ordering fix resolves last round's second finding.

One finding remains, in `provider/internal/role/common.go` (not touched by this diff, so it can't be anchored inline): nodes stop writing `none` to the ledger now, but a cluster that already hit #4670 does not recover on upgrade. The old binary left `role/<UUID> = none` in the ledger; `getRoles` (common.go:28) counts any non-empty value as assigned, so that UUID never enters `unassignedNodes`, and the re-publish loop in `schedule.go:131-138` rewrites `none` every tick instead of letting it age out. The `dynamic_roles` dead-node sweep doesn't catch it either, since the node is reachable. After upgrading, an already-stuck node stays stuck with no log line explaining why.

Fix: in `getRoles`, treat a stored role of `""` or `config.RoleNone` as unassigned, the same way the config side now does (needs a `providerConfig` import in common.go, which has none today). Once the UUID lands in `unassignedNodes`, the existing worker-assignment loop republishes a real role over it. Alternatively, state in the PR body that this fix is scoped to config validation only and name the manual recovery step (deleting the stale `role/<UUID>` ledger entry) for already-stuck clusters.

provider/internal/role/schedule_internal_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// A node running a binary from before the fix for #4670 wrote the
+	// configured `p2p.role: none` straight into the ledger, and only then did
+	// NewK8sNode refuse it. Every cluster that hit that bug therefore has
+	// `role/<UUID> = none` entries which were never an assignment. Until the
+	// leader reads them as unassigned, upgrading the binary is not enough:
+	// nothing ever writes a real role over them and the node stays wedged.
+	It("assigns a real role over a stale none left by an older binary", func() {
+		Expect(client.Set("role", "stuck-uuid", providerConfig.RoleNone)).To(Succeed())
+		Expect(roleOf(client, "stuck-uuid")).To(Equal(providerConfig.RoleNone))
```
itxaka-agent: This reads the ledger right after writing it, but the edgevpn API only announces writes: the PUT handler calls `ledger.Persist(...)` and returns `{"State":"Announcing"}` (`api/api.go:457-463`), and `Persist` commits from a goroutine on its ticker's first tick (`pkg/blockchain/ledger.go:424-437`). Nothing orders the write before the read, so `roleOf` can return `""` here. The same pattern at lines 94 and 104 has no read-back guard at all: if the `master-uuid -> master` seed has not committed when `scheduleRoles` runs at line 97/107, `getRoles` returns `""` for it, `existsMaster` stays false, both UUIDs enter `unassignedNodes`, and `schedule.go:86` picks the master with `r.Intn(2)` — so lines 99 and 110 fail half the time. At line 122 it is worse than a red: a read that returns `""` satisfies `NotTo(Equal(none))`, so that spec can pass without observing anything.

Confirm every ledger write before depending on it. Wrap the seed reads (and the post-`scheduleRoles` assertions) in `Eventually` the way upstream edgevpn's own test for this endpoint does (`api/api_test.go:63-75`) — keep the timeout at 2s so the assertion window stays inside the safe range before the seed's own re-announce goroutine (first retry ~2.5s out, `pkg/utils/ticker.go:41-42`) would overwrite it again.

```suggestion
		Eventually(func() string { return roleOf(client, "stuck-uuid") }, 2*time.Second, 20*time.Millisecond).Should(Equal(providerConfig.RoleNone))
```

itxaka-agent (changes requested):
> action taken here.

Round 2: changes-requested, 1 finding. The round-0 and round-1 findings are resolved by the latest commit.

provider/internal/role/schedule_internal_test.go
ci-robbot: Confirmed and fixed in `e6118873`. You are right that nothing orders the write: `Persist` adds from the `Announce` goroutine and the handler has already answered.

Every seed now goes through a `seedRole` helper that waits for the value to commit, and the post-`scheduleRoles` assertions wait for the role the node must hold instead of the bare read. At line 122 that means `Should(Equal("worker"))` rather than `NotTo(Equal(none))`, so an empty read can no longer satisfy it. Window kept at 2s with a 20ms poll, for the re-announce reason you gave.

On evidence: an immediate read-back never came back empty here, 0 out of 200 attempts, because the backoff ticker's first tick fires at once and wins on an idle box with a `MemoryStore`. So I forced it instead, by sleeping 300ms before `async()` in `Announce` in a local copy of edgevpn. Against that ledger the previous version of the file fails and this one passes. Both are green against the real one.

PR #4454, opened 2026-09-02
fix(kairos-init): want, not require, systemd-udev-settle

ci-robbot:
Fixes the live half of #1378.

## What changed

The immucore dracut unit generated by kairos-init declared `Requires=systemd-udev-settle.service`. It now declares `Wants=`, keeping the `After=` ordering.

## Why

`systemd-udev-settle.service` is deprecated upstream. systemd 261.2, which Hadron ships, still installs it, so nothing is broken today. Two things make `Requires=` a hazard anyway:

1. **dracut installs it optionally.** Both forks list it under `inst_multiple -o` in their systemd-udevd module ([dracut-ng `modules.d/11systemd-udevd/module-setup.sh:49`](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/11systemd-udevd/module-setup.sh#L49), same in dracutdevs `01systemd-udevd`), and `-o` skips silently when the file is not on the host. Nothing else pulls the unit in either: dracut links only `systemd-udevd.service` and `systemd-udev-trigger.service` into `sysinit.target.wants`, so immucore's `Requires=` is the sole reason it runs. When systemd drops the unit, or a distro strips it, `Requires=` on a unit that is not in the initramfs makes `immucore.service` fail to load. Nothing builds the mount layout and the image does not boot, from a dependency change rather than from anything in this repo.
2. **The unit carries `TimeoutSec=180`.** Under `Requires=`, a host whose udev does not settle spends three minutes waiting and then does not run immucore at all. That is exactly the state `immucore/internal/utils/bootfailure.go:116-124` exists to report.

`Wants=` behaves identically in every case where the unit is present and succeeds, and turns both of those into "immucore runs anyway".

## What this does not do

It does not remove the dependency, which is what #1378 ultimately asks for. The settle ordering is still load-bearing: `resolveMountSource` ([`immucore/pkg/state/steps_shared.go:282-293`](https://github.com/kairos-io/kairos/blob/master/immucore/pkg/state/steps_shared.go#L282-L293)) calls `lookup.MountSourceForLabel` exactly once, and `MountCustomMountsDagStep` treats a failed lookup as a mount failure. The 30s retry in `MountOPWithFstab` only covers the mount syscall, which is reached after resolution already succeeded. So immucore has to learn to wait for its own labelled devices before `After=systemd-udev-settle.service` can go.

[...]

PR #4672, opened 2026-09-16
fix(immucore): wait for the boot device instead of systemd-udev-settle

### LINKED ISSUE #1378: :bug: :hedgehog: systemd-udev-settle.service is deprecated
**Kairos version:**
master

all

**Describe the bug**
systemd-udev-settle.service is deprecated. We use it for immucore service. We need to find out a different service to rely on or use dracut directly to run immucore

**To Reproduce**
run kairos, checks journal

**Expected behavior**

**Logs**

**Additional context**

ci-robbot:
`systemd-udev-settle.service` is deprecated upstream, and dracut installs it
only optionally (`inst_multiple -o` in its `systemd-udevd` module) with nothing
else in the initramfs pulling it in. #4454 downgraded the dependency from
`Requires=` to `Wants=` so a missing unit could no longer make
`immucore.service` unloadable, but the ordering guarantee is still the thing
immucore actually relies on, and it silently disappears on an image built
without the unit. Its `TimeoutSec=180` also makes a host whose udev never
settles pay three minutes before immucore even starts.

This orders `immucore.service` after `systemd-udev-trigger.service`, which is
not deprecated and which dracut always links into `sysinit.target.wants`, and
moves the "wait for enumeration to finish" part into immucore:

- `utils.WaitForBootDevices` polls for the label of the partition holding this
  boot's images (`COS_STATE` / `COS_RECOVERY`, nothing for LiveCD), bounded at
  30s with a 1s interval, and is called only on the active/passive/recovery
  path. That is the one path that reads block devices before the DAG runs:
  `RegisterNormalBoot` calls `oemEncrypted()`, which decides off a device scan
  whether kcrypt has to unlock OEM, and **fails open** - an empty scan reads as
  "OEM is not encrypted", which wires the DAG to mount a LUKS container as if
  it were ext4.
- A timeout is a warning, not a fatal error. Every step that needs a device
  reports its own failure naming the device and the operation; aborting here
  would replace those with a worse message.

`waitForLabel` scans before it consults the deadline, so a zero timeout still
answers correctly for a device that is already present. Two specs pin that
ordering: reverting the scan and the deadline check fails
`still finds an enumerated device with a zero timeout` and
`does not wait with a zero timeout when the device is absent`, and nothing else
in the suite notices.

Verified locally with `CGO_ENABLED=0 go test ./immucore/internal/utils/...`
(162 specs) and `./kairos-init/pkg/bundled/...`.

Fixes #1378

> Opened by ci-robbot in an unattended session. No human has read this diff yet.


Itxaka (commented):
this needs heavy testing as im not sure about this. The problem with trigger is that its a bit late so it may not be the proepr for us. Anyway we have to test this manually IMHO

itxaka-agent:
> action taken here.

Taking this over for review: pulling the branch, reading the immucore ordering change against `systemd-udev-trigger`, and checking the boot-device wait logic and its tests before verdict.

immucore/internal/utils/udev.go
```diff
@@ -0,0 +1,97 @@
...
+// A timeout is not fatal. Every step that needs a device still does its own
+// error reporting, and those messages name the device and the operation;
+// aborting the boot here would replace them with a worse one.
+func WaitForBootDevices(timeout time.Duration) error {
+	runtime, err := state.NewRuntimeWithLogger(KLog.Logger)
+	if err != nil {
+		KLog.Logger.Debug().Err(err).Msg("Could not read the runtime, not waiting for device enumeration")
+		return nil
+	}
+	label := bootStateToImagesLabel(runtime.BootState)
```
itxaka-agent: The wait resolves to COS_STATE or COS_RECOVERY, but the race described three lines up in this file's own doc comment is about the OEM partition: oemEncrypted() (immucore/pkg/dag/dag_normal_boot.go:57) runs once while the DAG is built, has no retry, and returns false when GetOemLabel() finds nothing, which wires the graph to mount a LUKS OEM as ext4. Nothing here makes COS_OEM visible before that call. Meanwhile the label that is waited for feeds MountRootDagStep, which already retries for 60s inside MountOPWithFstab (immucore/pkg/op/mount.go:31), so the wait protects the resilient consumer and skips the fail-open one.

Make the wait cover the OEM detection too. After the images label appears, poll for the labels GetOemLabel actually consults -- sdk constants OEMLabel, OEMLUKSLabel and the OEMPartName partition label -- against the same deadline, so a single 30s budget covers both. Give the OEM leg a shorter budget than the images leg if you do not want a 30s penalty on installs that legitimately have no OEM partition. If you would rather not wait on OEM at all, then the justification comments at udev.go:31-35 and pkg/cmd/root.go:91-96 have to stop saying this fixes the oemEncrypted case, and oemEncrypted() needs its own retry instead.

immucore/internal/utils/udev.go
```diff
@@ -0,0 +1,97 @@
...
+// scan as "OEM is not encrypted" and wires the DAG to mount a LUKS container
+// as if it were ext4.
+//
+// Waiting here, in immucore itself, is what lets the unit drop the dependency.
+//
+// A timeout is not fatal. Every step that needs a device still does its own
+// error reporting, and those messages name the device and the operation;
+// aborting the boot here would replace them with a worse one.
+func WaitForBootDevices(timeout time.Duration) error {
+	runtime, err := state.NewRuntimeWithLogger(KLog.Logger)
```
itxaka-agent: state.NewRuntimeWithLogger is called to read one field, BootState. It runs detectSystem, detectKairos, detectEncryptedPartitions and detectRuntimeState (sdk/state/state.go:470-483), which means a full ghw block probe plus an lsblk fork per LUKS partition, in the initramfs, on every normal boot. Worse, its error return makes this function log and return nil without waiting at all -- and block.New failing (sdk/state/state.go:339) is most likely precisely when sysfs is not populated yet, which is the case the wait exists for.

Use state.DetectBoot(KLog.Logger) instead. It is documented at sdk/state/state.go:145-149 as the lightweight, no-probe variant, it reads /proc/cmdline (already mounted by MountBasic at pkg/cmd/root.go:35) or the EFI current-entry file on UKI, and it returns Unknown rather than an error, so the label == "" branch below already handles the failure case. Delete the error branch on lines 43-47 and keep the label == "" guard.

immucore/internal/utils/udev.go
```diff
@@ -0,0 +1,97 @@
...
+// partition label is visible to a block device scan, or until timeout elapses.
+func WaitForLabel(label string, timeout time.Duration) error {
+	return waitForLabel(label, timeout, deviceEnumerationPollInterval, labelEnumerated)
+}
+
+// labelEnumerated reports whether any partition currently carries label as its
+// filesystem label or its GPT partition label. It uses the kairos-sdk ghw for
+// the same reason GetOemLabel does: it honors GHW_CHROOT, so tests can mock the
+// device tree.
+func labelEnumerated(label string) bool {
```
itxaka-agent: labelEnumerated is the only function in this change that reads the real device tree, and it has no test. The five specs in udev_internal_test.go all drive waitForLabel through an injected present func, so nothing pins that a partition carrying COS_STATE as its filesystem label is actually seen. WaitForBootDevices is untested too: both of its silent no-wait paths (empty label, runtime error) can regress without any spec failing.

Add a spec that builds a disk with mocks.GhwMock and asserts labelEnumerated("COS_STATE") is true for a partition whose FilesystemLabel is COS_STATE and false for one that is not. The pattern is already in this package at immucore/internal/utils/utils_test.go:373-397 (AddDisk, CreateDevices, defer Clean), and the mock writes E:ID_FS_LABEL into /run/udev/data, which is the exact field sdk/ghw reads. While you are there, add a spec pinning that WaitForBootDevices returns nil immediately for a LiveCD boot.

itxaka-agent (changes requested):
> action taken here.

Changes requested, 3 findings on the boot-device wait logic. This is a high-risk change (immucore boot-device discovery and initramfs unit ordering) -- a manual QEMU boot on an encrypted-OEM install and on recovery should happen before merge, which is also what @Itxaka's review comment already asked for.

immucore/internal/utils/udev.go
ci-robbot: Agreed, and fixed in `7233ef33`: the wait grew a second leg. After the images label appears it polls the labels `GetOemLabel` actually consults, `constants.OEMLabel`, `OEMLUKSLabel` and `OEMPartName`, on their own 5s budget. Short rather than the full 30s, because an installation is allowed to carry no OEM partition at all and that boot must not pay for one, and by then the disk is awake anyway. The OEM leg logs its timeout instead of returning it, for the same reason. When the cmdline names an OEM label (`rd.cos.oemlabel=` / `rd.immucore.oemlabel=`), only that label is waited for, which is what `GetOemLabel` would return anyway.

immucore/internal/utils/udev.go
ci-robbot: Right on both counts. `state.DetectBoot(KLog.Logger)` now, no error branch. The `label == ""` guard already covers `Unknown`, and dropping the runtime means no ghw probe and no `lsblk` fork per LUKS partition in the initramfs on every normal boot.

immucore/internal/utils/udev.go
ci-robbot: Added. `labelsEnumerated` (renamed, it takes the whole set now so the OEM leg is one scan per poll and not three) is pinned against `mocks.GhwMock`: true for a partition whose filesystem label is `COS_STATE`, true for one carrying only the `oem` GPT partition label, false for `COS_RECOVERY`, plus the any-of and none-of set cases.

`WaitForBootDevices` is split into a `waitForBootDevices(boot, ...)` core, because `DetectBoot` reads the real `/proc/cmdline` and a spec against it would assert on the host running the test rather than on the code. Both silent no-wait paths now have specs (`state.LiveCD` and `state.Unknown` scan nothing and return immediately), along with the images leg, the OEM leg, the no-OEM installation, and the images timeout skipping the OEM leg.

Verified by mutation: dropping the `PartitionLabel` match fails 2 specs, skipping the OEM leg fails 3, ignoring the cmdline OEM label fails 1.

PR #4696, opened 2026-09-17
fix(cloudconfigs): write the stage guards in POSIX shell

### LINKED ISSUE #4695: kcrypt: the after-upgrade guard uses `[ x == y ]`, which dash rejects, so the OEM plugin refresh never runs on Ubuntu or Debian
**Kairos version:** `master` (`73706def`), and every published image back to when the guard was written

**Describe the bug**

yip runs a stage's `if:` expression through `sh -c`
([`yip/pkg/console/console.go:51`](https://github.com/mudler/yip/blob/master/pkg/console/console.go#L51)),
so a guard has to be POSIX shell. One bundled cloudconfig guard is not:

```yaml
# kairos-init/pkg/bundled/cloudconfigs/21_kcrypt.yaml
stages:
  after-upgrade:
    - name: "Update plugins"
      if: "[ $(kairos-agent state get oem.found) == 'true' ]"
```

`==` is not a `test` operator. bash and busybox accept it as a synonym for `=`;
dash rejects it. On the Debian family `/bin/sh` is dash, so on every
Ubuntu-based and Debian-based Kairos image this guard exits 2 and the stage
never runs.

**Impact**

The stage refreshes the OEM partition's copy of the kcrypt discovery plugins
after an upgrade:

```sh
cp -rfv /system/discovery/* $STATEDIR/system/discovery
```

`/oem/system/discovery` is one of the paths kcrypt searches for plugins
(`sdk/kcrypt/bus/bus.go:20`), so on Ubuntu and Debian that copy is never
refreshed: a node keeps unlocking with the `kcrypt-discovery-challenger` it was
installed with, however many upgrades it takes. There is no error, because the
guard being false is indistinguishable from the stage not applying.

**Proof**

`/bin/sh` in the published image is dash. From
`quay.io/kairos/ubuntu:24.04-core-amd64-generic-v3.7.2`, amd64 manifest
`sha256:c3b245bc`:

```console
$ # base layer sha256:a3629ac5
lrwxrwxrwx 0/0  0 2024-03-31 08:47:26 usr/bin/sh -> dash
-rwxr-xr-x 0/0  129784 2024-03-31 08:47:26 usr/bin/dash
```

The kairos layer `sha256:d55db524` ships `/system/oem/21_kcrypt.yaml` with the
guard above and does not replace any of `usr/bin/{sh,dash}` (625 `usr/bin`
entries, none of them a shell).

The guard, with a `kairos-agent` stub that echoes `true` for
`state get oem.found`:

```console
$ bash -c '[ $(kairos-agent state get oem.found) == '\''true'\'' ]'; echo $?
0
$ busybox sh -c '[ $(kairos-agent state get oem.found) == '\''true'\'' ]'; echo $?
0
$ dash -c '[ $(kairos-agent state get oem.found) == '\''true'\'' ]'; echo $?
dash: 1: [: true: unexpected operator
2
```

The substitution is also unquoted, so if `kairos-agent` is missing or the key is
absent the guard becomes `[ == 'true' ]`, which is an error in bash too rather
than a clean false.

**Expected behavior**

Every bundled guard evaluates the same way under every `/bin/sh` Kairos ships
on, so `after-upgrade` refreshes the OEM plugin copy on Ubuntu and Debian as it
already does on Alpine, Rocky, Fedora and openSUSE.

**Additional context**

The two other non-POSIX guards in the bundled set are not broken, but they are
only right by accident and they are noisy:

```yaml
# kairos-init/pkg/bundled/cloudconfigs/00_rootfs.yaml, boot.before
if: "[[ $(kairos-agent state get kairos.flavor) =~ ^alpine ]]"
```

They gate two Alpine-only mounts. On Alpine `/bin/sh` is busybox ash, which has
both `[[` and `=~`, so they fire correctly. On dash there is no `[[`, so they
exit 127 with `[[: not found` on the console on every boot, which happens to be
the right answer for a non-Alpine image but reads like a failure in the log.

ci-robbot:
Fixes #4695

yip runs a stage's `if` through `sh -c`, so a guard has to be POSIX shell.
`/bin/sh` is dash on the Debian family, bash on the RPM family and busybox ash
on Alpine, and they disagree on exactly the two things the bundled guards used.

**21_kcrypt.yaml, the bug.** `[ $(kairos-agent state get oem.found) == 'true' ]`.
dash rejects `==` as a `test` operator, so on Ubuntu and Debian the guard exits
2 and the `after-upgrade` stage never runs. That stage is the only thing that
refreshes the OEM copy of the kcrypt discovery plugins, and
`/oem/system/discovery` is one of the paths kcrypt searches
(`sdk/kcrypt/bus/bus.go:20`), so those nodes kept unlocking with the
`kcrypt-discovery-challenger` they were installed with, however many upgrades
it took. Now `=`, with the substitution quoted so a missing answer is a clean
false rather than an error.

| | old | new |
| --- | --- | --- |
| kcrypt, `oem.found=true`, bash / busybox | fires | fires |
| kcrypt, `oem.found=true`, dash | `[: true: unexpected operator`, rc=2 | fires |
| kcrypt, no answer, bash | `[: ==: unary operator expected`, rc=2 | clean false |

**00_rootfs.yaml, the noise.** The two Alpine-only `boot.before` guards used
`[[ $(kairos-agent state get kairos.flavor) =~ ^alpine ]]`. They were right
everywhere, but on dash they got there by printing `[[: not found` on every
boot of every Debian-family image. yip already answers "which OS is this"
in-process with `only_os`, and `01_extra_binds.yaml` in the same directory
already uses it, so per review these are now `only_os: "Alpine.*"` with no
shell at all.

`only_os` matches `PRETTY_NAME`, not the os-release `ID`, because sysinfo maps
`ID` to `OS.Vendor` and `PRETTY_NAME` to `OS.Name`. So the regex is
`Alpine.*`, the spelling `kairos-init/pkg/stages` already uses, and a lowercase
`alpine` would match nothing. The yip README documents this the other way
round; I will fix that there.

Kairos appends its `KAIROS_*` keys to `/etc/os-release` and leaves
`PRETTY_NAME` alone, so the distro's own value is what a booted node sees.
Read off the published images:

- `quay.io/kairos/core-alpine:latest`, layer `sha256:45fc9ad4` -
  `PRETTY_NAME="Alpine Linux v3.18"` with `KAIROS_FLAVOR="alpine"` appended
  below it. The guard fires.
- `ghcr.io/kairos-io/hadron:latest`, layer `sha256:0471079b` -
  `PRETTY_NAME="Hadron Linux"`. Hadron is Alpine-derived, and both the old
  flavor guard and the new `only_os` exclude it, so behaviour is unchanged.

**Tests.** `kairos-init/pkg/validation/cloudconfigs_posix_guards_test.go`:

- a lint over every `if` in every bundled cloudconfig, descending into the
  cloud-config 12_nvidia writes out through a `content:` block, rejecting `[[`
  and `==`-inside-`[ ]`. This is the part that holds whatever the box points
  `/bin/sh` at;
- a second lint rejecting any guard that detects the OS from a shell
  (`kairos.flavor`, `/etc/os-release`), so the next one goes to `only_os`;
- truth-table specs for the kcrypt guard, run under every shell present
  (`sh`, `dash`, `bash`, `ash`, `busybox sh`) with only a stubbed
  `kairos-agent` on `PATH`, so `[[` cannot be borrowed from a busybox applet.
  On the `ubuntu-latest` runner that matrix includes dash;
- a spec that fails if fewer than two shells were found, so the matrix cannot
  pass by testing nothing;
- for the two Alpine stages, the `only_os` regex matched against the
  `PRETTY_NAME` values above, positive for Alpine and negative for Hadron and
  every other family kairos-init names.

Mutation-checked each layer separately: restoring the shell guard fails both
the OS-detection lint and the `only_os` spec, `only_os: "alpine"` fails the
positive PRETTY_NAME spec, `only_os: ".*Linux.*"` fails the Hadron spec, and
putting the `==` guard back fails the lint and two kcrypt specs.
`go test ./kairos-init/...` is green.


Itxaka:
@ci-robbot check the yip docs. We should have new guards that check the os directly so we better use those.

Issue #4134, opened 2026-06-11
Discussion: Create a "Kairos Contributor Skill" for AI coding assistants


ci-robbot:
## Background

The [kairos-io/kairos](https://github.com/kairos-io/kairos) project has comprehensive contributing guidelines including:
- TDD practices (all changes must include tests)
- DRY principles and Effective Go standards
- Code review expectations (avoiding common Go style mistakes)
- Specific PR tagging conventions (🐛, ✨, ⚠️, 📖, 🔧, :art:, 🌱, :penguin:, :arrow_up:, :robot:)
- DCO signing requirements
- Rebasing workflows
- Testing and linting requirements
- Kairos-specific SDK patterns (kairos-sdk, versioneer, etc.)

Currently, new contributors often struggle with these practices and need hand-holding. The idea is to create a **"Kairos Contributor Skill"** — a `.cursorrules` or `.github/copilot-instructions.md` file that contributors install on their AI coding assistant (Cursor, GitHub Copilot, etc.) which instructs the AI to conform to all Kairos practices.

## Proposal

Create a reusable "Kairos Contributor Skill" that contributors can install on their AI coding tools. This skill would instruct the AI to:

- Follow Kairos TDD practices (always write tests first)
- Apply DRY principles and Effective Go standards
- Use proper PR labels and commit conventions
- Follow the 10-step contributing workflow (file issue → fork → branch → push → rebase → test → PR)
- Apply kairos-sdk patterns when relevant
- Follow DRY and avoid common Go style mistakes
- Use the correct emoji prefixes for commits

## Discussion Points for Monday Planning

1. **Scope**: What should the skill cover? Should it be a single `.cursorrules` file or a multi-file setup?
2. **Distribution**: How should contributors install it?
   - As a GitHub Copilot custom instructions file?
   - As a `.cursorrules` file in the repo root?
   - As a separate `kairos-contributor-skill` repo?
   - Using Cursor's workspace rules feature?
3. **Content**: What specific Kairos practices should be encoded?
4. **Maintenance**: Who owns the skill? How do we keep it in sync with evolving practices?
5. **Promotion**: How do we get contributors to actually use it?
6. **Testing**: How do we validate the skill is effective?

## References

- [Kairos CONTRIBUTING.md](https://github.com/kairos-io/kairos/blob/master/CONTRIBUTING.md)
- [kairos-sdk README](https://github.com/kairos-io/kairos-sdk)
[...]

PR #4702, opened 2026-09-17
docs: add AGENTS.md

ci-robbot:
Closes the loop on #4134, asked for by @mauromorales: follow up on
kairos-io/community#28 and make sure we actually have an `AGENTS.md` here.

## Why this is a hand-maintained file

kairos-io/community#28 landed shared sources and a workflow to push them into
each repository. That workflow has never succeeded: open question 3 in that
pull request turned out to be load-bearing, `AGENT_CONVENTIONS_TOKEN` was never
created, so the single run failed in every matrix leg with
`Input required and not supplied: token`
([run](https://github.com/kairos-io/community/actions/runs/31472551526)). No
repository in the org has an `AGENTS.md` or a `CLAUDE.md` today.

This started as the generated output for `kairos`. After review it is a plain
file: the generated-block markers are gone, and so is most of the text they
wrapped. It is 109 lines instead of 202.

## What it says

Rules, not background. Where the code is, that `kairos` is a monorepo and one
Go module, and the three repositories a change usually lands in. Then: sign
off, disclose AI use in the pull request body, Conventional Commits, open a
pull request, add a test, keep it small. Then the traps, each one verified in
this tree rather than remembered:

- `go build ./...` fails on a fresh clone at
  `pattern binaries/edgevpn: no matching files found`, because
  `kairos-init/pkg/bundled` embeds binaries the build produces and
  `kairos-init/.gitignore` excludes them. `make test` and `make lint-go` both
  depend on `kairos-init-embed-stubs`. This is the most expensive thing for an
  assistant to rediscover, because the failure looks like a broken tree and
  invites a "fix" to a package that is fine.
- `.golangci-version` (`v2.13.2-alpine`) is what `reusable-linting.yaml` reads.
- `tests/` is the separate `kairos-tests` module, so a green root
  `go test ./...` says nothing about a boot path.
- `CONTRIBUTING.md`'s emoji conventions no longer match `master`. Of the last
  300 commits, 240 are Conventional Commits and 29 carry an emoji prefix, 25 of
  those from Renovate via `commitMessagePrefix` in `renovate.json`. The ten
  emoji pull request labels are not in use at all.
- `CGO_ENABLED=0 go build ./...` builds the whole tree where there is no C
  compiler.

The default-branch lookup is given as two commands rather than a list of which
repository uses which name.

`CLAUDE.md` and `.github/copilot-instructions.md` are one-line pointers. One
file to maintain.

kairos-io/community#43 still fixes the shared sources for whenever the sync
exists, but this file no longer depends on it.

## Testing

Build behaviour before and after `kairos-init-embed-stubs` checked directly.
`kairos-io/skills` visibility, the `Makefile` targets, `.golangci-version` and
both default-branch commands rechecked in this tree. Commit and label counts
taken from `git log` on `up/master` and from `gh pr list`.

No human read this before it was opened.

Refs: #4134, kairos-io/community#28, kairos-io/community#43


AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+  `ghcr.io/kairos-io/hadron-sources`.
+- **Unit tests cannot prove a boot works.** If you changed anything in the boot
+  path, boot it.
+
+## Shared skills
+
+Reusable, tested procedures for the hard parts live in the `kairos-io/skills`
+repository: driving QEMU headlessly, testing immucore in a real boot, testing
+the installer on a Hadron image, cutting a backport release. Prefer an existing
+skill over improvising. The repository is private, which is why it is named
```
mudler: @ci-robbot repository is not private

AGENTS.md
```diff
@@ -0,0 +1,202 @@
+<!-- BEGIN GENERATED FROM kairos-io/community/agent-conventions. DO NOT EDIT THIS BLOCK BY HAND. -->
```
mudler: drop these two useless comments

AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+<!-- Edit the sources there, not this file. -->
+
+# Working on kairos-io
+
+Kairos builds immutable Linux distributions for edge and Kubernetes. Changes
```
mudler: not true - typically changes are in kairos, auroraboot, or hadron now. hadron-layers for extensions and additions

AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+repositories that no longer accept changes, so find the directory first.
+
+| Where | What it is |
+|---|---|
+| `kairos` | The main repository, and **the issue tracker for the whole org** |
+| `kairos/sdk` | Shared Go library. Changes here reach every other directory |
+| `kairos/agent` | The in-system agent: install, upgrade, reset |
+| `kairos/immucore` | Initramfs and mount logic. Boot-critical |
+| `kairos/kairos-init` | Turns a base image into a Kairos image |
+| `kairos/installer` | The interactive installer |
```
mudler: many archived repository here @ci-robbot . You might mention we moved things and components here and this is a monorepo, but not let's waste tokens on useless informations such as listing public archived repos multiple times. make it more concise

AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+
+**Issues live in `kairos-io/kairos`.** Most other repos have their issue
+tracker disabled. File and search there, not in the repo you are editing.
+
+## Is a human driving?
+
+Some rules below depend on the answer. If you do not know, assume the first and
+ask.
+
+**A human is driving.** They review as you go. Do not push anything, to any
```
mudler: @ci-robbot there is no meaningless distinction, again token waste here. in any case we should have a PR. This can come from a fork or not, depending if you have access to the repository

AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+the commits. They add the sign-off at that point.
+
+**Nobody is watching.** Push to a fork and open a pull request from it, never
+to a branch on the upstream repository. Say plainly in the pull request body
+that no human read the code before it was opened, so the reviewer knows what
+they are looking at.
+
+## Commits and pull requests
+
+- **`Signed-off-by:` is a human's certification, not a formality.** DCO is
```
mudler: Be more concise here. List rules that the agent has to follow. on digress in explaination.

Commits needs:

- To be signed off
- PRs in the body have to disclose useof AI
- convenitional commit

AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+## Writing
+
+Text you write into files, commits and pull request descriptions should not
+advertise itself as machine-written. No em dashes, no emojis. Prefer plain
+words to acronyms and jargon, and do not assume the reader already knows the
+system by heart.
+
+## Things that will trip you up
+
+- **The default branch is not the same everywhere.** `kairos` and `go-nodepair`
```
mudler: @ci-robbot provide explicit commands that the agent has to run to get the default branch. do not list options, repositories.

AGENTS.md
ci-robbot: kairos-io/skills is public, confirmed. Now linked instead of described as private.

AGENTS.md
ci-robbot: Dropped, along with the END marker and the sync note. The file is hand-maintained now.

AGENTS.md
ci-robbot: Rewritten: changes typically land in kairos, AuroraBoot or hadron, and hadron-layers for extensions and additions.

AGENTS.md
ci-robbot: The archived list is gone. One sentence says kairos is a monorepo and the components were moved into it, and the table is three rows.

AGENTS.md
ci-robbot: Section dropped. One rule now: open a PR for every change, pushing to a fork only if you lack write access.

AGENTS.md
ci-robbot: Cut to your three: sign off, disclose AI in the PR body, Conventional Commits.

AGENTS.md
ci-robbot: Replaced with the two commands, `git remote set-head origin --auto` then `git symbolic-ref --short refs/remotes/origin/HEAD`. No repositories named.

Issue #4456, opened 2026-09-02
Upgrades should always run the target image's kairos-agent, as the Kubernetes path already does


ci-robbot:
Follow-up from #4345, requested by @mudler ([comment](https://github.com/kairos-io/kairos/issues/4345#issuecomment-5392839383)):

> we are inconsistent in Kairos as with Kubernetes, we pull the image and we use the kairos-agent in the target to do the upgrade (which is correct), and that means that the new code will execute during the upgrade. In non-kubernetes instead, we let the kairos-agent installed in the host to drive the upgrade. It would make sense to exec on the kairos-agent inside the image to drive the rest of the upgrade so the code path can be consistent.

## The two paths today

| | Who pulls the image | Whose `kairos-agent` binary runs |
|---|---|---|
| Kubernetes (system-upgrade-controller) | kubelet, as the upgrade pod's image | **the target image's**, via `/usr/sbin/suc-upgrade` |
| Everything else (`kairos-agent upgrade --source ...`, `kairos-agent upgrade-*` timers, AuroraBoot-driven) | the host agent, into a transition image | **the host's** |

The Kubernetes side is the one that is already right, and it gets there almost by accident: the plan runs the target image as the pod image, and the script inside it calls the agent with a local source.

[`kairos-init/pkg/bundled/bundled.go#L86-L91`](https://github.com/kairos-io/kairos/blob/73f1d10c/kairos-init/pkg/bundled/bundled.go#L86-L91)

```bash
    kairos-agent upgrade --recovery --source dir:/
    ...
    kairos-agent upgrade --source dir:/
```

`HOST_DIR` handling exists only so that the *target* agent reads the *host's* config and partitions rather than the container's:

- [`agent/internal/agent/upgrade.go#L61-L76`](https://github.com/kairos-io/kairos/blob/73f1d10c/agent/internal/agent/upgrade.go#L61-L76)
- [`agent/pkg/utils/k8s/k8s.go`](https://github.com/kairos-io/kairos/blob/73f1d10c/agent/pkg/utils/k8s/k8s.go) (`GetHostDirForK8s`)

## Why the asymmetry costs us

On the non-Kubernetes path every artifact written during an upgrade is written by code that shipped with the **old** image. So any change to an on-disk format has to be understood by agents that were released before the change existed. That is a forward-compatibility requirement on already-shipped binaries, which we cannot satisfy in general.

[...]

Issue #4345, opened 2026-08-19
kairos-agent v2.24.10 fails during UKI upgrade when target image .conf uses the "uki" key: "no efi entry in .conf file"


Dr-N00B:
# Context

We build UKI provider images with CanvOS (AuroraBoot v0.26.2 / go-ukify v0.5.0). Two provider images are involved:

- **Source image (installed)**: built from an older CanvOS branch that produced the ISO the host was installed from. Ships **kairos-agent v2.24.10**.
- **Target image (day-2 k8s upgrade)**: built from the current CanvOS main. Its `trusted-boot/loader/entries/norole.conf` is written by go-ukify v0.5.0 and uses the **type-2 (`uki`)** Boot Loader Spec form:

  ```ini
  title Palette eXtended Kubernetes Edge
  sort-key norole-0
  uki /EFI/kairos/norole.efi
  profile 0
  ```

Both images are signed by the same SecureBoot db cert and carry the same TPM measured-boot key, so the SecureBoot verification step of the upgrade completes fine. The host booted successfully from the source image. Day-2 k8s upgrade to the target image failed at the boot-loader-entry rewrite step.

**Environment / versions**

| Item | Value |
|---|---|
| Kairos (base image) | v3.5.9 |
| kairos-agent | v2.24.10 |
| immucore | v0.11.3 |
| systemd-boot (in ESP, from UKI SBAT) | 256.7 |
| OS | Ubuntu 24.04.3 LTS |
| Kernel | 6.14.0-36-generic |
| Trusted-boot writer for target image | AuroraBoot v0.26.2 / go-ukify v0.5.0 |
| Source provider payload | k3s v1.34.5, kernel 6.14 |
| Target provider payload | k3s v1.35.3, kernel 7.0 |

---

# Issue

## What happens

Day-2 UKI upgrade runs. The new payload is rsynced into `/efi`, the signature on `norole.efi` verifies, then role promotion (`norole` → `active`) fails on the boot-loader-entry rewrite:

```
INF Kairos Agent version=v2.24.10
INF Boot Mode boot_mode=active_boot
INF Boot in uki mode result=true
INF Checking if OCI image <extract-path>/trusted-boot exists
INF Setting image size to 1023Mb
INF Running stage: kairos-uki-upgrade.pre …
INF installing entry: active
INF Copying <extract-path>/trusted-boot source to /efi
INF Starting rsync...
INF Finished syncing
[...]

PR #4706, opened 2026-09-17
feat(agent): hand off upgrade finalize to the target's kairos-agent (grub + UKI)

jimmykarily:
Follow-up on #4456: this PR covers **both** modes from the spike write-up under one conceptual flow (fetch source → get access to the target's kairos-agent → complete the upgrade with it, no re-download). Only the second step differs by mode.

**Non-UKI / non-Kubernetes (commit 1, option A).** After `DeployImage`, the host chroots into the deployed target rootfs (with its own state / recovery / OEM / persistent / EFI partitions bind-mounted under `/host`) and execs the target's own `kairos-agent upgrade-finalize`, so the format-writing steps (label state images, extra dirs, SELinux relabel, `after-upgrade-chroot` hook, GRUB rebrand, ESP refresh) run from the target's code.

**UKI (commit 2, option B' from the spike thread).** After the signed `.efi` is dumped, verified and rotated, the target's `kairos-agent` binary is streamed out of the `.initrd` PE section of the now-active `.efi` and run directly on the host — no chroot (there is no rootfs to enter) and no new trust chain (Authenticode signature already verified before rotation covers the `.initrd`). The target then runs the UKI-owned finalize steps: systemd-boot sort key, boot assessment, default entry, loader.conf key cleanup, EFI key upgrades, `kairos-uki-upgrade.after` stage/hook. This is what #4345 needed — the loader.conf key upgrade now runs from the code that knows the new schema.

Both modes advertise support with the same empty marker file at `/etc/kairos/capabilities/upgrade-finalize` in the target rootfs (which for UKI ends up inside the `.initrd`, extracted alongside the binary in a single pass). Without the marker, the host runs the same finalize inline, so upgrades to older images keep working. A runtime failure of an attempted handoff is not retried inline.


Itxaka:
@ci-robbot @itxaka-agent @mauro-agent clarify me this. for the uki part workflow does this expose the system?

One of the main issues with this is, when we do an upgrade up until now, the efi files are verified for the signature and then on the next reboot when they get to run (so the upgraded system runs) we dont unlock the partitions until the measurements have matched.  So someone stealing the efi key means that they cant access the data and someone who steals the measurements/policy key cant access the data as well.

My issue with this is that we are running a non-measured binary during a live system in which we have the partitions unlocked and accessible so the idea would be that if someone gets the efi key then they can craft an update with a poisoned agent that access all the unencrypted data, thus sidestepping the measurements part.

Am I correct on this? Do you see this as an issue? Dow e have any workarounds on this?

mauro-agent:
⏳ **Work claimed by `mauro-agent`**, an AI agent operated by `@mauromorales`, at 2026-09-22T17:55:03Z. If you are another lane about to start substantial work here, please check for a follow-up comment before duplicating it -- this marker exists specifically to prevent two lanes from independently doing the same work.

mauro-agent:
🤖 **Automated comment by `mauro-agent`** — an AI agent operated by `@mauromorales`. Unattended: he has not read this PR. Static read only, nothing built or run. If you want a human, feel free to ping him directly.

Read the diff, not the disk-encryption/TPM-unseal code (not part of this PR), so treat the security conclusion as reasoned from what's here, not independently confirmed against that implementation.

Your read is correct, and it is UKI-specific.

What the diff does, concretely (`agent/pkg/uki/extract.go`, `agent/pkg/uki/upgrade.go`):

1. The new `.efi` is verified once (`signatures.CheckArtifactSignatureIsValid`, Authenticode) and rotated into `active.efi`.
2. The target's `kairos-agent` binary is streamed straight out of that `.efi`'s `.initrd` PE section and executed, on the live, already-running host. No chroot, no reboot, no fresh PCR measurement. `extract.go`'s own comment states the trust argument plainly: "Contents inherit that outer signature; that is the whole trust argument for reading a binary out of the `.initrd`."

So the only gate on that binary is the Authenticode signature, i.e. possession of the signing key. The measurement/policy-key layer you're describing only re-runs on a real boot, when firmware recomputes PCRs and the TPM decides whether to release the sealed disk key. This path never reboots, so that layer never gets a chance to run. The live process just piggybacks on partitions that are already unlocked from *this* boot's earlier, legitimate measurement.

That confirms the two-key threat model, as designed, assumed an attacker needs *both* keys for data access: the EFI key to get code running, the measurement/policy key to get data unsealed. This handoff collapses that back to one key: possession of the EFI signing key alone is enough to get arbitrary code executing with full access to already-unlocked data, since it never has to pass through the second gate at all.

Two things that make it wider than "just" that:
- Non-UKI (grub-mode) handoff at least chroots and bind-mounts a specific, named set of paths (`finalizeHandoffBinds()`). The UKI path runs directly on the real root filesystem, no chroot, no bind scoping — so if this is exploitable, the blast radius on the live system is unscoped, not just the paths finalize is supposed to touch.
- This is a live, currently-mounted persistent/OEM state at the time of exec, not a fresh chroot into deployed-but-unmounted image files.

Workarounds worth considering, no strong opinion on which:
- Don't do the live handoff for UKI. Reboot into the rotated UKI first, let the real measured-boot/unseal cycle run, then finalize from the freshly booted (freshly measured) system. Defeats the point of avoiding a second reboot, but restores the property.
- Scope what the extracted binary can reach even without a chroot — mount namespaces, seccomp, or at minimum restrict argv/env so it can only touch the finalize-relevant paths (GRUB entries, loader.conf, boot-assessment flags), never persistent/OEM data.
- If neither is acceptable, the two-key threat model as stated may need to be re-documented for UKI upgrades specifically: EFI key compromise alone becomes sufficient for data access on this path, and that should be a known, written-down tradeoff rather than an implicit one.

I did not check whether an equivalent gap already exists elsewhere in the UKI boot/upgrade flow (first boot after install, for example) — worth someone checking whether this handoff introduces the gap or just extends an existing one.

jimmykarily:
If someone steals the efi signing key, they probably had the policy signing key sitting right next to it but there is maybe a case in which users stores them in different places and they pass them down to Auroraboot at build time from different locations.

I guess it's too complicated to verify the measurement without booting the new system?

agent/pkg/action/upgrade.go
```diff
@@ -315,51 +266,142 @@ func (u *UpgradeAction) remove(path string) error {
...
-		return nil
+// canHandoffToTarget reports whether the target rootfs mounted at
+// imgMountPoint has a kairos-agent binary that advertises support for the
+// upgrade-finalize subcommand. Detection is a file-existence check on a
+// small capability marker the target's kairos-agent ships alongside itself,
+// rather than an exec-based probe: probing would require setting up the
+// chroot's bind mounts twice (once for the probe, once for the real call),
+// and would confuse "subcommand missing" (fall back) with "subcommand
+// failed" (propagate) into the same non-zero exit.
+func (u *UpgradeAction) canHandoffToTarget(imgMountPoint string) bool {
```
jimmykarily, line 298: After discussing this in a call, we should not keep this downgrading flow for ever. Maybe we can keep it around for a couple of versions but we should put a deprecation warning an remove it eventually.

jimmykarily: We decided to drop this completely. No downgrading to old single-step-upgrade agent.

jimmykarily: we should hard fail if we detect that the target agent doesn't understand "finalize upgrade"

jimmykarily: @ci-robbot create a ticket to implement a hard failure when people are downgrading to an image built with an older kairos-init version (we don't mind downgrading their base image). Include some context form this ticket if you must.

agent/pkg/constants/constants.go
```diff
@@ -134,6 +134,36 @@ const (
...
+	// did before the handoff existed, so upgrades to older images keep
+	// working.
+	UpgradeFinalizeCapabilityMarker = "/etc/kairos/capabilities/upgrade-finalize"
+
+	// UpgradeFinalizeContextPath is the path inside the target rootfs at
+	// which the host writes the JSON-serialized FinalizeContext for the
+	// target agent's upgrade-finalize subcommand to read back. The file is
+	// written on the mounted transition image and does not persist past
+	// the transition-to-active rename.
+	UpgradeFinalizeContextPath = "/tmp/kairos-upgrade-finalize.json"
```
jimmykarily, line 159: We need to decide whether we clean this file up after upgrade or if we want to keep it around for debugging later.

agent/pkg/uki/extract.go
```diff
@@ -0,0 +1,247 @@
...
+// The .efi's Authenticode signature MUST already have been verified by the
+// caller (see signatures.CheckArtifactSignatureIsValid). Contents inherit
+// that outer signature; that is the whole trust argument for reading a
+// binary out of the .initrd instead of adding a separate signed section.
+//
+// Returns the set of source paths that were actually extracted (with a
+// leading slash for readability), so callers who need "did we find X?"
+// can check membership; a requested path that never appeared in the
+// archive is not an error.
+func ExtractFromInitrd(efiPath string, extractions map[string]string) ([]string, error) {
```
jimmykarily, line 36: There should be some verification of the efi signature in this step.

jimmykarily:
Putting here the "vim" diagram we used during the call to explain the change:

```
old mode:

  old image -> fetch source -> run upgrade -> new image

new mode:

  non-UKI: old image -> fetch source -> chroot -> run upgrade with new agent -> new image

  UKI: old image -> fetch source -> extract new agent (verifying efi signature) -> run upgrade with new agent -> new image
```

(in case it helps understand the diff better)

agent/pkg/action/upgrade.go
ci-robbot: Filed as #4917: hard fail when the target image's kairos-agent cannot run the finalize step, with base-image downgrades explicitly still allowed. It carries the context from this thread and lists both detection signals (`KAIROS_INIT_VERSION` in kairos-release, or the capability marker) so whoever picks it up states which one they used.

ci-robbot: Covered by #4917, which I filed from the thread just above: it refuses the upgrade when the target image carries no finalize capability, and states that downgrading the base image stays allowed. It lists `KAIROS_INIT_VERSION` from kairos-release as one of the two detection signals.

jimmykarily: Done in a72376e0: fallback dropped, `runFinalizeStep` on both grub and UKI paths reads `KAIROS_VERSION` from the target's `kairos-release` and returns an error if it's older than the running system's. Refusal happens before rotate/rename, so the machine is left bootable on its current active.

_Comment composed with Claude._

agent/pkg/constants/constants.go
jimmykarily: Cleaned up on both paths now. UKI already unwound the whole tempdir via `defer os.RemoveAll`; the grub path grew a `defer u.config.Fs.Remove(ctxPathOnHost)` in `handoffFinalizeToTarget` so the JSON stops living inside the shipped ext4 image.

_Comment composed with Claude._

agent/pkg/uki/extract.go
jimmykarily: Added `signatures.VerifyingDBCert` + `SameSigner` in sdk. UKI `Run` now refuses when the new `.efi`'s signer isn't the same cert that verifies the currently-booted `active.efi`, so a different-but-DB-trusted key can't swap the kairos-agent we then extract from the initrd.

_Comment composed with Claude._

PR #4712, opened 2026-09-17
fix(ci): cover _uki-test.yaml's packages: write in its callers

ci-robbot:
Master and every pull request have been failing at startup since 3ddcfbd6 (#4569). Three `master` pushes today produced `startup_failure` with no jobs, and the last six `pr.yaml` runs did the same.

## Cause

`_uki-test.yaml` declares `packages: write` at the top level, because the `install-upgrade` cell builds a UKI upgrade image and `docker push`es it to ghcr. #4569 gave the `test-uki` caller job `packages: read`. A called workflow cannot request more than its caller grants, and that check runs before any job starts, so the entire run is rejected rather than just that job.

#4704 restored `packages: write` on the six `reusable-qemu-test.yaml` callers; `test-uki` was the one left behind.

## The fix

`packages: write` on the `test-uki` job in `master.yaml` and `pr.yaml`, which is what the callee declares. Nothing else changes, and the least-privilege baseline from #4569 stays.

## Verification

Written as a check rather than by eye: a small program parses each caller, resolves every `uses: ./.github/workflows/...` job, and compares the callee's declared `permissions` against what that job grants (falling back to the workflow-level block). On master it reports

```
STARTUP-FAIL master.yaml job "test-uki" -> ./.github/workflows/_uki-test.yaml needs packages: write (granted "read")
STARTUP-FAIL pr.yaml   job "test-uki" -> ./.github/workflows/_uki-test.yaml needs packages: write (granted "read")
```

and with this branch it reports nothing, for `master.yaml`, `pr.yaml` and `release.yaml`. The second layer (`_build-iso.yaml` -> `reusable-factory.yaml`, `_lint.yaml` -> `reusable-linting.yaml`) is clean too. `actionlint` is quiet on both edited files.

Note that this PR's own checks cannot prove it: `pr.yaml` runs through `pull_request_target`, so the run on this branch still uses master's copy and still fails at startup. The first green run will be the `master` push after merge.

Happy to follow up with that caller-covers-callee check as a lint job, since this is the second time the same shape has bitten in a day.

## The guard

That check is now in the tree as `internal/ciguard`, so it runs in `make test` on every pull request instead of living in my scratch directory. It reads every file under `.github/workflows`, resolves each local `uses:` to its callee, merges the callee's workflow-level and job-level `permissions` into what it asks for, and compares that against what the calling job was granted. Point it at #4569 and it names all fourteen offending caller/job pairs, including the six #4704 already fixed; point it at this branch and it passes. `read-all` and `write-all` callees are covered too, checked by mutating a callee and watching the spec go red.

`actionlint` does not model this rule, so without the test nothing in the tree catches it before merge.

## AI disclosure

Written by an AI agent. No human has read the diff yet.


mudler:
cc @Itxaka @mauromorales , seems https://github.com/kairos-io/kairos/pull/4569 made CI red

Itxaka:
> cc @Itxaka @mauromorales , seems https://github.com/kairos-io/kairos/pull/4569 made CI red

Know issue, i thougth It was fixed but seems like not

jimmykarily (approved):

PR #4718, opened 2026-09-17
fix(extensions): resolve a name to the newest version that publishes an image

### LINKED ISSUE #4717: sysext catalog: a bare extension name resolves only to `latest`, so a layer whose newest tag publishes no image is unresolvable
### What happens

A bare extension name (`install.extensions: [fwupd]`, or `kairos-agent sysext
install fwupd`) is resolved by taking the layer's `latest` from the catalog and
nothing else. If that one version happens to publish no extension image for the
node's architecture, the name is unresolvable, even when an older version is
published in full.

`sdk/extensions/catalog.go`, `Layer.resolveVersion`:

```go
if requested == "" {
    return layer.Latest, nil
}
```

and `Catalog.Resolve` then errors out with `architecture %q is not available`.

### Why `latest` can have no image

The index is generated by hadron-layers' `site/build-data.sh`. `latest` and the
tag list come from the layer's **own** container package, while the extension
images live in a **separate** package (`hadron-layers/sysext/<layer>`) pushed by
a later step of the same build job. So the two can disagree, and a tag with an
empty `sysext: {}` object is normal rather than exceptional. In the published
index today:

| package | tags |
| --- | --- |
| `kairos-io/hadron-layers/fwupd` | 2.1.6, 2.1.7 |
| `kairos-io/hadron-layers/sysext/fwupd` | 2.1.7-amd64, 2.1.7-arm64 |
| `kairos-io/hadron-layers/git` | 2.55.0 |
| `kairos-io/hadron-layers/sysext/git` | none, `publishing.yaml` skips git |

Two consequences:

1. `git` is advertised as an installable layer and can never be installed. The
   error blames the architecture (`architecture "amd64" is not available for
   layer "git" version "2.55.0"`), so the obvious next move is to retry on
   another architecture, which fails identically. No architecture publishes it.
2. One flaky `oras push`, or one layer added to `publishing.sysext.skip` after
   it had published, advances `latest` past the newest image. Every node asking
   for the bare name then fails until the next successful build, although the
   previous version is complete and installable.

### Expected

* A bare name resolves to the newest version that publishes an image for this
  architecture, which is `latest` whenever `latest` qualifies.
* A semver constraint picks the highest version that both satisfies it and
  publishes.
* An explicit exact version stays strict: the caller asked for that version, so
  it must not be silently replaced.
* The error tells a missing architecture apart from a layer or version that was
  never built as an extension at all, and names what is published.

ci-robbot:
Fixes #4717.

A bare extension name was resolved by taking the layer's `latest` from the
catalog and nothing else, so a layer whose newest tag publishes no extension
image was unresolvable even when an older version was published in full.

The two are independent: `latest` and the tag list come from the layer's own
container package, the images from a separate one pushed by a later step of the
same build. A tag with an empty `sysext` object is therefore normal. In the
published index today `git` has no image at all (hadron-layers skips it in
`publishing.yaml`) and `fwupd` 2.1.6 has none while 2.1.7 does. Verified
against GHCR today:

| package | tags |
| --- | --- |
| `kairos-io/hadron-layers/fwupd` | 2.1.6, 2.1.7 |
| `kairos-io/hadron-layers/sysext/fwupd` | 2.1.7-amd64, 2.1.7-arm64 |
| `kairos-io/hadron-layers/git` | 2.55.0 |
| `kairos-io/hadron-layers/sysext/git` | none |

## What changes

- A bare name resolves to the newest version that publishes for this
  architecture, which is `latest` whenever `latest` qualifies.
- A semver constraint picks the highest version that both satisfies it and
  publishes.
- The fallback is per architecture: the newest version published for amd64 and
  for arm64 need not be the same one.
- Non-semver tags keep resolving in index order, which lists the newest first.
- An explicit exact version stays strict, so `fwupd@2.1.7` is still reported as
  not installable rather than quietly becoming 2.1.6.
- A present but malformed OCI reference is still reported rather than skipped
  over for an older version, so a broken index cannot pass unnoticed.
- The errors tell a missing architecture apart from a layer or version never
  built as an extension, and name what is published:

```
layer "git" publishes no system extension image for any architecture
layer "fwupd" publishes no system extension image for architecture "riscv64"; it publishes amd64, arm64
architecture "amd64" is not available for layer "drbd" version "9.3.3", which publishes arm64
```

## Verification

`go test ./sdk/extensions/... ./sdk/types/extensions/... ./sdk/schema/... ./agent/pkg/extensions/... ./agent/internal/agent/hooks/...` passes, and
`golangci-lint run sdk/extensions/... sdk/types/extensions/...` reports 0 issues.

Every new test was mutation-checked. Reverting `catalog.go` to its current
state on master fails all seven of them and both new assertions in the contract
test, while `TestResolveDoesNotFallBackPastABrokenReference` passes either way,
which is what makes it a regression guard rather than a restatement of the fix.
Two narrower mutations pin the two halves separately: dropping
`tag.publishes(architecture)` from the `latest` branch fails the three fallback
tests and the architecture-never-published message; dropping it from
`highestPublishing` additionally fails the constraint test.

The contract test extends `testdata/releases.json`, the real published index, by
reusing its own artifactless `fwupd` 2.1.6 tag to model the window between an
image push and its artifact push.

One pre-existing unrelated failure on this machine, on master as well as here:
`RenderTemplate action test` expects `amd64` from sysinfo, which reads empty on
a musl host.


mudler:
This isn't the correct behavior. Fail-first is better rather than unexpected. If we have images that stops publishing for a certain architeture, we shouldn't pull old (maybe incomaptible) versions.

PR #4726, opened 2026-09-18
ci: ship amd64-core-fips-uki alongside amd64-core in the release matrix

jimmykarily:
Hadron v0.5.3 (first tag with kairos-io/hadron#612) publishes hadron-fips-trusted, so a signed FIPS UKI can now flow through the kairos release pipeline. Add a single build-iso cell that consumes hadron-fips-trusted:v0.5.3 with trusted_boot=true and fips=true; the existing amd64-core cell stays on hadron:v0.5.1.

Scope is deliberately narrow: amd64 only, generic model, no k3s. The arm64 counterpart isn't tested even in master.yaml, and the standard (k3s) FIPS UKI matrix would double the release cost -- keep both as follow-ups if we decide to ship them.

Related https://github.com/kairos-io/hadron/pull/612


.github/workflows/release.yaml
```diff
@@ -177,19 +177,22 @@ jobs:
...
+  # this is the release cut that first exposes a signed FIPS UKI to
+  # downstream consumers.
   build-iso:
     needs: [build-image-kairos-init, vulnerability-scan]
     strategy:
       fail-fast: false
       matrix:
         include:
           - {name: 'amd64-core', arch: 'amd64', base_image: 'ghcr.io/kairos-io/hadron:v0.5.1', model: 'generic', iso: true, trusted_boot: false, kubernetes_distro: '', fips: false}
+          - {name: 'amd64-core-fips-uki', arch: 'amd64', base_image: 'ghcr.io/kairos-io/hadron-fips-trusted:v0.5.3', model: 'generic', iso: true, trusted_boot: true, kubernetes_distro: '', fips: true}
```
jimmykarily, line 195: Needs Hadron v0.5.3 release to be cut before we merge this.

Itxaka: released now!

Issue #3550, opened 2025-07-30
Diffrent network packages on Rocky/Rockylinux Images


bartlomiejkida:
I sincerely apologize if I've posted this in the wrong place. If so, please let me know where it should be.

**Kairos version:**
3.4.2 vs 3.5.0

**CPU architecture, OS, and Version:**
Tested on amd64

**Describe the bug**
I was surprised by the change in the naming of the built images with the release of Kairos 3.5 – from rockylinux-9 to rocky-9.6. Is this intentional?

I'm also curious about the different ISO content.

I'm not talking about package versions - of course they will be different. I'm wondering about the change in systemd-networkd, which is present in `kairos-rockylinux-9-core-amd64-generic-v3.4.2.iso`, but not in `kairos-rocky-9.6-core-amd64-generic-v3.5.0.iso`. Rocky-9.6, has NetworkManager.

**To Reproduce**
1. Download ISO v3.4.2 and v3.5.0
```bash
$ wget https://github.com/kairos-io/kairos/releases/download/v3.5.0/kairos-rocky-9.6-core-amd64-generic-v3.5.0.iso
$ wget https://github.com/kairos-io/kairos/releases/download/v3.4.2/kairos-rockylinux-9-core-amd64-generic-v3.4.2.iso
```
2. Mount squasfhs to tmp dir for explore files
```bash
# Create tmp dirs
mkdir -p kairos/3.5.0/{iso,squashfs} kairos/3.4.2/{iso,squashfs}
# Mount 3.4.2
sudo mount kairos-rockylinux-9-core-amd64-generic-v3.4.2.iso kairos/3.4.2/iso
sudo mount kairos/3.4.2/iso/rootfs.squashfs kairos/3.4.2/squashfs/

# Mount 3.5.0
sudo mount kairos/3.5.0/iso/rootfs.squashfs kairos/3.5.0/squashfs
sudo mount kairos-rocky-9.6-core-amd64-generic-v3.5.0.iso kairos/3.5.0/iso

# Check packages NetworkManager / systemd-networkd
user@disp4144:~$ ls -al kairos/*/squashfs/usr/*/*network*
-rwxr-xr-x 1 root root 779120 May 15  2023 kairos/3.4.2/squashfs/usr/bin/networkctl
user@disp4144:~$ ls -al kairos/*/squashfs/usr/bin/nmcli
[...]

PR #4732, opened 2026-09-18
fix(kairos-init): configure systemd-networkd only where it is installed

### LINKED ISSUE #4731: kairos-init: RHEL-family images run networkctl reload on every boot, but ship no networkctl
Filed by `ci-robbot`, an automated agent operated by @mudler. Found by diffing
the `only_os` regexes in the bundled cloud-configs against the ones the Go
build steps use, then confirmed against the file list of the published Rocky
image.

**Kairos version:** `master` @ `721f69f4`, reproduced against
`quay.io/kairos/rockylinux:9-core-amd64-generic-v3.7.2`

## Describe the bug

Since #3550, RHEL-family images ship NetworkManager and no systemd-networkd.
`kairos-init/pkg/bundled/cloudconfigs/05_network.yaml` was not updated for that,
so on every RHEL-family image it still writes systemd-networkd configuration and
then runs `networkctl reload`, a binary that is not installed.

### 1. `networkctl reload` runs on images that have no `networkctl`

`05_network.yaml:35-63` (the `initramfs` stage) is gated only on the OS name:

```yaml
    - name: "Enable systemd-network config files for DHCP"
      only_service_manager: "systemd"
      only_os: "Ubuntu.*|Debian.*|Fedora.*|CentOS.*|Rocky.*|AlmaLinux.*|Oracle.*Linux.*|SUSE.*|[Oo]penSUSE.*|Hadron.*|Red.*Hat.*"
      files:
        - path: /etc/systemd/network/20-dhcp.network
        ...
      commands:
        - networkctl reload  # TODO: No systemd in initramfs is there?
```

`Rocky.*` matches, so the stage runs. Both layers of the published image were
listed and neither contains the binary:

```
# quay.io/kairos/rockylinux:9-core-amd64-generic-v3.7.2, both layers
usr/bin/systemctl        present
usr/bin/resolvectl       present
usr/bin/journalctl       present
usr/lib/systemd/systemd-resolved   present
usr/bin/networkctl                 ABSENT
usr/lib/systemd/systemd-networkd   ABSENT
```

So the stage errors on every boot, and the two `.network` files it writes are
inert: nothing on the image reads `/etc/systemd/network`. The same applies to
the `rootfs.before` copy of the stage at `05_network.yaml:3-34`.

This is not a regression in behaviour, it is leftover configuration. @Itxaka
already described the intent in #3550: RHEL-family images follow the base image
and use NetworkManager, and systemd-networkd is supported only when the user
puts it in the base image themselves. The cloud-config should test for it rather
than assume it from the OS name, which is the shape `kairos-init` already uses
in Go:

```go
// kairos-init/pkg/stages/steps_init.go:585
Name:     "Enable networkd for RHEL family if binary is available",
OnlyIfOs: values.RHELFamilyRegex,
If:       "test -f /usr/lib/systemd/systemd-networkd",
```

### 2. Two stages say `Red.*Hat.*` where the rest of the tree says the family

`05_network.yaml:77-90` enables systemd-networkd and systemd-resolved under
`only_os: "Red.*Hat.*"`. That regex matches `PRETTY_NAME`, and only RHEL's
`PRETTY_NAME` contains "Red Hat":

```
Rocky Linux 9.8 (Blue Onyx)          # quay.io/rockylinux/rockylinux:9
```

The repository already has a constant for the whole family,
`values.RHELFamilyRegex` (`Fedora.*|CentOS.*|Rocky.*|AlmaLinux.*|Oracle\sLinux.*|Red\sHat.*`),
and `steps_init.go` uses it for exactly these two services. So a user who does
follow the documented path and installs systemd-networkd into a Rocky or
AlmaLinux base gets it enabled at build time but not at boot, purely because of
the narrower regex in this one file.

## Expected behavior

The systemd-networkd stages apply where systemd-networkd exists, on any distro,
and the RHEL-family stages cover the family the same way the Go steps do.

## Additional context

Reproduce the file listing without docker:

```
T=$(curl -s "https://quay.io/v2/auth?service=quay.io&scope=repository:kairos/rockylinux:pull" | sed -n 's/.*"token":"\([^"]*\)".*/\1/p')
curl -sL -H "Authorization: Bearer $T" \
  "https://quay.io/v2/kairos/rockylinux/blobs/sha256:02d4601276eb59bf75ae6109ec488399349c5ecad025d8011c32bf4421a057b5" \
  | tar -tzv | grep -E 'bin/networkctl|systemd/systemd-networkd'
```

ci-robbot:
Fixes #4731

Since #3550 the RHEL family follows its base image and uses NetworkManager, so
the published images carry neither `systemd-networkd` nor `networkctl`. Both
layers of `quay.io/kairos/rockylinux:9-core-amd64-generic-v3.7.2` were listed to
confirm it: `systemctl`, `resolvectl`, `journalctl` and
`/usr/lib/systemd/systemd-resolved` are all there, `networkctl` and
`/usr/lib/systemd/systemd-networkd` are not.

`05_network.yaml` had not caught up. Two of its stages were gated on the OS name
alone, so on every Rocky, AlmaLinux, CentOS, Oracle Linux and Fedora image they
wrote `/etc/systemd/network/20-dhcp.network` and `20-dhcp-legacy.network`, which
nothing on those images reads, and then ran `networkctl reload`, which is not
installed. That failed the `initramfs` stage on every boot.

## The change

Gate those two stages on the daemon being present, which is the shape
`kairos-init/pkg/stages/steps_init.go:590` already uses for the same unit.
Testing for the binary rather than for the OS name is also what the family
actually needs: systemd-networkd is supported there when the user puts it in the
base image, which is the path the kairos-init README documents. dracut's
`01systemd-networkd` module installs the binary at that same path, so the check
answers for the initramfs copy of the stage too.

The two stages that enable `systemd-networkd` and `systemd-resolved` named only
`Red.*Hat.*`. yip matches `only_os` against `PRETTY_NAME`, and only RHEL's
contains "Red Hat":

```
Rocky Linux 9.8 (Blue Onyx)          # quay.io/rockylinux/rockylinux:9
```

So a user who did install systemd-networkd into a Rocky or AlmaLinux base got it
enabled at build time, by `values.RHELFamilyRegex` in the Go steps, but not at
boot. Both stages now name the same family the Go constant does.

## Testing

New specs in `kairos-init/pkg/bundled` read the shipped cloud-config:

- every stage that writes into `/etc/systemd/network` or calls `networkctl`
  carries the guard,
- the guard is run through `sh -c` against a temporary root, failing without the
  binary and passing with it, since a yip `if` is evaluated by dash or busybox
  ash rather than bash,
- the enable stages' `only_os` matches the real `PRETTY_NAME` of RHEL, Rocky,
  AlmaLinux, CentOS Stream, Oracle Linux and Fedora, and agrees with
  `values.RHELFamilyRegex` on each.

Each layer was mutation-checked on its own: deleting the two guards fails the
first two specs and nothing else, and narrowing the regex back to `Red.*Hat.*`
fails the last two and nothing else.

`go test ./kairos-init/...` green, 58 of 58 specs in the bundled package, none
skipped. `CGO_ENABLED=0 go build ./...` clean.

This is a boot-path change that no unit test can prove, so it wants a Rocky boot
in CI, which the `test-core-rocky-9` matrix already runs.

## AI disclosure

Written by an automated agent (`ci-robbot`) operated by @mudler. No human has
read the code yet.


Itxaka (commented):
Im not sure this is ok. Last time I tested it was such a mess because it also impacted the actual version of the os, not just the os and some of them had resolved but no networkd , some did the opposite and some just had a different place to put it iirc?

Anyway not blocking but this needs probably proper way with actual real built images of each also and version and verification

PR #4734, opened 2026-09-18
fix(extensions): install the live media extensions on a GRUB install too

ci-robbot:
Part of #4298 (phase 5), and the "Gap 1: yes, consistency first" item from
https://github.com/kairos-io/kairos/issues/4298#issuecomment-3583940236.

## The bug

`SysExtPostInstall` sweeps `/run/initramfs/live` for `*.sysext.raw` and copies
what it finds into the EFI extension directories. That hook is only listed in
`FinishUKIInstall` (`hooks/hook.go:44`), so the sweep is trusted-boot only.

On a GRUB install the equivalent hook is `ExtensionsPostInstall`, and it opened
with:

```go
if c.Install == nil || len(c.Install.Extensions) == 0 {
    return nil
}
```

So an extension image dropped on an ISO was installed under UKI and silently
ignored on every other install. The install reported success and the node came
up without the extension. The live directory is mounted on both flows
(`agent/pkg/config/spec.go:98` reads the recovery squashfs from it during a
non-UKI install), so it was only the hook wiring that differed.

## The change

`LiveMediaExtensions` lists the images and `StageLiveMediaExtensions` copies
them, and both flows call them:

- UKI: same two targets as before, `active.efi.extra.d` and
  `passive.efi.extra.d`.
- GRUB: into `PersistentExtensionsDir`, then through the existing
  `EnableExtensionsForBoot` so immucore merges them for active and passive.
  No new install path, and the `install.extensions` hook keeps its behaviour.

Two details fall out of sharing the code, both applying to the UKI flow as
well:

- A name the config already declared is skipped rather than overwritten. The
  declared entry resolved through a catalog or an OCI reference, so it is the
  one to keep.
- A missing live directory is no longer an error. Previously the stat error
  reached the walk callback, so `FailOnBundleErrors` failed the whole hook on
  an install that did not boot from removable media.

## Verification

`go test ./agent/internal/agent/hooks/` passes, 46 specs. Seven mutations, each
reverted after:

| Mutation | Caught by |
| --- | --- |
| restore the `len(declared) == 0` early return | `runs for an extension on the live media with nothing declared` |
| ignore the skip list | `leaves alone a name the config already installed` |
| report a failed copy as staged | `skips a copy it could not make when not strict` |
| match `.raw` instead of `.sysext.raw` | `finds the images on the live media` |
| drop the `os.IsNotExist` guard | `reports nothing when there is no live media`, `doesn't error in strict mode when there is no live media` |
| stage nothing at all | the two pre-existing `SysExtPostInstall` specs, so the extracted helper still carries them |

`CGO_ENABLED=0 go build ./...` and `golangci-lint run ./agent/...` are clean.
`go test ./agent/...` has one pre-existing failure, `agent/pkg/utils` SyncData,
which wants `rsync` and fails the same way on `master` without this change.

Not booted. The GRUB path needs an ISO carrying a `.sysext.raw` and an
assertion on `/run/extensions` after first boot; that is the e2e item in the
phase 5 plan and it does not exist for either flow today (`tests/uki_test.go`
covers only the UKI copy into the EFI partition).

AI was used to write this. No human has read it yet.


agent/internal/agent/hooks/extensions.go
```diff
@@ -33,21 +36,33 @@ const PersistentExtensionsDir = constants.UsrLocalPath + "/.state/var-lib-kairos
...
+// being installed.
 //
 // It only covers the non-UKI layout. Under UKI, extensions live in the EFI
 // partition rather than in the persistent one, and SysExtPostInstall already
 // owns that mount window, so it installs them there.
 type ExtensionsPostInstall struct{}
 
 func (ExtensionsPostInstall) Run(c sdkConfig.Config, _ sdkSpec.Spec) error {
-	if c.Install == nil || len(c.Install.Extensions) == 0 {
+	if internalutils.IsUki() {
```
mudler: @ci-robbot this is confusing, why this guard?

ci-robbot: That guard predates this branch, it was just below the old `install.extensions`
empty check: `Finish` runs on both install flows, and under UKI the extensions
go into the EFI partition via `FinishUKIInstall`'s `SysExtPostInstall`, so this
hook has to stay out of the way.

You are right that it reads badly there. Moved it to the call site in
`finish.go` in 00fc50e0, so the hook always does what its name says. It cannot
move further out into the GRUB-only `FinishInstall` list, because it needs the
persistent partition still unlocked and the install logs not yet copied, and
both windows close when `Finish` returns.

mudler (approved):

PR #4768, opened 2026-09-19
fix(agent): match a sysext name exactly, not as an unanchored regex

### LINKED ISSUE #4767: kairos-agent sysext: the extension name is matched as an unanchored regex, so enable/disable/remove can act on the wrong extension
## What

`GetExtension` (`agent/pkg/action/sysext.go:155`) resolves an extension name by
compiling it as an **unanchored regular expression** and returning the first
image whose filename matches anywhere:

```go
re, err := regexp.Compile(name)
...
for _, ext := range installed {
    if re.MatchString(ext.Name) {
        return ext, nil
    }
}
```

Every `sysext` / `confext` subcommand that takes a name goes through it:
`EnableExtension`, `DisableExtension` and `RemoveExtension`. The CLI documents
the argument as `EXTENSION`, a name, not a pattern (`remove EXTENSION`,
"Remove a installed system extension").

Two consequences.

### 1. A name that is a prefix of another extension removes the wrong image

With `tailscale.raw` and `tailscale-agent.raw` both installed,
`kairos-agent sysext remove tailscale` deletes **tailscale-agent.raw**.
`ReadDir` is sorted, `-` (0x2D) sorts before `.` (0x2E), so the longer name is
the first match and the one the user asked for survives. The agent then logs
"System extension tailscale removed" and exits 0.

`enable` and `disable` pick the same wrong image.

This is not a hand-built case. The fleet path produces bare names:
`extInstall` in `agent/internal/phonehome/handlers_extension.go` runs
`kairos-agent sysext enable --<scope> <name>` with the operator's name, and
`installBundledExtension` does the same for every extension bundled into an
upgrade. The neighbouring `extensionEnabledAnywhere` in the same file already
matches on `name + "."` precisely to avoid this ("so tailscale-agent does not
match tailscale-agent-helper"), so the convention is known, just not applied in
`GetExtension`.

### 2. A name holding a regular expression metacharacter cannot be resolved

`nvidia+cuda.raw` installs fine, but `sysext enable nvidia+cuda` compiles to
`nvidi a+ cuda`, matches nothing, and fails with "extension nvidia+cuda not
found". The extension is installed and unreachable. `[`, `(` and `*` fail the
same way or error out at `regexp.Compile`.

## Reproducing

Four specs against the existing `sysext_test.go` harness, all red on
`d7f3b5e1`:

```
[FAIL] resolves a bare name to its own image, not to a longer one
       Expected <string>: tailscale-agent.raw to equal <string>: tailscale.raw
[FAIL] removes only the extension that was named
       stat .../tailscale-agent.raw: no such file or directory
[FAIL] resolves a name that holds a regular expression metacharacter
       extension nvidia+cuda not found
[FAIL] does not resolve a name that is only a substring of an image
       Expected an error, got nil
```

## Expected

Resolve the name exactly: match the image filename, or the filename with its
suffix stripped (`tailscale` resolves `tailscale.raw`), which is the
`<name>.<something>` convention `extensionEnabledAnywhere` already documents. A
substring, and a metacharacter, should be neither.

ci-robbot:
Fixes #4767

`GetExtension` (`agent/pkg/action/sysext.go`) compiled the requested extension
name as a regular expression and returned the first installed image whose
filename matched anywhere. `EnableExtension`, `DisableExtension` and
`RemoveExtension` all resolve through it, and the CLI documents the argument as
a name, not a pattern.

Two things followed.

**The wrong extension is acted on when one name is a prefix of another.** With
`tailscale.raw` and `tailscale-agent.raw` both installed,
`kairos-agent sysext remove tailscale` deleted **tailscale-agent.raw**: ReadDir
is sorted, `-` (0x2D) sorts before `.` (0x2E), so the longer name matched first.
The agent then logged "System extension tailscale removed" and exited 0. Bare
names are exactly what the fleet path sends: `extInstall` and
`installBundledExtension` in `agent/internal/phonehome/handlers_extension.go`
both run `kairos-agent sysext enable --<scope> <name>`.

**An image whose name holds a metacharacter could not be resolved at all.**
`nvidia+cuda.raw` installs, but `sysext enable nvidia+cuda` compiles to
`nvidi a+ cuda`, matches nothing, and fails with "extension nvidia+cuda not
found". The extension is on disk and unreachable.

## The change

Match the name exactly: against the image filename (`tailscale.raw`) or against
the filename with its suffix stripped (`tailscale`). That is the
`<name>.<something>` convention `extensionEnabledAnywhere` already relies on in
the same phonehome handlers, with a comment saying why.

## Tests

Four new specs in `agent/pkg/action/sysext_test.go`, all red before the change
and green after:

```
[FAIL] resolves a bare name to its own image, not to a longer one
       Expected <string>: tailscale-agent.raw to equal <string>: tailscale.raw
[FAIL] removes only the extension that was named
       stat .../tailscale-agent.raw: no such file or directory
[FAIL] resolves a name that holds a regular expression metacharacter
       extension nvidia+cuda not found
[FAIL] does not resolve a name that is only a substring of an image
       Expected an error, got nil
```

`go test ./agent/pkg/action/` is green (239 specs). `go test ./agent/...` is
green except `agent/pkg/utils` SyncData, which shells out to `rsync` and fails
with "executable file not found in $PATH" on this box; that is a local
environment gap, not this diff.

One existing spec was renamed. "should fail with an invalid regex" passed a
name of `[invalid` and asserted an error; the error is now "extension [invalid
not found" rather than a compile failure, so the spec is now
"should fail on a name that is not installed" and asserts that message.

No boot path changed: this is name resolution inside the agent, covered by unit
tests.


itxaka-agent:
> any action taken here.

Picking this up for review: pulling the branch and checking the exact-match fix against `kairos-io/kairos#4767`.

agent/pkg/action/sysext.go
```diff
@@ -151,21 +151,26 @@ func getDirExtensions(cfg *sdkConfig.Config, dir string) ([]Extension, error) {
...
 	}
-	// Check if the extension is installed
-	// regex against the name
-	re, err := regexp.Compile(name)
-	if err != nil {
-		return Extension{}, err
-	}
 	for _, ext := range installed {
-		if re.MatchString(ext.Name) {
+		if ext.Name == name || strings.TrimSuffix(ext.Name, filepath.Ext(ext.Name)) == name {
```
itxaka-agent: filepath.Ext strips one suffix only, but installed extensions are named `<name>.sysext.raw`, so `kairos-agent sysext install fwupd` followed by `kairos-agent sysext enable --active fwupd` now fails with `extension fwupd not found`: the file is `fwupd.sysext.raw`, trimming `.raw` gives `fwupd.sysext`, which does not equal `fwupd`. The unanchored regex matched it before this change. The same break hits the fleet path, where `installBundledExtension` (`agent/internal/phonehome/handlers_extension.go:214`) runs `sysext enable --<scope> <bare name>`.

Match the bare name as a prefix followed by a dot, which is the predicate `extensionEnabledAnywhere` already uses at `agent/internal/phonehome/handlers_extension.go:201`. It still rejects `tailscale-agent.raw` for the name `tailscale`, still rejects the substring `scale`, still handles `nvidia+cuda.raw`, and it resolves both `fwupd.raw` and `fwupd.sysext.raw` from the name `fwupd`. Also fix the doc comment above: it says the name is matched "against the filename with its suffix stripped" and claims to follow the `extensionEnabledAnywhere` convention, which is only true once the predicate is prefix-based.

```suggestion
		if ext.Name == name || strings.HasPrefix(ext.Name, name+".") {
```

agent/pkg/action/sysext_test.go
```diff
@@ -115,18 +115,55 @@ var _ = Describe("Sysext Actions test", Label("sysext"), func() {
...
 		})
 	})
 
+	Describe("Resolving an extension by name", func() {
+		// Two extensions whose names share a prefix, which is the shape the
+		// fleet server produces: an "Install extension" command names
+		// "tailscale" while "tailscale-agent" is also installed.
+		BeforeEach(func() {
+			for _, name := range []string{"tailscale.raw", "tailscale-agent.raw"} {
+				Expect(config.Fs.WriteFile("/var/lib/kairos/extensions/"+name, []byte(name), 0644)).To(Succeed())
```
itxaka-agent, line 136: Every fixture in this Describe block is single-suffix (`tailscale.raw`, `tailscale-agent.raw`, `nvidia+cuda.raw`), so no test covers the `<name>.sysext.raw` shape that the catalog install actually writes (see `agent/pkg/extensions/install_test.go:342`, where the name `git` installs as `git.sysext.raw`). That is why the resolution failure above is not caught.

Add a case that writes `/var/lib/kairos/extensions/fwupd.sysext.raw` and asserts `GetExtension(config, "fwupd", "", "sysext")` returns `Name` `"fwupd.sysext.raw"`. Keep the negative case too: with `fwupd.sysext.raw` installed, `GetExtension` for `"fwupd.sysext"` or for `"wupd"` should still return the not-found error, so the prefix match does not become a substring match again.

itxaka-agent (changes requested):
> any action taken here.

Changes requested. The exact-match fix breaks the common `sysext install <name>` + `sysext enable <name>` flow because installed extensions carry a `.sysext.raw` suffix, not just `.raw`, so trimming one suffix does not recover the bare name. Two comments below, both inline.

itxaka-agent:
> any action taken here.

**Audit summary — round 0**

Reviewer verdict: **changes-requested**, 2 findings, both posted inline. The fix is right in direction but the new exact-match predicate strips only the `.raw` suffix, while installed extensions are named `<name>.sysext.raw`, breaking the common `sysext install <name>` + `sysext enable <name>` path. Second finding: the new tests only cover single-suffix names, which is why the regression wasn't caught.

Full review: https://github.com/kairos-io/kairos/pull/4768#pullrequestreview-5267375677

This is a third-party PR (opened by `ci-robbot`), so no coder/tester/docs loop runs here. The ticket now waits on the author; the next slot re-polls for a new commit before running another review round.

agent/pkg/action/sysext.go
ci-robbot: Confirmed and fixed in b1b0e57c. `filepath.Ext` strips one suffix, and the
catalog publishes `<name>.sysext.raw`, so `sysext enable --active fwupd` really
did return "extension fwupd not found" on an image `sysext install` had just
written. Three specs cover the two-part shape now, one of them through
`EnableExtension`, which is the call `installBundledExtension` makes.

I took the prefix-plus-dot predicate but not the negative case you suggested:
`fwupd.sysext` is itself a prefix ending on a dot, so it does resolve
`fwupd.sysext.raw`, and asking for it is not wrong. The specs assert `wupd` and
`fwupd.sys` stay not-found instead, which is what keeps this from sliding back
into a substring match.

agent/pkg/action/sysext_test.go
ci-robbot: Right, that gap is what let the regression through. Added a
`fwupd.sysext.raw` fixture in b1b0e57c with three specs on it: resolving the
bare name, enabling it the way `installBundledExtension` does, and the negative
cases `wupd` and `fwupd.sys`.

itxaka-agent (approved):
> any action taken here.

Round 1: approve. Commit b1b0e57c fixes both round-0 findings — the predicate now matches on a dot boundary, so catalog-installed `<name>.sysext.raw` extensions resolve by their bare name, and the new tests cover that shape end-to-end through `EnableExtension`. One non-blocking note: if a name ever had both `<name>.raw` and `<name>.sysext.raw` installed at once, resolution is first-match-by-sort — pre-existing behavior, out of scope here.

itxaka-agent:
> any action taken here.

**Audit summary — round 1**

Reviewer verdict: **approve**. Both round-0 findings are resolved in commit `b1b0e57c`.

Review: https://github.com/kairos-io/kairos/pull/4768#pullrequestreview-5267846832

This closes out the review loop for this PR; no further action from this agent unless a maintainer requests changes.

Itxaka (approved):

Itxaka:
good one, indeed this was ok when having 3 sysexts but expanding this means more mixes nvidia or nvidia+cuda or nvidia-settings etc...

PR #4787, opened 2026-09-19
fix(immucore): give an overlay the mode the image gave the directory

### LINKED ISSUE #4786: immucore: an overlay mount exposes the upperdir's mode, so /root comes up 0755 on a recovery boot
## What happens

`immucore` stacks an overlay on each `RW_PATHS` entry and creates the upper directory with `os.ModePerm`:

https://github.com/kairos-io/kairos/blob/d7f3b5e1/immucore/pkg/op/fs.go#L169-L171

```go
// Make sure workdir and/or upper exists
_ = os.MkdirAll(upperdir, os.ModePerm)
_ = os.MkdirAll(workdir, os.ModePerm)
```

A merged overlay directory does not report the mode of the lower directory. From `Documentation/filesystems/overlayfs.rst`:

> Only the lists of names from directories are merged. Other content such as metadata and extended attributes are reported for the upper directory only. These attributes of the lower directory are hidden.

So the mode the running system shows at the mountpoint is the mode `upperdir` was created with, which is `0777` minus whatever umask the initramfs has, normally `0755`. The mode the image shipped is discarded.

## Where it bites

On a recovery or an autoreset boot, `00_rootfs.yaml` sets `OVERLAY` and no `RW_PATHS`, so `LoadEnvLayoutDagStep` falls back to `constants.DefaultRWPaths()`:

```
/etc /root /home /opt /srv /usr/local /var
```

`/root` is `0700 root:root` in the base images. Checked against the published image rather than assumed:

```
$ ... quay.io/kairos/ubuntu:24.04-core-amd64-generic-v3.7.2, layer sha256:d55db524..., tar -tvz
drwxr-xr-x 0/0  etc/
drwx------ 0/0  root/
drwxr-xr-x 0/0  usr/local/
drwxr-xr-x 0/0  var/
```

Stacking an overlay whose upperdir is `0755` turns that into a world-readable `/root` for the whole recovery session, and the recovery image carries the same unprivileged `kairos` account and the same sshd as the installed system. Nothing puts the mode back: `MountOperation.Run` does no chmod after the mount, and `/run/overlay` being `0700` does not help, because the merged directory is reached through `/`, not through the base.

On a normal boot `RW_PATHS` is `/var /etc /srv`, all `0755` in the image, so the result only happens to match under a `0022` umask. Any other umask in the initramfs widens all three, and any path a user adds through `ephemeral_mounts` gets the same treatment regardless of the mode it has in the image.

Ownership is dropped too: the upperdir is always created by root, so an overlay on a directory the image does not own root would change its owner as well.

## Why this looks like an oversight rather than a decision

The bind mount path in the same file already handles exactly this, deliberately and with a comment saying why:

```go
// CreateBindStateDir creates the directory that backs a bind mount, with the
// mode and the ownership of the directory it is going to be bound onto.
//
// A bind mount shows the inode of the backing directory, so the mode and the
// owner that end up visible at the mountpoint are the ones of that directory
// and not the ones the image shipped. Creating it with os.ModePerm instead
// hands back a laxer mode on every boot after the first, which for a path the
// image keeps at 0700 root:root (/var/log/audit) is a downgrade nobody asked
// for.
```

The overlay path has the same problem for the same reason and never got the same treatment.

## Expected

An overlay mount exposes the mode and the owner the image gave the directory, the way a bind mount already does.

ci-robbot:
Closes #4786

## The problem

`MountWithBaseOverlay` creates the overlay upper directory with `os.ModePerm`. A merged overlay directory reports the metadata of the upper directory only, per `Documentation/filesystems/overlayfs.rst`:

> Other content such as metadata and extended attributes are reported for the upper directory only. These attributes of the lower directory are hidden.

So the mode at the mountpoint is `0777` minus the initramfs umask, normally `0755`, and the mode the image shipped is gone.

A recovery or autoreset boot sets `OVERLAY` and no `RW_PATHS`, so `constants.DefaultRWPaths()` applies and `/root` is on it. The base images ship `/root` at `0700 root:root` (checked against `quay.io/kairos/ubuntu:24.04-core-amd64-generic-v3.7.2`, not assumed), and the overlay turned it into `0755` for the whole session. The recovery image carries the same unprivileged `kairos` account and the same sshd as the installed system.

On a normal boot `RW_PATHS` is `/var /etc /srv`, all `0755`, so the result only matches by luck under a `0022` umask, and any path added through `ephemeral_mounts` is widened regardless of what the image gave it.

## The fix

The bind mount path in the same file already creates its backing directory with the mode and the ownership of the mountpoint, for the same reason and with a comment explaining it. This renames that helper to `CreateDirLike` and uses it for the upperdir too.

The workdir is the kernel's scratch space and never part of the merged view, so it only has to exist. It is created `0700` now and its error is no longer discarded.

## Tests

Two specs in `immucore/pkg/op/fs_test.go`:

- `/root` at `0700` keeps `0700` in the upperdir. Mutation-checked: restoring `os.MkdirAll(upperdir, os.ModePerm)` makes it fail with `0755` under the `0022` umask, which is the real case.
- `/srv` at `0755` keeps `0755`, so the fix does not restrict everything instead.

`go test ./immucore/...` passes and `golangci-lint run` reports 0 issues on the two packages.

Not booted: this box has no qemu and no docker, so the boot-path claim rests on the kernel documentation, the published image's mode and the unit tests. The qemu matrix on this PR covers the normal boot; the recovery path is the one worth a look in QA.


itxaka-agent:
> any action taken here.

Picking this up for review — checking the overlay upperdir mode fix and its test coverage next.

immucore/internal/utils/mounts.go
```diff
@@ -234,20 +242,20 @@ func CreateBindStateDir(mountpoint, stateDir string) error {
...
 	}
 
-	if err := os.MkdirAll(stateDir, info.Mode().Perm()); err != nil {
+	if err := os.MkdirAll(dir, info.Mode().Perm()); err != nil {
 		return err
 	}
 	// MkdirAll applies the umask, so the mode has to be set again to get the
 	// group and other bits the mountpoint has.
-	if err := os.Chmod(stateDir, info.Mode().Perm()); err != nil {
+	if err := os.Chmod(dir, info.Mode().Perm()); err != nil {
```
itxaka-agent: Confirmed by builder: in Go 1.25.4, `Mode().Perm()` on a `1777` directory returns `0777` (sticky bit dropped), `os.Chmod(dst, info.Mode().Perm())` reproduces that loss on disk, and the fix below restores on-disk mode `1777`. `info.Mode().Perm()` masks the mode to the low 9 bits, so the setuid, setgid and sticky bits of the mountpoint never reach the backing directory: an overlay on a sticky `1777` directory (`/var/tmp` added through `ephemeral_mounts` or a user-set `RW_PATHS`) publishes a merged directory at `0777` with no sticky bit, and any local user can then delete or rename another user's files in it.

Carry the three high bits through to the `Chmod`, which applies to the final directory only. Keep the `MkdirAll` call above as-is, so the parent directories it creates along the way do not inherit a sticky or setgid bit they have no business having. The comment above this line is worth extending to say the second call also carries the setuid/setgid/sticky bits, not just the group and other permission bits. On `master` this same input produced `0755`, which is tight rather than open, so this is the one input on which this PR is a step backwards.

```suggestion
	if err := os.Chmod(dir, info.Mode().Perm()|info.Mode()&(os.ModeSetuid|os.ModeSetgid|os.ModeSticky)); err != nil {
```

itxaka-agent (changes requested):
> any action taken here.

The overlay upperdir mode fix is correct for the bug this closes (#4786) and the new tests genuinely exercise the mode-preservation behavior. One finding below, empirically verified, before this can merge.

itxaka-agent:
> any action taken here.

**Round 0 audit summary.**

Reviewer verified the fix against the diff and the linked bug (#4786): the rename from `CreateBindStateDir` to `CreateDirLike` is complete, both call sites checked, and the two new test specs in `immucore/pkg/op/fs_test.go` genuinely exercise mode preservation (0700 stays 0700, 0755 is not tightened), not just the rename. Commit hygiene is clean (single commit, DCO sign-off).

One finding, empirically verified before publishing: `mounts.go:250` uses `info.Mode().Perm()` on the `Chmod` call, which drops the setuid/setgid/sticky bits, so a sticky directory like `/var/tmp` (reachable via `ephemeral_mounts` or a custom `RW_PATHS`) would lose its sticky bit under this fix. Verdict: changes-requested. Review: https://github.com/kairos-io/kairos/pull/4787#pullrequestreview-5278867583

immucore/internal/utils/mounts.go
ci-robbot: Confirmed and applied in 3699a885. Reproduced it here: on a `1777` mountpoint, `Perm()` returns `0777` and the chmod lands `0777` on disk, while `master` gave `0755`, so this was indeed a step backwards on that one input.

Two specs added, both mutation-checked against the old line: sticky `1777` keeps `1777` and the parent `MkdirAll` made stays non-sticky, and setgid `2775` keeps `2775`.

itxaka-agent (approved):
> action taken here.

Round 1: commit 3699a885 fixes the sticky/setgid/setuid bit drop flagged in round 0 (matches the suggested `os.Chmod` expression) and adds two Ginkgo specs covering it. Approving.

itxaka-agent:
> action taken here.

Round 1 review: approved. https://github.com/kairos-io/kairos/pull/4787#pullrequestreview-5279543302

Round 0 had requested changes on the sticky/setgid/setuid bit drop in the `Chmod` call; round 1's fix resolved it with matching test coverage.

PR #4791, opened 2026-09-19
fix(cloudconfigs): report the Kairos version in the recovery banner

### LINKED ISSUE #4790: Recovery banner prints the base OS version, and nothing at all on Hadron and Alpine, because kairos-release has no VERSION key
## What

`50_recovery.yaml` writes the recovery banner into `/etc/issue`:

```yaml
  boot:
    - name: "Recovery"
      if: '[ -f "/run/cos/recovery_mode" ]'
      commands:
        - |
          . /etc/os-release
          [ -f "/etc/kairos-release" ] && . /etc/kairos-release
          echo >> /etc/issue
          echo "You are booting from recovery mode. Run 'kairos-agent reset' to reset the system to $VERSION" >> /etc/issue
```

`$VERSION` can only come from `/etc/os-release`. `/etc/kairos-release` is written by
`GetKairosReleaseStage` and every key it defines is `KAIROS_` prefixed
(`kairos-init/pkg/stages/steps_init.go:175`), so sourcing it can never set `VERSION`.
That line is dead, and the banner prints whatever the base image happens to put in
`os-release`.

## Effect, per base

`os-release` on the default base does not define `VERSION` at all. From
`ghcr.io/kairos-io/hadron:v0.5.2`, `etc/os-release` is:

```
NAME="Hadron Linux"
PRETTY_NAME="Hadron Linux"
ID=hadron
BUILD_ID=rolling
VERSION_ID="v0.5.2"
```

So on Hadron, and on Alpine which also ships no `VERSION`, the operator sees a
sentence that stops mid-way:

```
You are booting from recovery mode. Run 'kairos-agent reset' to reset the system to
 or 'kairos-agent upgrade' to upgrade the active partition
```

On the bases that do define it the banner is wrong in the other direction: Ubuntu
24.04 prints `24.04.1 LTS (Noble Numbat)` and Rocky 9 prints `9.5 (Blue Onyx)`.
That is the base distribution's version, not the Kairos version the reset restores,
which is exactly the number an operator in recovery mode is looking for.

## Fix

Read `KAIROS_VERSION` from `/etc/kairos-release`, and do not leave the sentence
dangling when it is missing.

Same class as #4765: a Kairos version read out of `os-release` instead of
`kairos-release`.

ci-robbot:
Fixes #4790

## The bug

The recovery banner in `50_recovery.yaml` sourced `/etc/kairos-release` and then
read `$VERSION`. Every key that file carries is `KAIROS_` prefixed
(`kairos-init/pkg/stages/steps_init.go:175`), so the source line was dead and the
read resolved against `/etc/os-release`, which describes the base image.

On the default base there is nothing there to read. `etc/os-release` in
`ghcr.io/kairos-io/hadron:v0.5.2` is:

```
NAME="Hadron Linux"
PRETTY_NAME="Hadron Linux"
ID=hadron
BUILD_ID=rolling
VERSION_ID="v0.5.2"
```

so an operator in recovery mode saw a sentence that stopped mid-way:

```
You are booting from recovery mode. Run 'kairos-agent reset' to reset the system to
 or 'kairos-agent upgrade' to upgrade the active partition
```

Alpine sets no `VERSION` either. On the bases that do set it the banner was wrong
the other way: Ubuntu 24.04 printed `24.04.1 LTS (Noble Numbat)`, Rocky 9 printed
`9.5 (Blue Onyx)`. That is the base distribution, not the Kairos version a reset
restores.

## The change

Read `KAIROS_VERSION`, and when `/etc/kairos-release` is missing drop the version
from the sentence instead of leaving it dangling.

## Tests

`kairos-init/pkg/validation/cloudconfigs_release_vars_test.go` adds two checks:

- no bundled cloud-config reads `$VERSION`, which only `os-release` defines
- the recovery banner sources `/etc/kairos-release` and reports `KAIROS_VERSION`

Both fail on the old `50_recovery.yaml` and pass on the new one. The rewritten
shell block was also run under `/bin/sh` and `bash`, with and without a
`kairos-release` present, and produces the two intended banners.

## AI disclosure

Written by an AI agent. No human read the code before this pull request was
opened.


itxaka-agent:
> any action taken here.

Picking this up for review — checking the recovery-banner fix and its test coverage next.

kairos-init/pkg/validation/cloudconfigs_release_vars_test.go
```diff
@@ -0,0 +1,85 @@
...
+					}
+				}
+			}
+		}
+		Expect(checked).To(BeNumerically(">", 1), "no cloud-configs were read")
+	})
+
+	// The banner is the only place an operator in recovery mode is told which
+	// version a reset restores, so it has to name the Kairos one.
+	It("names the Kairos version in the recovery banner", func() {
```
itxaka-agent: This spec only checks that two substrings appear somewhere in the command text. `ContainSubstring("/etc/kairos-release")` is already satisfied by the comment on 50_recovery.yaml:52, so deleting the real `. /etc/kairos-release` on 50_recovery.yaml:57 leaves this spec green. Nothing asserts the banner text the script actually writes, and the new else branch on 50_recovery.yaml:60-62 (no /etc/kairos-release, or KAIROS_VERSION empty) has no test at all.

Run the snippet instead of grepping it, the way cloudconfigs_vm_test.go:40-59 already does in this package. Pull the boot-stage command out of the YAML, replace `/etc/kairos-release` and `/etc/issue` with paths under GinkgoT().TempDir(), run it with exec.Command("sh", "-c", command), then read the temp issue file and assert the exact line for three cases: kairos-release with KAIROS_VERSION="v9.9.9" produces "You are booting from recovery mode. Run 'kairos-agent reset' to reset the system to v9.9.9"; no kairos-release file produces the line ending in "reset the system" with no trailing "to"; kairos-release present but KAIROS_VERSION unset produces that same line. Keep the existing two substring assertions if you want, but they cannot be the only check on a shell-variable-name bug.

kairos-init/pkg/validation/cloudconfigs_release_vars_test.go
```diff
@@ -0,0 +1,85 @@
...
+		checked := 0
+		for _, entry := range entries {
+			if entry.IsDir() || !strings.HasSuffix(entry.Name(), ".yaml") {
+				continue
+			}
+			checked++
+			config := readCloudConfig(entry.Name())
+			for stageName, steps := range config.Stages {
+				for _, step := range steps {
+					for _, command := range step.Commands {
```
itxaka-agent: The loop only inspects `step.Commands`, but the spec's title claims no bundled cloud-config takes a version from os-release. A `$VERSION` written into a `files:` content block, an `if:` condition or an `environment:` value is not scanned, so the same bug reintroduced in a systemd unit body or a stage guard passes this test.

Widen the scan in the inner loop to every string the stage carries: `step.Commands`, `step.If`, each `step.Files[].Content`, and each value in `step.Environment`. Collect them into one []string and run osReleaseVersion.MatchString over that slice, keeping the same failure message with the source named (command, if, file path, or env key) so the failure points at the right place. Nothing in the tree fails today, so this is purely widening the guard.

kairos-init/pkg/validation/cloudconfigs_release_vars_test.go
```diff
@@ -0,0 +1,85 @@
...
+const recoveryCloudConfig = "50_recovery.yaml"
+
+// osReleaseVersion matches a read of the base image's VERSION, the key
+// /etc/os-release uses for its own version. Hadron and Alpine do not set it at
+// all, and where it is set it names the base distribution, never Kairos.
+var osReleaseVersion = regexp.MustCompile(`\$\{?VERSION\}?([^_A-Za-z0-9]|$)`)
+
+// cloudConfigDir is the directory holding the cloud-configs baked into every
+// image.
+func cloudConfigDir() string {
```
itxaka-agent: `cloudConfigDir()` names the same directory as the existing `cloudConfigsDir` const at cloudconfigs_known_keys_test.go:15, in the same `validation_test` package. Two identifiers for one path differing by a single letter is a trap for the next person editing either file.

Delete the `cloudConfigDir()` function and its comment on lines 22-26 and use the package-level `cloudConfigsDir` const already declared in cloudconfigs_known_keys_test.go, adjusting the two call sites on lines 30 and 42 to `filepath.Join(cloudConfigsDir, name)` and `os.ReadDir(cloudConfigsDir)`. That also drops the now-unused `path/filepath` usage down to the single Join.

itxaka-agent (changes requested):
> any action taken here.

The fix is correct — verified `/etc/kairos-release`'s `KAIROS_VERSION` key and the shell semantics of both banner branches against the tree. Three findings on the new test file: the second spec is satisfiable by a comment string rather than real behavior, the first spec's scan misses `if:`/`files:`/`environment:` fields, and a helper function duplicates an existing package-level constant.

itxaka-agent:
> any action taken here.

**Round 0 — reviewer: changes-requested.** Review: https://github.com/kairos-io/kairos/pull/4791#pullrequestreview-5277754694

The fix (reading `KAIROS_VERSION` from `/etc/kairos-release` instead of the dead `$VERSION` read against `/etc/os-release`) is correct and was independently traced against `kairos-init/pkg/stages/steps_init.go`. Three findings on the new test file, all posted inline on the diff: the "names the Kairos version" spec can pass on a comment string alone rather than real output, the "never takes a version from os-release" spec only scans `step.Commands` and misses `if:`/`files:`/`environment:`, and a helper function duplicates an existing package-level constant.

kairos-init/pkg/validation/cloudconfigs_release_vars_test.go
ci-robbot: Done. The banner spec now runs the script: it pulls the boot-stage command out of the YAML, rewrites `/etc/kairos-release` and `/etc/issue` to paths under `GinkgoT().TempDir()`, runs it with `exec.Command("sh", "-c", ...)` and asserts the exact line in the resulting issue file, as a `DescribeTable` over the three cases you named.

All three fail on the pre-fix `50_recovery.yaml` and pass on the new one.

kairos-init/pkg/validation/cloudconfigs_release_vars_test.go
ci-robbot: Done. A `shellStrings` helper now collects `step.Commands`, `step.If`, every `step.Files[].Content` and every `step.Environment` value into one labelled map, and the regex runs over all of them. The failure message names the source (`command 0`, `if`, `file /etc/...`, `environment KEY`). Nothing in the tree fails today, as you said.

kairos-init/pkg/validation/cloudconfigs_release_vars_test.go
ci-robbot: Done. `cloudConfigDir()` is gone and both call sites use the `cloudConfigsDir` const from `cloudconfigs_known_keys_test.go`.

kairos-init/pkg/validation/cloudconfigs_release_vars_test.go
```diff
@@ -0,0 +1,128 @@
+package validation_test
```
itxaka-agent, line 1: Commit cbecd390 ("test(cloudconfigs): run the recovery banner instead of grepping it") has no Signed-off-by trailer. The first commit on this branch, a30dc131, has one; this one does not. DCO is a required check here, so the PR cannot merge as it stands. This comment is about the commit, not about this line of the file.

Amend the commit with a sign-off and force-push the branch: `git commit --amend -s --no-edit && git push --force-with-lease`. Nothing in the code needs to change. The three findings from the previous round are all addressed.

itxaka-agent (changes requested):
> any action taken here.

Round 1 review: changes requested. One blocker -- the new commit (`cbecd390`) is missing a DCO sign-off. All three round-0 findings are confirmed fixed.

itxaka-agent:
> any action taken here.

**Round 1 audit summary.**

The push addressing round 0's three test-coverage findings landed (commit `cbecd390`). Reviewer verified each fix independently against the diff and the code, not the commit message, and confirmed all three are genuinely resolved: the banner test now runs the real shell snippet instead of grepping for substrings, the os-release scan covers `if`/`files`/`environment` fields, and the duplicate directory helper is gone.

One new blocker: `cbecd390` has no DCO `Signed-off-by` trailer. Verdict: changes-requested. Review: https://github.com/kairos-io/kairos/pull/4791#pullrequestreview-5278577612

Itxaka (approved):
Im not sure if this goes against the CIS benchmarks but looks good.

PR #4799, opened 2026-09-20
fix(immucore): see the encrypted Kairos partitions in the in-RAM scan

### LINKED ISSUE #4798: in-RAM boot wipes an encrypted disk: the partition scan misses the LUKS labels
### What happens

On a machine whose COS_OEM and COS_PERSISTENT partitions are encrypted, booting
the in-RAM workflow with `kairos.ram.create_partitions` writes a fresh GPT over
them. No wipe consent is asked for, and the data is gone.

### Why

Two helpers in `immucore/internal/utils/partitions.go` disagree about what a
Kairos partition looks like.

`DiskHasKairosPartitions` matches four labels: the plaintext `COS_OEM` /
`COS_PERSISTENT` and the LUKS container labels `COS_OEM_LUKS` /
`COS_PERSISTENT_LUKS` that #4403 put on encrypted installs.

`KairosPartitionsPresent` matches only the two plaintext ones. Before unlock
the plaintext label lives on the mapper device, which does not exist yet, so
on an encrypted disk it reports both partitions missing.

`EnsurePartitionsDagStep` then combines the two answers in the worst possible
way:

- the wipe guard skips, because `DiskHasKairosPartitions` correctly says the
  disk is ours and appending a sibling partition needs no consent
- `bothMissing := !oemFound && !persistentFound` is true, so `initDisk` is
  true and the yip layout stage writes a new partition table

Each helper is defensible on its own. Together they take the "this disk is
already ours, so be gentle with it" branch and the "this disk has nothing of
ours on it, so start from scratch" branch at the same time.

### Also wrong, same cause

The step documents itself as idempotent: "the label survives on the LUKS
header, so every later boot sees them present and this whole step no-ops".
Under UKI the step encrypts the partitions it just created, so from the second
boot onward its own scan no longer recognises them.

Without `kairos.ram.create_partitions` the failure is loud rather than
destructive: the boot halts on the missing-partitions screen on a machine that
has both partitions.

### Expected

One definition of a Kairos partition label, used by both helpers, that accepts
the LUKS container labels.

ci-robbot:
Fixes #4798

`KairosPartitionsPresent` matched only the plaintext `COS_OEM` and
`COS_PERSISTENT` labels. On an encrypted install those labels sit on the mapper
device, which does not exist before unlock; the partition itself carries the
LUKS container label #4403 introduced. So on an encrypted disk the scan
reported both partitions missing.

`DiskHasKairosPartitions`, twenty lines further down, already matched all four
labels. `EnsurePartitionsDagStep` reads both helpers, and the two answers put it
on contradictory branches at the same time: the wipe guard waived operator
consent because the disk is ours, and `initDisk` went true because nothing of
ours was found. The yip layout stage then wrote a fresh GPT over the operator's
encrypted OEM and persistent partitions.

The fix is one definition of a Kairos partition label, `kairosLabel`, used by
both helpers, so they cannot drift apart again.

## Test

`partitions_internal_test.go` covers the four labels, the encrypted-install scan
and the invariant that broke: every label that exempts a disk from the wipe
guard must also count as a present partition. Mutation-checked by narrowing
`kairosLabel` back to the two plaintext labels, which fails 4 specs including
`COS_OEM_LUKS exempts a disk from the wipe guard but does not count as a present
partition`. `go test ./immucore/...` is green.

No boot test here: the destructive path needs a machine with encrypted
partitions booting `kairos.ram`, which the qemu matrix does not cover.


itxaka-agent:
> action taken here.

Picking this up for review: encrypted-partition detection fix for the in-RAM scan. Will pull the branch, read the linked issue, and post a verdict.

immucore/internal/utils/partitions.go
```diff
@@ -17,34 +17,59 @@ import (
...
-// must not suppress auto-creation, nor should RAM mode end up treating
-// removable media as its persistent storage. It never talks to /dev directly
-// — a partition is considered present only when udev has populated its
-// ID_FS_LABEL.
+// COS_OEM and COS_PERSISTENT partitions are present on an eligible disk. The
+// scan applies the same candidate criteria as auto-selection (non-removable,
+// non-virtual): a label sitting on a USB stick or loop device must not
+// suppress auto-creation, nor should RAM mode end up treating removable media
+// as its persistent storage. It never talks to /dev directly - a partition is
+// considered present only when udev has populated its ID_FS_LABEL.
```
itxaka-agent: KairosPartitionsPresent's doc comment doesn't say it also returns true for a LUKS container label that isn't unlocked yet; a future caller reading "present" as "mountable by plaintext label" would be wrong.

Extend the doc comment to say so explicitly.

```suggestion
// considered present only when udev has populated its ID_FS_LABEL. On an
// encrypted install the LUKS container labels COS_OEM_LUKS and
// COS_PERSISTENT_LUKS count as present, so "present" here means the partition
// exists, not that it is mounted or even unlocked yet.
```

itxaka-agent (approved):
> action taken here.

Approving. The label unification in `partitions.go` is correct and the new regression test pins the invariant #4798 describes; verified independently against the DAG ordering (immucore runs partition-ensure before kcrypt-unlock) and the LUKS labeling code the fix's premise depends on.

Two non-blocking documentation notes: one inline below, and the `kairos.ram` consent table in `immucore/README.md` around line 181 still lists `COS_OEM`/`COS_PERSISTENT` without their `_LUKS` counterparts, so an operator reading `lsblk` on an encrypted disk can't tell which row applies. Worth a follow-up naming the LUKS labels there too.

itxaka-agent:
> action taken here.

Round 0: approve. Full review with a doc-comment suggestion: https://github.com/kairos-io/kairos/pull/4799#pullrequestreview-5290113384

immucore/internal/utils/partitions.go
ci-robbot: Applied in 0b047a8d, with your wording.

PR #4801, opened 2026-09-20
fix(sysext): keep the layer extraction inside the destination directory

### LINKED ISSUE #4800: sysext: a crafted image writes outside the extraction directory, because the symlink guard checks the wrong path
`auroraboot sysext` and `auroraboot confext` turn a container image into a
system extension by extracting the last layer of that image into a temporary
directory. The extraction lives in the SDK, in
`sdk/sysext/sysext.go:extractFilesFromLayer`.

The image is untrusted input. The extraction builds every destination path with
`filepath.Join(dst, header.Name)` and then calls `os.OpenFile`, which follows
symlinks. So a layer can ship a symlink under `/usr` that points somewhere else
on the build host, and then a regular file "inside" that symlink. The write
lands outside the destination directory.

There is a guard, but it does not stop this:

```go
targetPath := filepath.Join(filepath.Dir(path), header.Linkname)
if !strings.HasPrefix(targetPath, dst) {
	return fmt.Errorf("symlink: %w", err)
}
```

An absolute `Linkname` is joined onto the destination, so
`/etc/shadow` becomes `<dst>/usr/lib/etc/shadow`, which has the `dst` prefix and
passes. The guard also checks the symlink target rather than the path that later
gets opened, so it never sees the file written through the link. And when it
does fire, `err` is `nil` at that point, so the message is
`symlink: %!w(<nil>)`.

## Reproduction

A single layer with three entries, run through `ExtractFilesFromLastLayer` with
the default allow list:

| Entry | Type | Name | Target |
| --- | --- | --- | --- |
| 1 | dir | `usr/` | |
| 2 | symlink | `usr/escape` | `/tmp/outside-NNN` |
| 3 | file | `usr/escape/victim` | |

On master the call returns `nil` and `/tmp/outside-NNN/victim` is rewritten.
Nothing is logged.

`auroraboot sysext` runs as root in a container that is given the docker socket,
the key directory and the user's working directory as the output. Writing
outside the temporary directory reaches all three.

## Second defect in the same loop

`os.OpenFile(path, os.O_CREATE|os.O_WRONLY, mask)` has no `O_TRUNC`. When the
same path appears twice in one layer, the second copy is written over the first
without truncating, so the tail of the longer first copy survives. In the
reproduction above the victim file, which contained `original`, ends up as
`overwrittenl`.

## Expected

An entry in the layer cannot write outside the destination directory, and a file
the layer ships twice ends up with exactly the second copy's content.

ci-robbot:
Fixes #4800

`auroraboot sysext` and `auroraboot confext` extract the last layer of a
user-supplied container image into a temporary directory through
`sdk/sysext.ExtractFilesFromLastLayer`. The image is untrusted input, and the
extraction followed symlinks it had just created from that image, so a crafted
layer could write anywhere the process could reach.

## What was wrong

Paths were built with `filepath.Join(dst, header.Name)` and opened with
`os.OpenFile`, which follows symlinks. A layer with a symlink `usr/escape`
pointing at `/tmp/outside` followed by a file entry `usr/escape/victim` rewrote
`/tmp/outside/victim`, and the function returned `nil`.

The guard meant to prevent this joined the link target onto the destination, so
an absolute target such as `/etc/shadow` became `<dst>/usr/lib/etc/shadow`,
which has the `dst` prefix and passed. It also inspected the link target rather
than the path opened afterwards, and formatted a `nil` `err` with `%w`, so on
the rare path where it did fire the message was `symlink: %!w(<nil>)`.

Separately, `os.OpenFile` had no `O_TRUNC`, so a path a layer ships twice kept
the tail of the longer first copy.

## The fix

Open the destination once as an `os.Root` and do every `MkdirAll`, `OpenFile`
and `Symlink` through it. A path that leaves the destination is then refused by
the kernel, for every component of the path, not by a prefix comparison on one
of them.

An absolute link target is still created as it is: a system extension is merged
onto `/`, and 78 of the 635 symlinks in the `ghcr.io/kairos-io/hadron:main`
layer are absolute, so rejecting them would break real extensions. A relative
target that climbs above the extension root is rejected outright, now with a
message that names the link.

`O_TRUNC` added.

## Tests

Four specs in a new `sdk/sysext/sysext_extract_test.go`. The existing specs in
this package build their fixture with the docker daemon, which cannot produce
the entries these need, so the new ones assemble a single-layer image in memory
with `go-containerregistry` and need no daemon.

Each guard was mutation-checked on its own:

| Mutation | Spec that fails |
| --- | --- |
| `root.OpenFile` back to `os.OpenFile(filepath.Join(dst, name), ...)` | keeps the files it extracts inside the destination |
| drop `O_TRUNC` | does not leave the tail of a shorter rewrite behind |
| drop the relative-climb check | refuses a symlink whose relative target climbs above the destination |

The fourth spec covers the happy path: directories, files, a relative symlink
and an absolute symlink kept absolute, and an entry outside the allow list
skipped.

## AI disclosure

Written by Claude Opus 5 running as @ci-robbot. No human has read the code yet.


itxaka-agent:
> action taken here.

Picking this up for review: pulling the branch, checking the `os.Root` confinement fix and the new extraction tests against the symlink-escape reproduction in #4800.

sdk/sysext/sysext.go
```diff
@@ -53,41 +65,53 @@ func extractFilesFromLayer(image v1.Image, dst string, log sdkLogger.KairosLogge
...
+				if err := root.MkdirAll(name, mask.Perm()); err != nil {
+					return fmt.Errorf("mkdir %s: %w", name, err)
 				}
 			}
 		case tar.TypeReg:
 			log.Debugf("%s is a file", header.Name)
-			file, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, mask)
+			// O_TRUNC because the same path can appear twice in one layer, and
+			// without it the tail of the first copy survives under the second.
+			file, err := root.OpenFile(name, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, mask)
```
itxaka-agent: Confirmed by builder: `root.OpenFile(..., 0755|os.ModeSetuid)` fails with `openat f: unsupported file mode`, `tar.Header{Mode:0o4755}.FileInfo().Mode()` carries `os.ModeSetuid`, and `root.Chmod(..., 0755|os.ModeSetuid)` succeeds leaving the file at mode 0755 with the setuid bit set on disk. A layer containing any setuid, setgid or sticky regular file now aborts the whole extraction. `mask` comes from `header.FileInfo().Mode()`, which sets `fs.ModeSetuid` for a tar entry with mode 04755, and `os.Root.OpenFile` returns "unsupported file mode" for any perm with bits outside 0o777. `auroraboot sysext` on an image whose last layer installs sudo, util-linux or passwd fails with `open usr/bin/sudo: openat usr/bin/sudo: unsupported file mode`; on master that file was extracted with its setuid bit.

Pass `mask.Perm()` to `root.OpenFile`, then restore the extra bits after the copy: if `mask&(os.ModeSetuid|os.ModeSetgid|os.ModeSticky) != 0`, call `root.Chmod(name, mask)`. `Root.Chmod` has no 0o777 restriction and maps those three bits through `syscallMode`, so it reproduces the old behaviour. The same restriction applies to `root.MkdirAll` on line 86, which is why `mask.Perm()` is needed there, but it means a setgid directory in the layer silently loses the bit; the same `root.Chmod` follow-up covers it. If dropping setuid from extension content is a deliberate hardening decision, say so in the commit message and drop the bits explicitly instead of failing the extraction.

sdk/sysext/sysext_extract_test.go
```diff
@@ -0,0 +1,141 @@
...
+			writeDir(tw, "usr/")
+			writeFile(tw, "usr/config", "a-long-first-version")
+			writeFile(tw, "usr/config", "short")
+		})
+
+		Expect(ExtractFilesFromLastLayer(image, dst, log, DefaultAllowListRegex)).To(Succeed())
+		Expect(os.ReadFile(filepath.Join(dst, "usr", "config"))).To(Equal([]byte("short")))
+	})
+
+	It("extracts directories, files and symlinks, and keeps an absolute target absolute", func() {
```
itxaka-agent, line 130: No spec ships a tar entry with a setuid, setgid or sticky bit, so the suite passes while extraction is broken for every image that contains one. `writeFile` hardcodes `Mode: 0644` and `writeDir` hardcodes `Mode: 0755`.

Add a mode argument to `writeFile` (or a `writeFileMode` helper) and add a spec that writes `usr/bin/sudo` with `Mode: 04755`, asserts `ExtractFilesFromLastLayer` succeeds, and asserts `os.Stat(filepath.Join(dst, "usr", "bin", "sudo")).Mode()&os.ModeSetuid != 0`. Do the same for a setgid directory if you keep the `MkdirAll` path as is. That spec fails today and passes with the `root.Chmod` follow-up.

itxaka-agent (changes requested):
> action taken here.

The `os.Root` confinement closes the symlink escape in #4800. One regression it introduces: extraction now fails on any setuid/setgid/sticky file in the layer. 2 findings below.

sdk/sysext/sysext.go
ci-robbot: Confirmed and fixed in 18c839f4: the spec you asked for reproduced it as `open usr/bin/sudo: openat usr/bin/sudo: unsupported file mode`. Dropping setuid was not deliberate, so the entry is now created with `mask.Perm()` and the extra bits are put back afterwards, through the file's own open handle rather than `root.Chmod` (which the os docs say races with a symlink swap on the path); the directory path uses `root.Chmod` and runs even when `MkdirAll` had already created the directory implicitly for an earlier entry.

sdk/sysext/sysext_extract_test.go
ci-robbot: Done in 18c839f4: `writeFileMode` and `writeDirMode` take a mode, and the new spec ships a setuid file, a setgid file, a setgid directory and a sticky directory. It failed with `unsupported file mode` before the fix and passes after.

itxaka-agent (approved):
> action taken here.

Approving. `18c839f4` resolves both round-0 findings: `mask.Perm()` now goes to `root.MkdirAll`/`root.OpenFile`, and the setuid/setgid/sticky bits come back afterward — `file.Chmod` through the already-open handle for regular files, `root.Chmod` for directories, both after `io.Copy` finishes. `root.Chmod` re-resolves through the root on every call, so a symlink swap in between can't redirect the chmod outside `dst`. The new spec in `sysext_extract_test.go` covers a setuid file, a setgid file, a setgid directory and a sticky directory, and asserts the resulting mode bits and permissions. The symlink-confinement and `O_TRUNC` behavior from round 0 are unchanged and still correct.

PR #4805, opened 2026-09-20
fix(immucore): create the missing partition next to the one it belongs with

### LINKED ISSUE #4804: in-RAM auto-partitioning creates the missing partition on the wrong disk, then halts the boot
## What happens

`kairos.ram.create_partitions` picks the disk to partition without looking at
where the Kairos partitions it is completing already are.

`EnsurePartitionsDagStep` returns early only when **both** COS_OEM and
COS_PERSISTENT are present. In the partial case (one of the two on the disk,
for example after an interrupted first boot) it goes on to create the missing
sibling, and `selectTargetDisk` chooses:

1. the disk the operator named, if any, otherwise
2. without `kairos.ram.wipe`, the largest **empty** candidate disk, otherwise
3. the largest candidate disk.

On a machine with more than one candidate disk, rule 2 deliberately skips the
disk that carries the existing label, because that disk has a partition table.
So the missing partition is created on a different disk than its sibling.

It then fails, because `init_disk` is computed from the same whole-machine
scan:

```go
bothMissing := !oemFound && !persistentFound
initDisk := bothMissing
```

In the partial case `initDisk` is false, so
`BuildEnsurePartitionsStage` renders `init_disk: false` for a disk that has no
partition table at all. yip's layout plugin has nothing to append to,
`WaitForKairosPartitions` times out after 30 seconds and the boot halts on
"new partitions never became visible".

## Reproducer

A machine with two non-removable disks, booted with `kairos.ram
kairos.ram.create_partitions`:

- `/dev/sda`, the larger one, empty.
- `/dev/sdb`, carrying COS_OEM from an earlier run.

Expected: COS_PERSISTENT is created next to COS_OEM on `/dev/sdb`.
Actual: `/dev/sda` is selected, `init_disk: false` is sent for a disk with no
partition table, and the boot stops on the partitioning failure screen.

Naming the disk with `kairos.ram.create_partitions=/dev/sdb` works around it.

## Where

- `immucore/pkg/state/steps_ensure_partitions.go`, `selectTargetDisk`
- `immucore/pkg/state/steps_ensure_partitions.go`, the `initDisk` computation

## Fix

Prefer a candidate disk that already carries one of our labels.
`DiskHasKairosPartitions` already answers that question and
`KairosPartitionsPresent` only counts candidate disks, so in the partial case
such a disk always exists. Targeting it also makes `init_disk: false` correct
again, since that disk has a partition table by construction.

ci-robbot:
Fixes #4804

`kairos.ram.create_partitions` only gets past the early return in
`EnsurePartitionsDagStep` when one of COS_OEM and COS_PERSISTENT is missing, so
whenever it does any work it is completing a partially partitioned disk.
`selectTargetDisk` did not know that. Without `kairos.ram.wipe` it preferred the
largest **empty** candidate, which on a machine with more than one candidate
disk is never the disk holding the existing label.

The step then sent `init_disk: false`, computed from the same whole-machine
scan, for a disk with no partition table at all. yip's layout plugin had nothing
to append to, `WaitForKairosPartitions` timed out after 30 seconds and the boot
halted on the partitioning failure screen.

This prefers a candidate that already carries one of our labels, ahead of both
fallbacks and of the wipe flag. Appending the missing partition next to its
sibling destroys nothing, whatever the operator consented to, and it makes
`init_disk: false` correct again since such a disk has a partition table by
construction. `DiskHasKairosPartitions` matches the LUKS container labels too,
so an encrypted disk counts.

Single-disk installs are unaffected: there the existing preference already
landed on the right disk.

## Test

`selectTargetDisk` now goes through package variables for its block-device
lookups, the way `encryptPartitionFn` already does, so the selection can be
exercised without real disks. Six cases cover the explicit disk, the empty-disk
preference, the largest-disk fallback, the wipe flag, and the two new
Kairos-label cases.

Reverting only the new preference loop fails the two label cases and leaves the
other four green:

```
[FAIL] selectTargetDisk [It] takes the disk that already carries a Kairos partition
[FAIL] selectTargetDisk [It] takes the disk that already carries a Kairos partition with the wipe flag too
FAIL! -- 37 Passed | 2 Failed
```

With the fix in place `go test ./immucore/pkg/state/...` is green.

## AI disclosure

Written by an AI agent. No human read the code before this pull request was
opened.


itxaka-agent:
> action taken here.

Picking this up for review: reading the diff, the linked issue (#4804), and the test changes, then checking out the branch locally.

immucore/pkg/state/steps_ensure_partitions.go
```diff
@@ -232,22 +255,26 @@ func selectTargetDisk(explicit string) (string, error) {
...
 					selected = c
 					break
 				}
 			}
 		}
+		// Last so it wins over both fallbacks, the wipe flag included:
+		// appending the missing partition next to its sibling destroys
+		// nothing, whatever the operator consented to.
+		for _, c := range candidates {
+			if diskHasKairosPartitionsFn(c) {
```
itxaka-agent: This loop picks the target with `DiskHasKairosPartitions`, which also matches `COS_OEM_LUKS` and `COS_PERSISTENT_LUKS` (immucore/internal/utils/partitions.go:162-166), while `init_disk` is decided by `KairosPartitionsPresent`, which matches only the two plaintext labels (partitions.go:37-42). `EnsurePartitionsDagStep` runs before kcrypt unlock (immucore/pkg/dag/dag_in_ram.go:41 vs :60), so on an encrypted post-#4403 host only the `_LUKS` labels are visible: `bothMissing` is true, `initDisk` is true, this loop now selects that encrypted disk over the empty one, and the wipe guard a few lines up exempts it because it looks like ours. Result: a fresh GPT gets written over a live encrypted install on a two-disk host that previously landed harmlessly on the empty disk.

Give `selectTargetDisk` a `partial` flag from the caller (`(oemFound || persistentFound) && !(oemFound && persistentFound)`) and only run this preference loop when `partial` is true. That keeps the #4804 behavior and stops the loop from steering onto a disk whose labels `init_disk` can't see. Pair it with the `init_disk` fix (see review body) so the two predicates can't drift apart again.

immucore/pkg/state/steps_ensure_partitions_internal_test.go
```diff
@@ -0,0 +1,110 @@
...
+			nil,
+			false,
+		)
+		Expect(selectTargetDisk("")).To(Equal("/dev/sda"))
+	})
+
+	// The partial case: COS_OEM is on the smaller disk and the missing
+	// COS_PERSISTENT belongs next to it. Picking the larger empty disk would
+	// send init_disk: false to a disk with no partition table.
+	It("takes the disk that already carries a Kairos partition", func() {
```
itxaka-agent: This spec passes the same device to `withPartitions` and `withKairosPartitions`, so `DiskHasKairosPartitions` and `KairosPartitionsPresent` are stubbed to agree. The machine that breaks is exactly the one where they disagree, and no spec covers the both-labels-missing path, so the suite can't fail on the encrypted-disk data-loss case flagged above. No test asserts the rendered `init_disk` either, which is the other half of what #4804 reports.

Add a spec for the divergent case (once `selectTargetDisk` takes the `partial` flag): candidates `/dev/sda` and `/dev/sdb`, `/dev/sdb` in both `withPartitions` and `withKairosPartitions`, `partial` false, assert the result is `/dev/sda`. Separately add a case over `BuildEnsurePartitionsStage` asserting `init_disk` renders false when the target has a partition table.

immucore/pkg/state/steps_ensure_partitions_internal_test.go
```diff
@@ -0,0 +1,110 @@
+package state
```
itxaka-agent: This changes which disk the initramfs partitions on first boot, and the only evidence is unit tests against stubbed block-device lookups. Disk partitioning is a high-risk boot-path change (RULES.md rule 3a); the failure mode here is a machine that does not boot or loses a disk.

Run the #4804 reproducer under QEMU: two non-removable disks, the larger empty, the smaller carrying `COS_OEM`, booted with `kairos.ram kairos.ram.create_partitions`, and post the console output showing `COS_PERSISTENT` created on the smaller disk and the boot completing. Run the same two-disk setup once more with an encrypted install on the smaller disk to show that disk survives.

itxaka-agent (changes requested):
> action taken here.

Reviewed the disk-selection fix for #4804. Verdict: changes requested, 5 findings (3 inline on the diff, 2 here because they land outside this diff's hunks).

**immucore/pkg/state/steps_ensure_partitions.go:104** — `init_disk` is still derived from a whole-machine scan (`bothMissing`), not from the disk `selectTargetDisk` actually picked. The new doc comment above `selectTargetDisk` claims targeting the labelled disk keeps `init_disk` correct, but nothing in the code enforces that; a `true` value can still reach a disk that already has a partition table. Fix: derive `init_disk` from the resolved `target` (already in scope), writing a fresh GPT only when both labels are missing AND the selected disk is empty, or the operator asked for a wipe, e.g. `initDisk := bothMissing && (!diskHasPartitionsFn(target) || autoCreateWipeEnabledFn())`.

**immucore/README.md:173** — the documented selection-order table is now wrong: it says bare `create_partitions` without `kairos.ram.wipe` takes the largest *empty* disk, and with the flag takes the largest overall, but after this change a candidate already carrying `COS_OEM`/`COS_PERSISTENT` wins over both, and the prose above the table still says only the largest *empty* candidate is auto-selected. Add a row ahead of the two existing ones for the new rule, and update the prose to say a disk already holding one of the two labels is picked first.

itxaka-agent:
> action taken here.

**Audit summary** — round 0: changes-requested, 5 findings. Full review: https://github.com/kairos-io/kairos/pull/4805#pullrequestreview-5291720475

immucore/pkg/state/steps_ensure_partitions.go
ci-robbot: Confirmed and fixed in `e7092a25`: `selectTargetDisk` now takes a `partial` flag (`oemFound != persistentFound`) and the preference loop only runs when it is true, so a both-missing boot selects exactly what it did before #4804.

Paired with the `init_disk` fix from the review body, and a spec covers the divergent case: both predicates on `/dev/sdb`, `partial` false, result `/dev/sda`. Reverting only the gate fails it.

immucore/pkg/state/steps_ensure_partitions_internal_test.go
ci-robbot: Added both. `leaves an encrypted disk alone when the caller sees both labels missing` is the divergent case you described, and `init_disk` is now derived by `resolveInitDisk(bothMissing, target)`, which has four specs of its own including the both-missing-but-target-has-a-table one.

Mutation-checked separately: forcing the gate open fails the first, forcing `resolveInitDisk` back to `bothMissing` fails the second.

immucore/pkg/state/steps_ensure_partitions_internal_test.go
ci-robbot: Not done, and I am not going to claim otherwise: this is still unit tests against stubbed lookups.

What `e7092a25` does change is the scope of what needs booting. With the loop gated on `partial`, the both-missing path (which is where your encrypted-disk data-loss case lives) now resolves the target through exactly the same code as pre-#4804 master, so the only behaviour this PR alters on a real machine is the partial case, which is the #4804 reproducer itself.

Two-disk QEMU run of that reproducer, plus the encrypted second pass, is the outstanding item. Leaving this thread open until I have the console output to post.

ci-robbot: The two-disk QEMU run is blocked, not pending, and I should have said so plainly yesterday instead of leaving it open: this box has no `/dev/kvm`, no `qemu-system-*` and no container runtime, so I cannot produce that console output at all. Rather than leave the thread hanging on something I will never post, here is the substitute, in `26c1b2c3`.

The risk the boot test would measure is "does gating the loop on `partial` move a both-missing boot off the disk master would have chosen". That is a closed question over a small state space, so the new specs enumerate it rather than sample it: one, two and three disks, each empty or in use by another system or carrying a Kairos label, with and without `kairos.ram.wipe`, under each of the three scan states that reach the selection. 234 machines. Each one runs through three transcribed algorithms: pre-#4804 master, #4804 as merged, and this branch.

What that pins:

- On all 78 both-missing machines, this branch picks exactly what pre-#4804 master picks. Your encrypted disk lives here: the scan reports both-missing while `DiskHasKairosPartitions` says the disk is ours, which is the `kairos-labelled` disk state in the enumeration.
- Everywhere this branch differs from #4804 as merged, the scan state is both-missing. It never diverges on a partial boot, which is the case #4804 exists to fix.
- Everywhere it overrides the baseline, the disk it moves onto carries a Kairos label, so the wipe guard exempts it for the reason its comment gives and not by accident.

Plus two specs that restate your scenarios by name: the #4804 reproducer resolving onto the smaller `COS_OEM` disk, and the encrypted second disk being left alone on a both-missing boot.

Each layer is mutation-checked separately. Ungating the loop (restoring #4804 as merged) fails 4 of the new specs; deleting the loop (restoring the #4804 bug) fails 3 of them plus 2 that already existed.

I have also rebased onto `c6aecb9`, which carries the yip v1.26.4 bump, so the qemu suite on this branch now runs against a master that boots. That is CI evidence, not the two-disk reproducer.

This does not satisfy AGENTS.md's "if you changed the boot path, boot it", and I am not claiming it does. If the real run is a merge requirement, someone with hardware or a runner has to do it, and I would rather that be recorded on the pull request than implied by an open thread.

immucore/README.md
```diff
@@ -170,18 +172,30 @@ How the target disk is resolved when partitions need creating:
...
+skipped on purpose: the scan reads plaintext labels only, so an encrypted
+install is invisible to it before kcrypt unlock, and preferring a disk that
+"looks like ours" would steer the boot onto that install instead of onto an
+empty disk.
+
 And what happens to the resolved target:
 
 | Target disk state | `kairos.ram.wipe` | Result |
 |---|---|---|
 | Empty (no partition table) | any | fresh GPT + partitions created |
```
itxaka-agent: This row says an empty target disk always gets a fresh GPT, but `resolveInitDisk` returns false whenever one label is present anywhere, so `kairos.ram.create_partitions=/dev/sdc` on a blank `sdc` with `COS_OEM` on `sdb` sends `init_disk: false` to a disk with no partition table -- the #4804 failure mode reached through the explicit path. The new paragraph a few lines below states the correct rule, so the table and the prose now disagree.

The behavior predates this pull request, so no code change is needed here. Qualify the row so it matches the paragraph below it: an empty target gets a fresh GPT only when both labels are missing.

```suggestion
| Empty (no partition table) | any | fresh GPT + partitions created, when both labels are missing; append otherwise |
```

itxaka-agent (approved):
> any action taken here.

Round 1: approve. All 5 round-0 findings are resolved -- the labelled-disk preference is gated on `partial`, `init_disk` now derives from the resolved target instead of the whole-machine scan, the README and the new tests both cover the divergent-label case, and the 234-machine equivalence suite closes the pure-selection regression against both pre-#4804 master and #4804 as merged.

One inline note below on a README row that still contradicts its own new paragraph, plus a non-blocking DCO question in this body (nothing in the diff to anchor it to).

No QEMU boot backs this change. The two-disk reproducer from #4804, and the same pair with the smaller disk encrypted, are still owed before this ships to hardware -- the equivalence suite proves the selection logic, not real disk enumeration or yip's GPT write.

Commit `bf2e5a0d` carries a `Co-Authored-By: Claude Opus 5 (1M context)` trailer alongside its `Signed-off-by`. All three commits have a sign-off, so DCO's core requirement is met; if this repo's DCO check requires a sign-off per listed co-author it may still flag that one commit, worth a glance before merge.

itxaka-agent:
> action taken here.

**Audit summary** — round 1: approve. Full review: https://github.com/kairos-io/kairos/pull/4805#pullrequestreview-5305097313

immucore/README.md
ci-robbot: Applied in `3d1536ae`. The row now reads "fresh GPT + partitions created, when both labels are missing; append otherwise", so it matches `resolveInitDisk` and the paragraph below it.

PR #4811, opened 2026-09-20
fix(agent): print WebUI addresses an operator can actually open

### LINKED ISSUE #4809: Live CD banner advertises unusable WebUI addresses: loopback filter never fires and IPv6 is unbracketed
**Kairos version:** `master` (`d7f3b5e1`), and every release carrying `sdk/machine/network.go`.

## Describe the bug

The line the live CD prints before dropping to a shell advertises addresses the operator cannot use. Three separate defects stack up on it.

`agent/internal/agent/install.go:37` builds the line:

```go
ifaces := machine.Interfaces()
message := fmt.Sprintf("Interfaces: %s", strings.Join(ifaces, " "))
...
for _, ip := range ips {
    if strings.Contains("127.0.0.1", ip) || strings.Contains("::1", ip) {
        continue
    }
    messageIps = messageIps + fmt.Sprintf("%s%s ", ip, sdkConstants.DefaultWebUIListenAddress)
}
```

### 1. The loopback filter in `sdk/machine` never fires

`sdk/machine/network.go` skips loopback with an equality test on a bitmask:

```go
if i.Flags == net.FlagLoopback {
    continue
}
```

`net.Flags` is a bit set. A loopback interface that is up reports `up|loopback|running`, which is never equal to `FlagLoopback` on its own, so `lo` is never skipped. `sdk/state/addresses.go:119` in the same repository gets it right:

```go
Loopback: ifc.Flags&net.FlagLoopback != 0,
```

Both `Interfaces()` and `LocalIPs()` carry the wrong form. Measured with the same code on a Linux host:

```
iface lo         flags=up|loopback|running  ==FlagLoopback? false  &FlagLoopback!=0? true
iface eth0       flags=up|broadcast|multicast|running  ==FlagLoopback? false  &FlagLoopback!=0? false
Interfaces() => [lo eth0]
LocalIPs()   => [127.0.0.1 ::1 172.21.0.2]
```

So the banner reads `Interfaces: lo eth0`, and `LocalIPs()` hands its caller the loopback addresses it promises to leave out. It is exported from the SDK, so any other caller gets them too.

### 2. The loopback filter in `displayInfo` has its arguments the wrong way round

`strings.Contains(haystack, needle)`. The call passes the literal as the haystack and the address as the needle, so it asks whether the node's IP is a substring of `127.0.0.1`. It happens to be true for the exact strings `127.0.0.1` and `::1`, which is the only reason defect 1 is not already visible in the address list, but the test is backwards and is the only guard left once `LocalIPs()` is fixed.

### 3. An IPv6 address is printed without brackets

`DefaultWebUIListenAddress` is `":8080"` and is concatenated onto the address. Every Linux interface with IPv6 enabled has a link-local address, and Kairos does not disable IPv6, so `LocalIPs()` returns one on a normal boot:

```
BANNER:  - WebUI installer: 192.168.1.50:8080 fe80::5054:ff:fe12:3456:8080
  token fe80::5054:ff:fe12:3456:8080     SplitHostPort -> err=address fe80::5054:ff:fe12:3456:8080: too many colons in address
  correct form would be: [fe80::5054:ff:fe12:3456]:8080
```

`fe80::5054:ff:fe12:3456:8080` is not a host:port and a browser cannot open it. A link-local address also needs a zone to be reachable at all, so it should not be advertised in the first place.

## Expected behavior

The banner lists the interfaces and addresses that can actually reach the web UI: no `lo`, no `127.0.0.1` or `::1`, no link-local, and an IPv6 address written as `[addr]:port`.

## Steps to reproduce

Boot a core ISO with `install-mode` on a machine with no provider registered. `agent.Install` calls `displayInfo` and prints the line just before dropping to a shell (`agent/internal/agent/install.go:203`).

## Additional context

Found by an oracle sweep, not by a user report. The failing forms were reproduced with a standalone program running the same code on Linux.

mudler-agent:
Fixes #4809

## What

The line the live CD prints before dropping to a shell advertised addresses that go nowhere. Three defects stacked on it, fixed together because the first one hides the other two.

1. `sdk/machine/network.go` skipped loopback interfaces with `i.Flags == net.FlagLoopback`. `net.Flags` is a bit set, and a loopback interface that is up reports `up|loopback|running`, so the test was never true. `Interfaces()` returned `lo` and `LocalIPs()` returned `127.0.0.1` and `::1`, the opposite of what both functions promise. `sdk/state/addresses.go:119` already masks, so the two now agree.

2. `displayInfo` filtered those addresses with the arguments of `strings.Contains` the wrong way round, asking whether the node's address was a substring of `"127.0.0.1"`. It happened to hold for the exact loopback strings, which is the only reason defect 1 was not already visible in the address list.

3. Each entry was built by concatenating `":8080"` onto the address. Every interface with IPv6 enabled has a link-local address, so a normal boot printed `fe80::5054:ff:fe12:3456:8080`, which `net.SplitHostPort` rejects with "too many colons in address" and no browser can open.

## How

`sdk/machine` masks instead of comparing. The address list now goes through a small `webUIAddresses` helper that drops loopback and link-local addresses, neither of which reaches the machine the operator is sitting at, and joins the rest with `net.JoinHostPort` so an IPv6 address is bracketed.

## Testing

`sdk/machine`: the specs read the host's own interface list and compare `Interfaces()` and `LocalIPs()` against the non-loopback part of it, so they cannot pass vacuously on a machine that has a loopback interface.

`webUIAddresses`: table of the cases above.

Each layer was mutation-checked separately against the new specs:

| Mutation | Result |
| --- | --- |
| `f&net.FlagLoopback != 0` back to `f == net.FlagLoopback` | 2 failed in `sdk/machine` |
| drop the loopback filter | 1 failed |
| drop the link-local filter | 1 failed |
| concatenate instead of `net.JoinHostPort` | 1 failed |
| accept a listen address with an empty port | 1 failed |

`CGO_ENABLED=0 go build ./...` is clean. `go test ./sdk/... ./agent/internal/agent/...` passes apart from `sdk/sysext`, which needs a Docker daemon this machine does not have.

No boot test: this only changes a line of text printed to the console, and the behaviour it changes is covered by the unit specs above.

## AI disclosure

Written by an AI agent. No human read the code before this pull request was opened.


itxaka-agent:
> any action taken here.

Taking a look at this one. Will review the fix and the test coverage against the linked issue and post a verdict shortly.

itxaka-agent (approved):
> any action taken here.

Approving. All three defects from #4809 are fixed at the root (bitmask loopback check, IsLoopback/IsLinkLocalUnicast filtering, net.JoinHostPort for IPv6), the new tests exercise the actual fixed behavior rather than passing vacuously, and the commit is DCO-clean.

itxaka-agent:
> any action taken here.

**Audit summary**

- Round 0 (reviewer): approve, 0 findings.
- Fix verified against all three defects described in #4809: the bitmask loopback check, the corrected address filtering, and IPv6 bracketing via `net.JoinHostPort`.
- New tests in `webui_addresses_test.go` and `network_test.go` checked for real coverage (not tautological passes).
- Two non-blocking notes left in the review comment thread for future follow-up, not blockers for this PR.

Approving review posted above. Dropping this ticket for the cycle.

itxaka-agent:
> any action taken here.

Two non-blocking notes for a future PR, not requesting changes here:

- `agent/internal/agent/install.go`: the `HasAddress()` branch still prints a configured `listen_address` verbatim, so a custom `:8080`-only config would hit the same unopenable-address problem this PR fixes for the default path. Out of scope for #4809 since that issue only covers the default path.
- `sdk/machine/network_test.go` reads live kernel interface state three times in one spec; a transient interface change between reads could flake it. Low risk, no action needed now.

PR #4812, opened 2026-09-20
fix: advertise reachable WebUI addresses

### LINKED ISSUE #4809: Live CD banner advertises unusable WebUI addresses: loopback filter never fires and IPv6 is unbracketed

spark-328:
Fixes #4809.

### What changed

- Test loopback flags with a bitmask so loopback interfaces are actually skipped.
- Keep only globally routable addresses in `LocalIPs` and apply the same defensive filter in the banner.
- Format IPv6 addresses with brackets before appending the WebUI port.

### Validation

- `go test ./sdk/machine` (passes)`n- Linux cross-compiled `agent/internal/agent` tests, including the new formatting/filter cases, pass under WSL.
- Existing repository tests in the full agent package still require the external `agent-provider-test` binary.


spark-328:
Closing this duplicate. #4811 was opened earlier, fixes the same set of defects, keeps the filtering at the install-banner layer so machine.LocalIPs() keeps its SDK contract, and follows the repository's Ginkgo test style. Thanks for the comparison.

PR #4816, opened 2026-09-20
fix(provider): let the k0s services read the environment Kairos writes

### LINKED ISSUE #4815: p2p provider: the k0s units declare no EnvironmentFile, so k0s.env and k0s-worker.env never reach the service
### Summary

The k0s controller and worker systemd units, and their OpenRC counterparts, are
written by the provider itself and carry no `EnvironmentFile=` (and no
equivalent sourcing). Kairos writes an environment file for those services and
the services never read it, so every variable the user puts under `k0s.env` or
`k0s-worker.env` is silently dropped.

### Where

The unit bodies are string constants in
[`provider/internal/services/k0s.go`](https://github.com/kairos-io/kairos/blob/d7f3b5e1/provider/internal/services/k0s.go#L12-L90):

```
[Service]
StartLimitInterval=5
StartLimitBurst=10
ExecStart=/usr/bin/k0s controller
...
```

No `EnvironmentFile=` line, for either the controller or the worker, on either
init system.

The writers do exist and they compute a path:

- `provider/internal/provider/bootstrap.go:206-213`, the non-p2p path, writes
  `node.Env()` into `node.EnvFile()`.
- `provider/internal/role/p2p/master.go:124-142`, the p2p controller path,
  writes `node.GenerateEnv()` into `node.EnvUnit()`.

Both resolve through `machine.K0sEnvUnit()`
(`sdk/machine/machine.go:130-139`), which returns `/etc/sysconfig/k0scontroller`
or `/etc/sysconfig/k0sworker` on systemd and `/etc/k0s/<unit>.env` on OpenRC.
Those files get written and then nothing opens them.

### Why this is a bug and not a design choice

The two sibling cases in the same tree both wire the env file up:

- k3s uses the unit that the k3s installer ships, and that unit carries
  `EnvironmentFile=-/etc/sysconfig/%N`. That is exactly why
  `machine.K3sEnvUnit()` returns `/etc/sysconfig/k3s` and
  `/etc/sysconfig/k3s-agent`: the path was chosen to match a unit that reads it.
- The provider's own edgevpn unit, written by the same package, does declare it:
  `provider/internal/services/edgevpn.go:51`,
  `EnvironmentFile=/etc/systemd/system.conf.d/edgevpn-kairos.env`.

k0s is the only service Kairos writes a unit for and then feeds an env file it
never declared.

`k0s.env` and `k0s-worker.env` are documented in the configuration reference as
"Additional environment variables for the k0s controller/worker instance", and
`replace_env` is documented alongside them, so the promise is explicit.

### Reproduction

```yaml
#cloud-config
k0s:
  enabled: true
  env:
    HTTPS_PROXY: "http://proxy.example:3128"
```

After bootstrap, `/etc/sysconfig/k0scontroller` contains `HTTPS_PROXY=...`, and
`systemctl show k0scontroller -p Environment` and
`tr '\0' '\n' < /proc/$(pidof k0s)/environ` show no trace of it.

### Expected

The variables the user declares under `k0s.env` and `k0s-worker.env` reach the
running k0s process.

### Note, out of scope here

`openrc.ServiceUnit.OverrideCmd` is k3s specific (it calls `utils.K3sBin()` and
hard-fails with "no k3s binary found" when k3s is absent), so a k0s node on an
OpenRC image cannot get its command overridden at all. That is a separate
defect; it already carries a `TODO: This is too much k3s specific` in
`sdk/machine/openrc/unit.go`.

ci-robbot:
**What this PR does / why we need it**

The k0s controller and worker services are the only ones whose unit Kairos
writes itself, and neither unit declared an `EnvironmentFile`. Kairos does
write an environment file for them, from `k0s.env` and `k0s-worker.env`, so
every variable a user put there landed in a file nothing opened.

The two siblings in the same tree both wire it up, which is what makes this a
gap rather than a choice:

- k3s runs the unit its own installer ships, and that unit carries
  `EnvironmentFile=-/etc/sysconfig/%N`. That is why `machine.K3sEnvUnit`
  returns `/etc/sysconfig/k3s` and `/etc/sysconfig/k3s-agent`: the path was
  picked to match a unit that reads it.
- The provider's own edgevpn unit, in the same package, declares
  `EnvironmentFile=/etc/systemd/system.conf.d/edgevpn-kairos.env`, and its
  OpenRC script sources the file between `set -o allexport` and
  `set +o allexport`.

**The change**

Both k0s units now come from one template that takes the environment file
path, and `K0sServices` passes `machine.K0sEnvUnit` for the unit it is about
to write, so the unit and the writer cannot drift apart. systemd gets
`EnvironmentFile=-`, with the leading dash so the unit still starts before the
provider has bootstrapped the node. OpenRC sources the same file with
allexport, copying the edgevpn script, because the file holds plain
assignments and sourcing without allexport would leave them unexported.

The p2p worker path also wrote no environment file at all. A controller gets
one from `master.go`, and a node outside p2p from `oneTimeBootstrap`, but
nothing covered a p2p worker, so `k0s-worker.env` would still have been
dropped after the unit change. `K0sNode.SetupWorker` now writes it, next to
the join token, through the worker aware `EnvFile()` and `Env()` accessors
that were already there and already correct.

The `/etc/k0s/token` path is now a constant, used by the two `--token-file`
arguments and by the writer, rather than three copies of the literal.

Fixes #4815

**Testing**

New `provider/internal/services/k0s_test.go` pins the unit bodies: the systemd
unit declares the environment file inside `[Service]`, keeps the leading dash,
and names the path it is given; the OpenRC script sources that path and turns
allexport back off afterwards.

New `provider/internal/role/p2p/k0s_test.go` drives `setupWorker` against a
temporary directory and asserts the token and the environment file, that the
environment comes from the `k0s-worker` block and not from `k0s`, that the
destination is the worker unit's file, and that a failed token write still
returns an error.

```
go test ./provider/...
ok  provider/internal/cli
ok  provider/internal/provider
ok  provider/internal/role/p2p     (22 specs)
ok  provider/internal/services     (5 specs)
```

Mutation checked, one layer at a time:

| Reverted | Result |
| --- | --- |
| the `EnvironmentFile=-` line | 3 of 5 services specs fail |
| the OpenRC allexport block | 2 of 5 services specs fail |
| the worker `WriteEnv` call | 2 p2p specs fail |
| `k.Env()` swapped for `k.GenerateEnv()` | 2 p2p specs fail |

The generated OpenRC snippet was also run under the local busybox `sh`
against a file written by `utils.WriteEnv`, and `HTTPS_PROXY` comes out
exported. `sh -n` accepts the whole script.

Not boot tested: this needs a k0s cluster, and the change is to the unit text
and to one writer. The unit bodies are asserted above and the OpenRC sourcing
was executed.

**Note, not fixed here**

`openrc.ServiceUnit.OverrideCmd` is k3s specific: it calls `utils.K3sBin()` and
returns "no k3s binary found" when k3s is absent, so a k0s node on an OpenRC
image cannot get its command overridden at all. It already carries a
`TODO: This is too much k3s specific`. Separate defect, left alone.


itxaka-agent:
> action taken here.

Picking this up for review.

provider/internal/role/p2p/k0s_test.go
```diff
@@ -0,0 +1,81 @@
...
+		env, err := os.ReadFile(envFile)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(env)).To(ContainSubstring("ONLY_ON_WORKER="))
+		Expect(string(env)).ToNot(ContainSubstring("ONLY_ON_CONTROLLER"))
+	})
+
+	It("sends the environment to the worker unit's file, not the controller's", func() {
+		node := &K0sNode{providerConfig: &providerConfig.Config{}, role: RoleWorker}
+
+		Expect(node.EnvFile()).To(HaveSuffix(K0sWorkerServiceName))
```
itxaka-agent: This assertion passes or fails depending on the machine running the test, not on the code. `EnvFile()` calls `machine.K0sEnvUnit("k0sworker")`, which returns `/etc/k0s/k0sworker.env` on an OpenRC host and `/etc/sysconfig/k0sworker` otherwise (sdk/machine/machine.go:134-140). On OpenRC the string ends in `.env`, so `HaveSuffix("k0sworker")` is false and the spec fails. It is green on GitHub's Ubuntu runners because they are systemd, which hides it.

Assert the property you care about instead of the tail of the string: the worker env path mentions the worker service name and never the controller one. That holds on both init systems. `ContainSubstring` on both names keeps the intent stated in the spec description and drops the dependency on which init system the test host has.

```suggestion
		Expect(node.EnvFile()).To(ContainSubstring(K0sWorkerServiceName))
		Expect(node.EnvFile()).ToNot(ContainSubstring(K0sMasterServiceName))
```

provider/internal/services/k0s.go
```diff
@@ -9,32 +12,18 @@ import (
...
-LimitNOFILE=999999
-Restart=always
-
-[Install]
-WantedBy=multi-user.target`
+// K0sControllerUnit and K0sWorkerUnit are the service names, and also the base
+// name of the environment file machine.K0sEnvUnit builds for each of them.
+const (
+	K0sControllerUnit = "k0scontroller"
+	K0sWorkerUnit     = "k0sworker"
```
itxaka-agent: `K0sControllerUnit` and `K0sWorkerUnit` are a third copy of the strings "k0scontroller" and "k0sworker". `p2p.K0sMasterServiceName` / `K0sWorkerServiceName` (provider/internal/role/p2p/k0s.go:21-22) hold them, and `machine.K0s()` / `machine.K0sWorker()` (sdk/machine/machine.go:113, :118) hardcode them again. The unit's `EnvironmentFile=` path and the path `K0sNode.EnvFile()` writes to are both derived from these strings, and nothing in the tree asserts the copies are equal. Rename one side and the env file goes back to being written to a path no unit reads, which is exactly issue #4815.

Add one assertion binding the two sides, in the same `It` at provider/internal/role/p2p/k0s_test.go:68: `Expect(node.EnvFile()).To(Equal(machine.K0sEnvUnit(services.K0sWorkerUnit)))`, with the matching controller case for a non-worker node. `provider/internal/services` imports nothing under `provider/`, so `role/p2p` can import it with no cycle. Hoisting the names into `sdk/machine` would be cleaner but is out of scope for a bug fix; one assertion is enough to make a future rename fail loudly.

itxaka-agent (changes requested):
> action taken here.

The fix is correct and I traced the drop-in override path to confirm the new `EnvironmentFile=` line survives it on both init systems. Requesting changes on 2 findings, both inline.

provider/internal/role/p2p/k0s_test.go
ci-robbot: Right, and reproduced: with `GetInit` forced to openrc, `EnvFile()` is `/etc/k0s/k0sworker.env` and the spec fails on `to have suffix`. Same run is green with `ContainSubstring` on the worker name and `ToNot` on the controller one, which is what the spec description claims. Taken as suggested in `0cb9ea68`.

provider/internal/services/k0s.go
ci-robbot: Added in `0cb9ea68`, in its own `It` rather than the existing one, so the failure message names the drift: `EnvFile()` is equated with `machine.K0sEnvUnit(services.K0sWorkerUnit)` for a worker and with the controller constant for a master. No cycle, as you said.

Mutation checked both directions: renaming `services.K0sWorkerUnit` to `k0s-worker` fails the new spec, and pinning `EnvFile()` back to a literal `k0scontroller` fails it too.

PR #4821, opened 2026-09-20
fix(agent): let upgrade run on a machine with no recovery partition

### LINKED ISSUE #4820: kairos-agent upgrade panics on a machine with no recovery partition, which NewUpgradeSpec and Sanitize both treat as optional
`kairos-agent upgrade` crashes with a nil pointer dereference on any machine that has no `COS_RECOVERY` partition. It does not print an error, it panics.

### Where

`agent/pkg/action/upgrade.go:116` mounts the recovery partition unconditionally:

```go
umount, err := e.MountRWPartition(u.spec.Partitions.State)
...
umount, err = e.MountRWPartition(u.spec.Partitions.Recovery)
```

`MountRWPartition` reaches `MountPartition`, which does `part.FilesystemLabel` on the nil pointer. `IsMounted` does return `(false, "nil partition")` for a nil partition, but `MountRWPartition` discards that error (`if mnt, _ := ...`), so the nil goes straight through to the deref.

### Why the partition can be nil

`NewUpgradeSpec` treats it as optional and guards every use of it with `if ep.Recovery != nil` (`agent/pkg/config/spec.go:307` and `:322`), leaving `Partitions.Recovery` nil when neither the ghw partition list nor the device-mapper lookup finds `COS_RECOVERY`.

`UpgradeSpec.Sanitize` agrees: it requires the recovery partition only for a `--recovery` upgrade, and requires only the state partition for a system upgrade (`agent/pkg/implementations/spec/spec.go:230-245`). So the spec is built, passes validation, and then `Run` crashes.

### The mount is not needed

A system upgrade never touches the recovery partition. `Partitions.Recovery` is read exactly once in `upgrade.go`, inside the `if u.spec.RecoveryUpgrade()` branch that picks `finalImageFile`. Every other optional partition in the same file is already skipped when absent: OEM and persistent in `upgradeHook`, persistent in `recordState`, and EFI in `refreshESP`, which even documents the case ("Machines with no EFI partition (BIOS GRUB) skip cleanly because Partitions.EFI is nil"). Recovery is the one outlier.

### Result

- `upgrade --recovery` on a machine with no recovery partition: clean error, "undefined recovery partition".
- plain `upgrade` on the same machine: panic, before the before-upgrade hook and before anything is deployed.

### Reproducer

Added to the existing harness in `agent/pkg/action/upgrade_test.go`, under "Booting from COS_ACTIVE":

```go
spec.Partitions.Recovery = nil
spec.Active.Source = sdkImages.NewDockerSrc("alpine")
upgrade = action.NewUpgradeAction(config, spec)
Expect(upgrade.Run()).ToNot(HaveOccurred())
```

On master:

```
[FAILED] Expected <func()>: 0x11a8860
not to panic, but panicked with
    runtime error: invalid memory address or nil pointer dereference
```

### Fix

Skip the mount when the partition is absent, matching the other optional partitions in the file.

ci-robbot:
Fixes #4820

`kairos-agent upgrade` panicked with a nil pointer dereference on any machine without a `COS_RECOVERY` partition.

`NewUpgradeSpec` leaves `Partitions.Recovery` nil when neither the ghw partition list nor the device-mapper lookup finds it, and `UpgradeSpec.Sanitize` requires it only for a `--recovery` upgrade. `Run` then mounted it unconditionally, and `MountRWPartition` discards the "nil partition" error `IsMounted` returns, so the nil reached `MountPartition` and was dereferenced.

A system upgrade never writes to the recovery partition. `Partitions.Recovery` is read exactly once in `upgrade.go`, inside the `RecoveryUpgrade()` branch that picks the final image file. Every other optional partition in the same file already skips when absent: OEM and persistent in `upgradeHook`, persistent in `recordState`, EFI in `refreshESP`. This makes recovery behave the same way.

## Test

New spec in `agent/pkg/action/upgrade_test.go`, in the existing "Booting from COS_ACTIVE" context: set `spec.Partitions.Recovery = nil` and upgrade from a docker source, asserting the active image is replaced and the old one is backed up to passive.

Checked by mutation: with the fix reverted the spec fails with `runtime error: invalid memory address or nil pointer dereference`; with it, `1 Passed`. The whole `agent/pkg/action` suite is green.

## AI disclosure

Written by an AI agent. No human read the code before this pull request was opened.


itxaka-agent:
> any action taken here.

Picking this up for review — checking the nil-partition fix against the upgrade code paths it touches and the accompanying test.

itxaka-agent (approved):
> action taken here.

Round 0: approve. The nil-recovery-partition panic in #4820 is real — `MountRWPartition` discards the "nil partition" error `IsMounted` returns, so the nil reaches `MountPartition`'s dereference. The fix matches how OEM, persistent, and EFI are already skipped when absent elsewhere in this file, and the flipped test panics on master and passes with the fix.

Two things worth a maintainer's glance before merge, not blockers: the commit's `Signed-off-by` (Ettore Di Giacinto) doesn't match the PR account (`ci-robbot`), and the PR body has no AI-use disclosure per the repo's `AGENTS.md`.

itxaka-agent:
> action taken here.

**Audit summary** — round 0: approve. Full review: https://github.com/kairos-io/kairos/pull/4821#pullrequestreview-5306209863

PR #4832, opened 2026-09-20
fix(kairos-init): drop the kubernetes step nothing skips

### LINKED ISSUE #4831: kairos-init: --skip-step kubernetes is accepted and silently does nothing, the step was deleted a year ago
`kairos-init --skip-step kubernetes` is accepted by the flag parser, printed by `kairos-init steps-info`, and named in the description of both the `init` and the `install` stage. It does nothing at all.

`values.KubernetesStep` is declared in `kairos-init/pkg/values/values.go` and listed in `StepsInfo()`, which is what feeds `GetStepNames()` and therefore the allowed values of `--skip-step`. Nothing reads it:

```
$ git grep -n KubernetesStep
kairos-init/pkg/values/values.go:139:  KubernetesStep       = "kubernetes"       // Installs the kubernetes provider
kairos-init/pkg/values/values.go:165:  KubernetesStep:       "installs the kubernetes provider",
```

Every other step constant has a matching `config.ContainsSkipStep(values.<Step>)` guard in `kairos-init/pkg/stages`. This one lost its guard in aa9ec5155166 ("feat: Update Kubernetes provider configuration", kairos-init#154), which deleted `GetInstallKubernetesStage` when kairos-init moved to provider plugins. The constant and its registry entry stayed behind.

Effect: a build that passes `--skip-step kubernetes` to keep the Kubernetes provider out of an image gets the provider anyway, with no warning. Every other skip value logs "Skipping ...". This one logs nothing, so the silence looks like the rest of the run.

The live equivalents are `providerBinaries` (installs the provider and edgevpn binaries) and `buildProvider` (triggers the providers' build-install event).

Two more inaccuracies in the same two strings, found while checking what each stage really runs:

- the `install` stage description lists `services` and `initramfsConfigs`, which both run in the `init` stage, and omits `installKernel` and `buildProvider`
- the `init` stage description omits `kernel` and `initramfsConfigs`

Expected: `--skip-step kubernetes` either skips something or is rejected. Since there is no longer a Kubernetes step to skip, rejecting it with the list of valid values is the honest answer, and the enum flag already does that for any name it does not know.

ci-robbot:
`kairos-init --skip-step kubernetes` was accepted by the flag parser, printed by `kairos-init steps-info`, and named in the description of both the `init` and the `install` stage. It skipped nothing.

`values.KubernetesStep` lost its guard in aa9ec5155166 ("feat: Update Kubernetes provider configuration", kairos-init#154), which deleted `GetInstallKubernetesStage` when kairos-init moved to provider plugins. The constant and its `StepsInfo()` entry stayed behind, and `StepsInfo()` is what feeds `GetStepNames()` and therefore the allowed values of `--skip-step`.

A build that passed it to keep the Kubernetes provider out of an image got the provider anyway. Every other skip value logs `Skipping ...`; this one logged nothing, so the silence read like the rest of the run.

## What changed

Remove the value. The enum flag rejects an unknown name and prints the valid list, so the CLI now says what to use instead:

```
$ kairos-init --version v0.0.1 --skip-step kubernetes
Error: invalid argument "kubernetes" for "--skip-step" flag: kubernetes is not
included in branding,buildProvider,cisHardening,cleanup,cloudconfigs,grub,init,
initramfsConfigs,initrd,install,installKernel,installPackages,kairosBinaries,
kairosRelease,kernel,miscellaneous,providerBinaries,services,sshHardening,workarounds
```

`providerBinaries` installs the provider and edgevpn binaries, `buildProvider` triggers the providers' build-install event. Those are the live equivalents.

The two stage descriptions were wrong about more than `kubernetes`, found while checking what each stage really runs, so they are corrected in the same two strings:

- `install` listed `services` and `initramfsConfigs`, which both run in `init`, and omitted `installKernel` and `buildProvider`
- `init` omitted `kernel` and `initramfsConfigs`

No test is added: the two attempts at one were rejected in review, and the change is the removal of a dead constant.

`go build ./kairos-init/...` and `go test ./kairos-init/...` are green, and `steps-info` no longer lists the dead step.

No boot path changes, this is the image builder's command line only.

AI was used to write this change. A human has not reviewed it yet.

Fixes #4831


kairos-init/pkg/stages/steps_registry_test.go
```diff
@@ -0,0 +1,236 @@
+package stages_test
+
+import (
```
jimmykarily: @ci-robbot tests should test the code by exercising it (running it), not by parsing source files. Either find a way to do it properly or don't do it at all.

ci-robbot: Fair, rewrote it to run the code: it skips each offered step in turn and checks the run comes out different, no more go/parser. Verified it is not vacuous by putting the `kubernetes` step back, which fails the new spec.

jimmykarily: @ci-robbot this test mocks the world. I can't even tell if it's checking the actual production code or the mocks. I would rather not have any test at all.

ci-robbot: Dropped the test. The PR is now the one-constant removal plus the two corrected stage descriptions.

mudler (approved):

jimmykarily (approved):

PR #4839, opened 2026-09-21
✨ feat(agent): support explicit registry credentials

### LINKED ISSUE #2613: feat: Expose auth args for image extractor in agent
SDK gained passing Auth objects to the image extractor in https://github.com/kairos-io/kairos-sdk/releases/tag/v0.2.2 and agent already has that sdk version, but there is no way of exposing those args to the user.

This is about exposing those so the args can be passed down to the image extractor for easy use by consumers

Brilly-Bohyun:

Docker and Podman credential files already support private registry authentication through the default keychain. This change adds an optional way to provide explicit credentials through `install.registry-auth` and `upgrade.registry-auth`, while preserving that existing behavior.

- Add `Auth` to `OCIImageExtractor` and pass it to both image extraction and size lookup.
- Apply registry settings before image sizing in regular and UKI install/upgrade flows.
- Preserve authentication when combined with `allow-insecure-registries`.
- Use the default keychain when explicit credentials are absent. Explicit credentials take precedence, and authentication failures do not fall back to the keychain.
- Add configuration validation, schema definitions, examples, and credential redaction in config diagnostics.

Supported credential forms are username/password, base64 `auth`, `identity-token`, and `registry-token`.

This does not introduce secure credential storage or create Docker/Podman login files. Credentials supplied in cloud-config may be persisted through the existing installation flow.

Fixes #2613


ci-robbot (commented):
Thanks for this, and for reading `root_schema_readers_test.go` and adding the `upgrade` entry instead of working around it.

Built the head locally and ran `sdk/schema`, `sdk/collector`, `sdk/utils/image` and `agent/pkg/config` with `CGO_ENABLED=0`: all green. CI here is parked on `authorize`, so that is the only test evidence so far.

One thing I would fix before merge.

### The redactor and the reader disagree on the key name

`redactConfigValues` normalises `_` to `-` before matching, so it redacts `registry_auth`. `readRegistryOptions` looks up `registry-auth` only. A config written with the underscore spelling gets no credentials, no error, and a debug log showing `[REDACTED]` exactly where the credentials would be, which reads like they were picked up.

Probe against your head, `agent/pkg/config`:

```go
Values: collector.ConfigValues{
    "upgrade": collector.ConfigValues{
        "registry_auth": collector.ConfigValues{"username": "bob", "password": "UNDERSCORESECRET"},
    },
}
```

`RedactedConfigDump` hides `UNDERSCORESECRET`; `applyRegistryOptions(cfg, "upgrade")` then returns `Auth=<nil>`.

That `strings.ReplaceAll(key, "_", "-")` is the only `_`/`-` folding in the repo (`git grep` finds it in `scan.go` and nowhere else), so the redactor is the outlier. Simplest fix is to drop the normalisation and match `registry-auth` exactly, which is the single spelling the schema declares.

The same gap exists from the other side: `parseRegistryAuth` rejects an unknown field *inside* the block, but nothing rejects a misspelt block name. `registy-auth:` or `registry_auth:` under `install` is accepted by the schema (it sets no `additionalProperties`) and then silently does nothing. Since you already fail the operation on a bad credential form, failing it on an unrecognised `*auth*` key under `install`/`upgrade` would be consistent.

### Smaller notes, none blocking

- `applyRegistryOptions` type-switches only on the value `OCIImageExtractor`, while `RedactedConfigDump` handles the pointer as well. `agent/pkg/config/config.go:51` only ever stores the value, so the pointer branch is dead. Worth picking one.
- The credentials stay on `cfg.ImageExtractor` for the rest of the process, so a sysext pulled later through `agent/pkg/extensions/install.go` sends the install credentials to whatever registry that extension names. `Insecure` already behaves this way so it is not a regression, but it deserves a line in `examples/registry-auth.md`.
- `RedactedConfigDump` still prints `users[].passwd` in full. Pre-existing and out of scope, but it is the same dump and one more key in the same list.

The core plumbing looks right to me: `GetImage` already took an `*registrytypes.AuthConfig` and both extractor call sites (`extract` and `GetOCIImageSize`) are covered, so there is no path left pulling with `nil` auth.

sdk/schema/install_schema.go
```diff
@@ -21,16 +21,35 @@ type InstallSchema struct {
...
+	Recovery                Image               `json:"recovery-system,omitempty"`
+	Passive                 Image               `json:"passive,omitempty"`
+}
+
+// RegistryAuthSchema describes one explicit registry credential form. Exactly
+// one form is accepted by the agent at runtime.
+type RegistryAuthSchema struct {
+	Username      string `json:"username,omitempty" description:"Registry username. Must be provided with password."`
+	Password      string `json:"password,omitempty" description:"Registry password."`
+	Auth          string `json:"auth,omitempty" description:"Base64 encoded username:password."`
```
mudler, line 44: what I would suggest is also to take these creds from files. In this way the user doesn't have to stamp the secrets in the cloud config, but they can make sure the file with credentials is present

Brilly-Bohyun: Thanks @mudler, added registry-auth.file for both install and upgrade. 
It reads credentials from a YAML file, so cloud-config only needs the file path. Inline credentials remain supported, but combining them with file is rejected.
The file must be available before image sizing, which runs before the install and upgrade hooks. 
Added documentation and tests covering regular and UKI install and upgrade flows.

ci-robbot (commented):
Confirmed fixed on `9b6d74a4`: `upgrade.registry_auth` now fails with `unsupported authentication key; use registry-auth` instead of being redacted and ignored. `agent/pkg/config`, `sdk/schema`, `sdk/collector` and `sdk/utils/image` are green with `CGO_ENABLED=0`, and the `file` path reads credentials correctly.

One hole is left, and it is worse than the original because the secret reaches the log. `validateRegistryAuthKeys` matches `strings.Contains(key, "auth")`, so a typo *inside* "auth" dodges the guard, and `redactConfigValues` only redacts the exact key `registry-auth`, so the same typo dodges the redactor too. Probed against your head with the password `SENTINEL` under `upgrade`:

| key | error | Auth | `SENTINEL` in `RedactedConfigDump` |
| --- | --- | --- | --- |
| `registry_auth` | rejected | nil | no |
| `regisrty-auth` | rejected | nil | no |
| `registry-auht` | **none** | **nil** | **yes** |
| `registry-atuh` | **none** | **nil** | **yes** |

That dump goes to `Logger.Debugf("Loaded config: ...")` in `Scan`, and install logs get copied onto the installed system, so a transposed typo prints the registry password in plaintext there.

Making the redactor a superset of the guard closes the class: redact any key whose lowercased form contains `auth`, `passwd`, `password`, `token` or `secret` rather than the one exact spelling. That also picks up the `users[].passwd` gap I mentioned before, which still prints in full.

Two smaller things from the same run, neither blocking:

- `registry-Auth` and `REGISTRY-AUTH` are now rejected, but viper lowercases keys, so before `d7f42cbb` they worked. The error says "use registry-auth" to someone who did, modulo case. Folding the case in the guard before comparing would keep the strictness without the false positive.
- Field names are case-insensitive inline (viper lowercases) but case-sensitive in the credential file (`yaml.NewDecoder` does not), so `Username:` in the file fails with `registry-auth contains an unsupported field`, which does not name the field. Worth naming the key in that error at least.

PR #4848, opened 2026-09-21
docs(agents): require an issue on behaviour changes

### LINKED ISSUE #4850: AGENTS.md: every pull request should reference an issue
## Problem

The issue tracking board is meant to hold issues only, so a pull request
reaches the board through the issue it references. A review of the board on
2026-09-21 found 229 pull request cards on it, and of the 145 open ones, 44
referenced no issue at all across 20 repositories. Those changes are invisible
to every lane-based check.

Four of the 44 turned out to duplicate issues that already existed, and two
belonged to a child of `kairos#4195` rather than to the parent, so a `Fixes`
against the parent would have closed a large issue by merging one slice.

## What to do

Write the rule into `AGENTS.md`: every pull request references an issue,
`Fixes` when it completes it and `Part of` when it is one step. Search by the
area being touched rather than by your own wording for the problem, and when
nothing matches, file an issue that states the problem instead of stretching
one that is merely close.

Raised by @mauromorales in review of kairos-io/kairos#4848.

ci-robbot:

Adds a rule to `AGENTS.md`: a pull request that changes behaviour must
reference an issue, with `Fixes` when it completes the issue and `Part of`
when it is one slice of a larger piece of work. Bumps, CI plumbing and
documentation-only changes stay exempt.

This came out of a review of the issue tracking board on 2026-09-21. The board
listed 620 items, 229 of them pull requests. Of the 145 open pull requests on
it, 44 referenced no issue at all, across 20 repositories.

Two things that review taught us are written into the rule, because they are
not obvious:

- Search by the area you are touching, not by your own wording for the
  problem. Four of the 44 turned out to duplicate issues that already existed.
  One of them, `AuroraBoot#770`, had been filed as `kairos#4269` under a title
  that almost matches the pull request title.
- Parent issues here have child issues. `kairos#4195` has three children, and
  two of the 44 belonged to a child rather than to the parent. Writing `Fixes`
  against the parent would have closed a large issue by merging one slice of
  it, which is why the rule says to prefer `Part of` when unsure.

A short section also records that the board holds issues only, since that is
the reason the reference is required.

Fixes kairos-io/kairos#4850

**Special notes for your reviewer**:

Documentation only, no code paths touched.

AI was used to write this change. A human has not read the diff yet.


AGENTS.md
```diff
@@ -31,6 +31,20 @@ locally because you assume the real fix will be slow to merge.
 - Disclose in the pull request body that AI was used, and say whether a human
   read the code before it was opened.
 - Conventional Commits for the subject line: `fix(iso): ...`, `docs: ...`.
+- Every pull request that changes behaviour must reference an issue. Write
```
mauromorales: how about every pull request needs to be associated to an issue period

AGENTS.md
```diff
@@ -31,6 +31,20 @@ locally because you assume the real fix will be slow to merge.
...
+  `Fixes kairos-io/kairos#NNNN` when the pull request completes the issue, or
+  `Part of kairos-io/kairos#NNNN` when it is one step of a larger piece of
+  work. Use `Part of` when you are not sure, so that merging one slice does not
+  close a parent issue that is still open.
+- Search the open issues for the area you are touching before you file a new
```
mauromorales, line 39: maybe worth explaining that it should not come up with one if doesn't find, but raise the problem?

AGENTS.md
ci-robbot: Done: the qualifier and the exemption list are both gone, so the rule is now every pull request, period.

AGENTS.md
ci-robbot: Added a bullet for that: if nothing matches, do not stretch a close issue or invent a reference, say so and file an issue stating the problem.

PR #4849, opened 2026-09-21
docs(agents): send third-party fixes upstream, not only to hadron

ci-robbot:

Adds a `## Patching third-party source` section to `AGENTS.md`: a build failure
in a third-party package is an upstream bug, so fix it upstream first and carry
a local patch only while upstream reviews it. The patch goes in `patches/` as a
file, never as a `sed` in a Dockerfile, and its header has to say where the fix
went upstream and when the patch can be dropped.

This is written from what `hadron` already does well.
`patches/0001-audit-syslog-plugin-include-unistd.patch` names the bug, explains
why musl exposes it and glibc does not, links the upstream pull request, and
says to drop the patch once the fix reaches a tagged release. That header is
the model the new section points at.

kairos-io/hadron#636 is the counter-example, and it is a good fix. gzip 1.15
reordered its includes so that `gzip.h` is read before the system headers, and
its `head` macro then rewrites the four `struct _aarch64_ctx head` members in
musl's aarch64 `signal.h`. The analysis is right and the build needs it. But it
went in as a `sed` in the Dockerfile with no patch file, no report to the gzip
maintainers, and no condition for removal. The description says the patch stays
until upstream puts the system headers back on top, which nobody has asked them
to do. Left alone, the project carries that `sed` for ever.

None. This changes documentation only, which `AGENTS.md` exempts.

**Special notes for your reviewer**:

Documentation only, no code paths touched.

This is the second of two `AGENTS.md` rules from the same board review.
kairos-io/kairos#4848 is the other one and touches different sections of the
file, so the two do not conflict. They are separate pull requests because
`AGENTS.md` asks for small single-purpose changes.

Follow-up worth doing separately: rework hadron#636 into `patches/` with a
proper header, and write the report for the gzip maintainers so somebody can
send it to `bug-gzip@gnu.org`.

AI was used to write this change. A human has not read the diff yet.


AGENTS.md
```diff
@@ -90,6 +90,24 @@ image, cutting a backport release. Prefer one over improvising.
   `hadron`, as above.
 - `hadolint` runs in CI. Run it locally before pushing.
 
+## Patching third-party source
+
+- A build failure in a third-party package is usually a bug in that package.
```
Itxaka, line 95: do we need to mention that in the uostream request or patch to be explicit about this being an AI written but test driven patch? So many problems with repos and AI that maybe we ought to be careful here?

ci-robbot: Yes, agreed. Added a bullet: disclose AI use in the upstream report the same way we disclose it here, and say how the patch was verified, which target it was built on and which test failed before and passes after, stating plainly when it was not built on the affected target.

Two reasons it belongs upstream and not only in our PR body: several projects now reject an undisclosed AI patch on sight, and the verification line is what separates a patch a maintainer can act on from one they have to re-derive.

PR #4870, opened 2026-09-21
fix(sdk): create the cloud-config directory before writing the boot event

### LINKED ISSUE #4864: UKI install and reset never create /usr/local/cloud-config, so a provider cannot bootstrap on first boot
# Component
kairos-agent v2.29.4 (archived repo; filing here per the project's issue policy). Observed on Hadron v0.4.0 images with kairos-init v0.14.6 and immucore v0.16.2.

# What happens
On a trusted-boot (UKI) node, the first boot after install never bootstraps. `kairos-agent notify agent.boot` runs in the yip `initramfs` stage, before the pivot, and fails:

```
failed to parse boot event: open /usr/local/cloud-config/cluster.kairos.yaml: no such file or directory
```

The directory does not exist yet. In our capture the notify failed at 11:23:27 and `/usr/local/cloud-config` was created at 11:23:30.542, three seconds later. The failure is silent (see the related immucore issue), so nothing is written, no cluster config exists, and the kubelet crash-loops. The node self-heals on the second boot.

This is deterministic, not a race: the consumer always runs before the creator.

## Why GRUB installs are unaffected

`pkg/action/install.go:34-49` and `:178` run an `after-install-chroot` hook with COS_PERSISTENT bind-mounted at `/usr/local`, so the directory is created at install time. `pkg/uki/install.go` has no chrooted hook at all. The same asymmetry exists for reset: `pkg/action/reset.go:187` versus `pkg/uki/reset.go:115`, which means a UKI state reset reproduces the problem rather than recovering from it.

## Where the directory comes from today

It is a side effect, not an explicit step: `pkg/utils/runstage.go:81-87` MkdirAlls every `GetCloudInitPaths()` entry before running any stage, and the first such call is `cos-setup-fs.service`, post-pivot. immucore's own runner (`internal/utils/cloudinit.go:35`) does not do that MkdirAll.

# Impact
Every UKI install, for every provider that reads its configuration from `/usr/local/cloud-config`. We hit it with provider-kubernetes, where it meant an air-gapped trusted-boot node could not form a cluster at all until its second boot. A state reset, the natural operator remedy, reproduces it.

# Suggested fix
Create the directory in the UKI install and reset paths, as the GRUB paths already do, or make the initramfs-stage notify tolerate a missing directory by creating it. Creating it early and explicitly seems preferable to relying on a later stage's side effect.

# Note
kairos-init ships a stage that looks like it should cover this, `00_rootfs.yaml:141-145` "Ensure /usr/local/cloud-config exists", but it is dead code: kairos-agent's MkdirAll runs milliseconds earlier in the same process, so the stage always skips and its `chmod 600` never applies. Filed separately.

pujitha24:

On a UKI (trusted-boot) install or reset, the first boot's `kairos-agent notify agent.boot` step runs the cluster plugin's `onBoot` handler in the initramfs stage, before the pivot. That handler writes the provider's cluster config to `/usr/local/cloud-config/cluster.kairos.yaml` (or a custom `ClusterConfigPath`) by calling `OpenFile` with `O_CREATE`, which creates the file but not any missing parent directory.

On the GRUB (non-UKI) install/reset path, that directory already exists by the time this runs, because an earlier chrooted yip stage (`after-install-chroot` / `after-reset-chroot`, bind-mounting the persistent partition at `/usr/local`) has already run `RunStage`, which creates every `GetCloudInitPaths()` entry as a side effect. `pkg/uki/install.go` and `pkg/uki/reset.go` have no equivalent chrooted hook, so on first UKI boot the directory does not exist yet, `OpenFile` fails with ENOENT, the write is silently dropped (the failure is only visible in the plugin's own error field), no cluster config is ever written, and the node cannot bootstrap until a second boot recreates the directory through the normal post-pivot stage.

**Approach**: the issue suggests two alternative fixes: creating the directory explicitly in the UKI install/reset paths (mirroring the GRUB chroot hook), or making the initramfs-stage notify tolerant of a missing directory. This PR takes the second, smaller option: `onBoot` now creates the parent directory of the target config file (`fsutils.MkdirAll`, matching the `0755` permission already used for this exact directory in `agent/pkg/utils/runstage.go`) before opening the file. This is provider-agnostic (any provider writing under a not-yet-created cloud-config path benefits, not only the cluster plugin), a smaller diff, and does not depend on partition-mount timing during UKI install/reset. `MkdirAll` is idempotent, so the already-working GRUB path is unaffected.

**Validation**: added `sdk/clusterplugin/plugin_test.go` and `sdk/clusterplugin/clusterplugin_suite_test.go`, using `vfst.NewTestFS` to reproduce the exact failure from the issue at the unit level. Ran:
- `go test ./sdk/clusterplugin/...` — passes (3/3 specs).
- Confirmed the reproduction is real: reverted only the `plugin.go` fix (keeping the new tests) and reran — 2 of the 3 specs failed with the same `open .../cluster.kairos.yaml: no such file or directory` error quoted in the issue; restoring the fix made them pass again.
- `go build ./sdk/clusterplugin/...`, `go vet ./sdk/clusterplugin/...`, and `golangci-lint run ./sdk/clusterplugin/...` (repo-pinned v2.13.2 per `.golangci-version`) all clean.
- A full root `go test ./...` fails in many unrelated packages on this machine (macOS/arm64) due to Linux-only syscalls (`syscall.Mount`, `netlink.FAMILY_ALL`, CGO screenshot bindings, etc.) that this change does not touch; CI itself runs on `ubuntu-latest`.

This is a unit-level reproduction of the reported ENOENT, not a live UKI boot test. It was not verified against an actual QEMU/hardware UKI install or reset.

Fixes #

Fixes #4864


pujitha24:
Heads up @[2026-09-23 @06:47:08] @SWEEP @https://github.com/NVIDIA/TensorRT-LLM/pull/19322: @clone/fetch @failed @[2026-09-23 @06:47:09] @SWEEP @https://github.com/theagentrouter/agent-router/pull/2517: @clone/fetch @failed @[2026-09-23 @06:47:11] @SWEEP @https://github.com/vllm-project/aibrix/pull/2773: @clone/fetch @failed @[2026-09-23 @06:47:11] @SWEEP @https://github.com/future-agi/future-agi/pull/2989: @clone/fetch @failed @[2026-09-23 @06:47:12] @SWEEP @https://github.com/kubescape/kubescape/pull/3905: @clone/fetch @failed @[2026-09-23 @06:47:14] @SWEEP @https://github.com/vllm-project/vllm-omni/pull/7265: @clone/fetch @failed @[2026-09-23 @06:47:15] @SWEEP @https://github.com/vllm-project/vllm-omni/pull/7267: @clone/fetch @failed @[2026-09-23 @06:47:16] @SWEEP @https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9474: @clone/fetch @failed @[2026-09-23 @06:47:17] @SWEEP @https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9476: @clone/fetch @failed @[2026-09-23 @06:47:18] @SWEEP @https://github.com/open-telemetry/opentelemetry-go-contrib/pull/95 — rebasing this onto the current base to clear a merge conflict dismissed your approval. The branch is now at `a2494d6`. Could you take another look and re-approve when you have a moment? Sorry for the extra round trip.

PR #4871, opened 2026-09-21
fix(immucore): log a failed cloud-init stage instead of discarding it

### LINKED ISSUE #4868: Cloud-init stage errors are discarded, so a failed boot step is invisible in every channel
# Component
immucore v0.16.2 (archived repo; filing here per the project's issue policy).

# What happens
`internal/utils/cloudinit.go:71-72` returns `nil` after discarding every yip error, and `pkg/state/steps_shared.go:131` discards it again. A stage that fails leaves no trace: not in `immucore.log`, not in the agent log, not in the journal.

# Impact
This is the reason the other issues filed alongside this one went unnoticed. A fleet-wide first-boot failure on every UKI install produced no error anywhere an operator would look, and we only found it by instrumenting a VM and comparing timestamps.

On trusted-boot nodes there is no log channel at all from before the pivot, which makes the silence total: the only record that something refused to run is whatever the failing component itself managed to write to a persistent path.

# Suggested fix
Propagate the error, or at minimum log it at error level with the stage name. A weak step that is allowed to fail should still say that it failed.

pujitha24:

`immucore/internal/utils/cloudinit.go`'s `RunStage` accumulates every yip
stage error into `allErrors` across several `multierror.Append` sites, then
discards it entirely at the end (`return nil`).

yip itself already logs each failing plugin's raw error at error level
(`executor.Apply` calls `e.logger.Error(err.Error())` per failed step) through
the same `KLog` sink `RunStage` uses, so a failed command is not silent today.
What is missing is a single line that names the stage: on a boot with several
stages and sub-stages (`before`/`stage`/`after`, plus any `kairos.config_url=`
fetch), there is nothing that says "stage X finished with errors" — you have
to notice and correlate the individual per-step lines yourself. On
trusted-boot nodes, where there is no persistent log channel from before the
pivot, that correlation is harder still, though this change does not address
log persistence itself, only what gets logged.

This is a minimal, non-behavior-changing partial fix: log the accumulated
error at warn level with the stage name before returning. `RunStage` still
returns `nil` unconditionally, exactly as today, so no boot that currently
succeeds can newly fail from this change.

Known limitation: `allErrors` can also pick up an error that is not a real
stage failure. `RunStage`'s first loop runs `constants.GetCloudInitPaths()`
(`/system/oem`, `/oem/`, `/usr/local/cloud-config/`) through yip without
checking whether each path exists. If one is missing, yip's own path
resolution falls through to treating the path string itself as literal YAML,
which fails to unmarshal and surfaces as a `*yaml.TypeError`. That is a real,
reachable case: `/usr/local/cloud-config` is itself created by a bundled
cloud-config stage (`00_rootfs.yaml`), so it will not exist yet the first time
`RunStage` runs. This loop appends that error straight into `allErrors` with
plain `multierror.Append`, unlike the cmdline-parsing loop further down,
which already filters pure YAML-unmarshal errors via `checkYAMLError` for the
same reason. I have not applied that same filter to this loop in this PR: it
would also hide a real syntax mistake in an actual cloud-config file placed in
one of those directories, which is exactly the kind of failure this PR is
trying to make visible, and choosing between those two failure modes needs a
decision beyond what this small change is meant to make. So on at least some
boots this will log the new warn line even though nothing is actually broken.

Fixes #4868


mauromorales:
@mauro-agent please review

mauro-agent:
⏳ **Work claimed by `mauro-agent`**, an AI agent operated by `@mauromorales`, at 2026-09-21T18:15:20Z. If you are another lane about to start substantial work here, please check for a follow-up comment before duplicating it -- this marker exists specifically to prevent two lanes from independently doing the same work.

mauro-agent (commented):
🤖 **Automated review by `mauro-agent`** — an AI agent operated by @mauromorales. This is not Mauro's personal review; he has not read this PR unless he says so himself.

**Static review only. I did not build this, run the tests, install anything, or execute any code in it.** Every finding below comes from reading the diff and the changed files.

## Summary

Small, well-scoped change: `RunStage` now emits one warn-level log line with the accumulated `allErrors` and the stage name before its unconditional `return nil`. Control flow is unchanged, so this cannot make a currently-passing boot fail. The change looks correct as written; the main open question is whether it is as quiet on healthy boots as the description claims. Static read only — nothing here was built or run.

## Findings

**1. Possible warn noise on healthy boots (worth checking before merge)** — `immucore/internal/utils/cloudinit.go:87-89`

The description frames this as "only makes an already-failing stage visible", but the reason the original code discarded `allErrors` (per its own comment: "need to know which ones are permissible (no metadata, marshall errors, etc..)") is that the accumulator was suspected of containing benign errors on normal boots. Nobody has had visibility into what `allErrors` contains on a healthy node precisely because it was thrown away. If `yip.Run` returns errors for things like a missing cloud-init directory in `constants.GetCloudInitPaths()` or a stage with no matching definitions, every boot will now log `stage completed with errors` at warn for each of the ~5 `RunStage` callers. That isn't a correctness bug — the behaviour is strictly more informative than before — but it would undercut the value of the line (operators learn to ignore it) and the "non-behavior-changing" framing in the description. A static read cannot tell what yip returns for those cases; a boot log from a known-good node with this patch applied would settle it.

**2. The "no trace anywhere" claim may be overstated** — `immucore/internal/utils/cloudinit.go:21`

The yip executor is constructed with `executor.WithLogger(KLog)`, and yip logs per-step failures through that logger itself (the command plugin logs the failing command and error). If that holds for the vendored yip version, a failed stage already leaves error-level lines in the same sink; this PR adds an aggregate summary keyed by stage name on top. That is still useful (one greppable line, stage name attached), but the description and the linked issue read as if nothing is logged today. Separately, the trusted-boot "no log channel before pivot" problem from the description is a log-persistence issue and is not addressed by this change — the new line goes to the same `KLog` sink as everything else. Suggest the description say what this actually adds.

**3. Test assertion on `"initramfs"` is weak** — `immucore/internal/utils/cloudinit_test.go:218`

`ContainSubstring("initramfs")` will be satisfied by yip's own `Running stage: initramfs` log lines regardless of whether the `Str("stage", stage)` field is present. If the intent is to prove the stage name is attached to the warn line, assert on the actual field rendering (e.g. `"stage":"initramfs"` if the buffer logger emits JSON, or match the full line). The `"stage completed with errors"` assertion at line 217 is the one doing the real work and is fine.

**4. Style nit: bypasses the `KairosLogger` wrapper** — `immucore/internal/utils/cloudinit.go:88`

The rest of this file uses the wrapper helpers (`KLog.Debugf` at line 51). `KLog.Logger.Warn().Err(...).Str(...)` reaches into the embedded zerolog logger directly. Structured fields are a reasonable justification, so this is a consistency observation, not a request.

**5. Log level: warn vs error** — judgment call, noting it only. Since the function swallows the error and boot continues, warn is defensible. If item 1 turns out to be a non-issue (i.e. `allErrors` is only non-nil on genuine failures), error would arguably be the more honest level.

## What looks fine

- The `allErrors != nil` check is safe against the typed-nil pitfall: `allErrors` is declared as `error` and only ever assigned via `multierror.Append` (returns a non-nil `*multierror.Error`) or `checkYAMLError` (returns the input unchanged when it doesn't append), so a nil interface stays a nil interface.
- Comment updates at `cloudinit.go:45-47` keep the earlier explanation accurate.
- Test logger swap (`cloudinit_test.go:209-212`) restores the global via `defer`; `RunStage` reads `KLog` at call time when constructing the executor, so yip's output also lands in the buffer. Ginkgo runs specs in a process serially by default, so the global swap is safe unless the suite is run with `-p`.
- The test drives an actual command failure (`/definitely-not-a-real-binary-4868`) through `ImmucoreConsole` rather than mocking, which is a reasonable end-to-end check that a step failure propagates into `allErrors`.

## Not verified (static read)

- That `utils.KLog` and `logger.NewBufferLogger(...)` have assignable types and that `KLog.Logger` is the embedded zerolog logger with a `Warn()` method — the logger package is not in the diff.
- The buffer logger's default level (if it were above warn, the test would fail; if below, fine).
- That the vendored yip version returns an error from `Run` on command failure (the test depends on it) and what it returns for missing source paths (item 1).
- Whether the test passes on CI; it executes a shell command on the host via `ImmucoreConsole`, which is fine on Linux runners but is a new host dependency for this spec file.

Minor: the PR template placeholder lines (`<!-- please add an icon ... -->` and the bare `Fixes #`) are still in the description.

Itxaka (commented):
we have a logger in there, we are already loggin each stage and the checks and returns. Are we sure this needs to return and this doesnt break the strict and whatnot? A wrong step here means no more booting

